PR #180, opened 2015-01-07
Update Travis CI -configuration

datakurre:
Should fix Travis CI -build (with build cache, which should speed up the builds after a few runs)

Build cache in Travis is only allowed for public repo builds without sudo, which requires enchant to be build locally.


svx:
thanks !!!

PR #181, opened 2015-01-08
More details and a link for Jsonmigrator

djowett:
Also reordered Fast Content import because it is unrelated to transmogrifier, 
whereas jsonmigrator is related to it

### DIFF
--- a/develop/plone/content/importexport.rst
+++ b/develop/plone/content/importexport.rst
@@ -216,7 +216,9 @@ collective.jsonmigrator
 =======================
 
 collective.jsonmigrator is basically a collective.transmogrifier pipeline that pulls Plone content from 
-to JSON views on an old site and writes it into your new site (this can also be done as a 2 step process).
+to JSON views on an old site and writes it into your new site. It's major advantage is that the JSON view
+product: collective.jsonify is very low on dependencies (basically just simplejson), so it can be installed on very old Plone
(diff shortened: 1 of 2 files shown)


polyester:
Hi @djowett, could you maybe fix two things first:

- make proper links out of the URL's in line 225, 227 etcetera in https://github.com/djowett/documentation/blob/master/develop/plone/content/importexport.rst ? While github automatically shows them as links, various testing tools don't, and it's a good habit to get into. Syntax is at http://docs.plone.org/about/styleguide.html#id9

-  use semantic linebreaks, as per http://docs.plone.org/about/styleguide.html#id2 for the new paragraphs you wrote?

I know, this seems like nit-picking, and there's tons of places where the linebreaks in the current docs are arbitrary. But we're trying to encourage better behaviour in new content, so over time the life of translators gets easier and the overall quality goes up.

Sorry for being a pain in the lower back regions, but once the habit is in your fingers it's easy to do.

and thanks for contributing!

djowett:
@polyester Ok, no worries - it's good to learn some stuff, well done on keeping things in order.

See if these changes are OK.

Just want to double check you saw I made changes to plone.app.i18n.locales.languageselector.rst on plone:master & djowett:master, so I guess gh will ask you to merge, but the changes I consider correct are in djowett:master

PR #182, opened 2015-01-13
possibly some enhancements to the resource registry documentation

thet:
@bloodbare @svx

### DIFF
--- a/adapt-and-extend/theming/resources.rst
+++ b/adapt-and-extend/theming/resources.rst
@@ -4,10 +4,11 @@ JS/CSS Resources
 
 .. admonition:: Description
 
-    Resources are js/css and its dependencies to add the frontend logic and design
-    on top of Plone. There are so many different kind of resources like widgets, design styles,
-    behavior logic and single page apps. In order to organize them we are using two
-    main standard technologies: require.js and less.
+    Resources are JavaScript, CSS or LESS resources and their dependencies.
+    They add frontend logic and design to Plone. There are many different areas
+    where there resources are used, like for widgets, design styles, behavior
+    logic and single page apps. In order to organize them we are using two main
+    standard technologies: require.js and LESS.
 
     The goal of this recipe is to help you confirm that everything is working.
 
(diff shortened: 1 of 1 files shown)


adapt-and-extend/theming/resources.rst
```diff
@@ -4,10 +4,11 @@ JS/CSS Resources
 
 .. admonition:: Description
 
-    Resources are js/css and its dependencies to add the frontend logic and design
-    on top of Plone. There are so many different kind of resources like widgets, design styles,
-    behavior logic and single page apps. In order to organize them we are using two
-    main standard technologies: require.js and less.
+    Resources are JavaScript, CSS or LESS resources and their dependencies.
```
gforcada: "Resources are .... LESS resources" can this second resources be changed to something else? files maybe?

bloodbare:
Thanks @thet for reviewing the documentation !

adapt-and-extend/theming/resources.rst
thet: AFAIK, in the resource registry only JavaScript, CSS and LESS are managed at the moment. Plain files can still be referenced in CSS/LESS/JS via static or resource directories.

bloodbare, line 7: It's possible to use references to folders for requirejs imports

thet:
ping. that's not fully merged. i think my other two commits make sense.

jensens:
What's up with this one here? ping @thet @svx @polyester

polyester:
Needs to be hand-merged, and please please please: use _semantic linebreaks_, do **not** break lines arbitrarily in the middle of a sentence.

Ideally we should see if this enhances or contradicts what we are saying in other places:
- https://github.com/plone/documentation/blob/5.0/develop/addons/upgrade_to_50.rst (which is basically a blogpost by Nathan)
- first paragraph of https://github.com/plone/documentation/blob/5.0/develop/addons/javascript.rst (which are the same blogposts by Nathan...)

I'm not qualified enough (yet) to check if we are using consistent terminology in all of these. 

I know this may sound pedantic, but **now** is the time to settle on a consistent terminology (even defining some of them in the glossary at https://raw.githubusercontent.com/plone/documentation/5.0/appendices/glossary.rst )

thet:
replaced by: https://github.com/plone/documentation/pull/397

PR #190, opened 2015-02-04
"hardcore" -> "hard-code"

daguar:
While I do think checking passwords into version control is hardcore, this should probably say "hard-code"


svx:
thanks !

PR #192, opened 2015-02-06
reworked/ modernized whole contribution to core process to reflect status quo

jensens:
Please review! @tisto also interesting for you I guess.

### DIFF
--- a/develop/plone-coredev/intro.rst
+++ b/develop/plone-coredev/intro.rst
@@ -6,29 +6,40 @@ Version Support Policy
 ----------------------
 If you are triaging or fixing bugs, keep in mind that Plone has a `version support policy <https://plone.org/support/version-support-policy>`_.
 
-Dependencies
-------------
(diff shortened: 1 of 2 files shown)


jensens:
btw, this is WIP, do not merge now!

svx:
Thanks for all you work !

Maybe there is another solution, since we have already docs in buildout.coredev we had the idea to 'fetch' these and include them into the documentation, with doing this we would make sure that we have them always up 2 date.
If we would do that, we have to figure out how to deal with different version, but I guess that would be doable. 
A big plus would be we do not have to update the docs about development in different places.

gforcada:
@svx buildout.coredev has also branches, so just pick the right branch for it, wouldn't that be enough?

svx:
@gforcada yes that would work !

polyester:
closing; this is fixed in 5.0 branch by using coredev.buildout directly; not worth the effort to fix up the 4.x branch.

PR #200, opened 2015-02-19
Improve example for 'Specify files and code from another package'

pysailor:
Hi @svx ,
following our little exchange on #plone today, here's my suggestion for improving "Specify files and code from another package".
I removed the old example, since I think having 2 examples might become too confusing.
Also, this is the first time I edit the documentation (shame, I know...), so I might have violated some rules.
For example, I implicitly assume that "BrowserView" will be a glossary term - but a link to a section that explains BrowserViews might be more appropriate.

### DIFF
--- a/develop/addons/components/zcml.rst
+++ b/develop/addons/components/zcml.rst
@@ -105,27 +105,45 @@ from another package, you can do so with using the
 configure tag which will then run the block of :term:`ZCML`
 in the context of that package.
 
-Here is an example of defining portlet manager to be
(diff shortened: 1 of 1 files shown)


svx:
@pysailor thanks a lot for working on that ! From the point of writing style guide, it looks OK for me, if we miss a glossary term or a link to the section I will fix that.  
Since all this zcml stuff is not really my strong suite, can someone else have a short look over it ? 
Thanks !

PR #213, opened 2015-04-08
Fix post-login-event example

lentinj:
Dragging over https://github.com/collective/collective.developermanual/pull/255

Unless I'm missing something, the example as-was made no sense. The return value of redirect_to_edit_access_folder wasn't even used. Steal the relevant chunk of code out of collective.onlogon to make it do what I think is something reasonable.

Also, remove grok whilst here.

This could be simplified further by removing the code to find the user folder (which isn't strictly relevant), but maybe this is useful for people? At very least, I've tried to separate the user-folder-finding part and the core of redirecting to a page on login.


svx:
thanks !

PR #228, opened 2015-04-24
Update filename

staeff:
The file in https://github.com/plone/papyrus.git is called bootstrap-buildout.py not bootstrap.py


svx:
thanks !

Issue #219, opened 2015-04-10
Add links to manual pages


svx:
Add in:
http://docs.plone.org/manage/installing/installation.html?highlight=wget#id4

the links to the docs parts were we have this documented

svx:
closed : https://github.com/plone/documentation/commit/9938d2e8717c17918dbb32a7d7de91756c62ae65

PR #229, opened 2015-04-24
Fix for #219

svx:
This will fix the broken reference link


polyester:
did some de-PEP8 on it while checkin.

PR #231, opened 2015-04-27
Blurb to mention Plock

aclark4life:
This blurb mentions the Plock installer; feel free to edit.

### DIFF
--- a/quickstart/index.rst
+++ b/quickstart/index.rst
@@ -69,6 +69,6 @@ Note however, that your chances of :doc:`getting help </askforhelp>` on setup qu
 
 Besides that, there are many people deploying Plone as part of other deployment tools, be they Ansible, Salt, Chef, Puppet or the like. If you are familiar with these, you will most likely find others in the Plone community that share your enthusiasm.
 
-You can also try `Plock: the Plone installer for the pip-loving crowd <https://plock.github.io>`. Plock is a Plone installer for Python folks already familiar with ``pip``. Plock creates a ``buildout.cfg`` for you and installs Plone with just few commands e.g. ``pip install plock; plock plone; plone/bin/plone fg``
+You can also try `Plock: the Plone installer for the pip-loving crowd <https://plock.github.io>`_. Plock is a Plone installer for Python folks already familiar with ``pip``. Plock creates a ``buildout.cfg`` for you and installs Plone with just few commands e.g. ``pip install plock; plock plone; plone/bin/plone fg``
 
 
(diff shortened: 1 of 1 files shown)


smcmahon:
Strongly -1 on this. Before we mention a tool, I think it needs to be stable and easy to install on all the platforms to which it seems to apply. Plock has its place, but it's going to be a lot of frustration for the average user. Try installing it on a clean copy of the latest Ubuntu. Or Windows. Or OS X.

If you can reword this to say exactly to whom it would be useful (and, no "pip-loving" doesn't tell us anything) and to point to reliable, community maintained installation instructions, then go for it.

svx:
Hi,

sounds definitely like an interesting project  ! Do I want to include into our main docs ? No, I am strongly -1, it is a very cool project and I definitely want to include it into our upcoming site about  3third party installer on the future plone.org but not into our official docs.

I will try to explain, I have different reasons ....

- there is a reason why we did not have 3third party installer like bitnami, openshift, nix and so on included into our official docs, for example there setup or way how the install plone is different.
Our main docs are based on our Unified-Installer and the 3third party installer are different, we do not want to confuse people, plock for example is doing things, different too, you work with pins, which is a different setup than we use in the UI.

That is about consistency of our docs.

- documentation: I miss documentation about plock, pins and all ! I just checked the repo on github, and there is none, besides the README.

- usage: I am not sure that plock will make it easier in its current state, sure doing 'pip install plock' sounds easy at least there are less steps involved than downloading and running the UI, but IMHO there are missing some important features, I was not able to find something about dependencies needed to run and install plone with plock, like system packages or libraries, meaning if I run this on a new machine only with pip installed it will not work to install plone with plock, I am not able to find checks for dependencies and something like meaningful error messages.
In the worst case this will give people who want to try is more frustration than good feelings :).

- tone of the readme: we try to increase the quality of docs.plone.org, so the readme and docs ? of plock would need a rewrite first.

Do not get me wrong, I think it is a cool project, and like I said I really would like to include it into our future site about 3third party installer, but I do not think that docs.plone.org is the right place :), but by all means keep on doing the good work you are doing !

vangheem:
my fault...

svx:
no not at all, actually I think it is good to have such discussions, since we still do not have all of this written down, it is important to talk about these things.

pigeonflight:
I am very much pro-Plock. I think the reservations expressed can be treated generally as a checklist of what needs to be in Plock 1.0 before it becomes an officially supported installer:

I can see the following as part of the checklist

- A more standardized approach (in some areas Plock varies from the rest of the docs for example using bin/plone istead of bin/instance). Any variations should probably be made options for the Plock commandline
- More comprehensive documentation (including how to work on windows (maybe vagrant))

I also would want to see how Plock or something that complements Plock can help to simplify the deployment story.

aclark4life:
In other news: anyone know why this build failed?

aclark4life:
@smcmahon @svx @pigeonflight thanks for the feedback! As @pigeonflight says, this is a great roadmap to Plock 1.0.

smcmahon:
This exchange seems to have reached a reasonable conclusion.

PR #233, opened 2015-04-28
Add link to zope.component docs

djowett:
I'm not sure how to put this actually, but the link I added shows that you can also register utilities, and so vocabularies with a factory.  "... In this case, the ZCML handler calls the factory (without any arguments) and registers the returned value as a utility.  Typically, you'd pass a class for the factory".

So a registration such as the one below is valid:

    <utility
        factory=".vocabularies.RowStylesVocabulary"
        provides="zope.schema.interfaces.IVocabularyFactory"
        name="collective.cover.RowColumnStyles"
        />


svx:
Can you also cherry-pick that to the Plone 5 branch ?

djowett:
Done

PR #240, opened 2015-05-03
Cleanup old docs, there are still in p3 docs of course

svx:
This removes old upgrade guides like upgrading from Plone 1 to Plone 2 and so on.

These docs are still available in our documentation for Plone 3, but we do not need yhem in out current docs.


polyester:
Would it be a good idea to reverse the order? So that the latest upgrade is at the top?

svx:
done

PR #242, opened 2015-05-04
give equal weight to ansible install

polyester:
if we're going to recommend ansible, might as well not hide it


svx:
@polyester can you also push that to the 5 branch ?

polyester:
not quite yet, I think, wasn't the Ansible playbook strictly 4.3 for now?

svx:
well for the momentyes, but as soon as we have a offical Plone 5, we will have also ansible for it. We do not do ansible for alpha and beta.

PR #246, opened 2015-05-06
Update wysiwyg.rst

Schemata:
Link to David's write up of how to disable html filtering instead. Note: I can't verify that David's method works as i was unable to get it working, but the currently doco'd method and link to SO certainly doesn't achieve the desired result in Plone 4.3.1.

### DIFF
--- a/develop/plone/forms/wysiwyg.rst
+++ b/develop/plone/forms/wysiwyg.rst
@@ -26,19 +26,15 @@ attacks. This will make Plone to strip away from HTML
 
 * Some other potentially unsafe tags and attributes
 
-If you need to put `<scrip>` tag on your content text in TinyMCE you can disable this security feature.
+If you need to put a `<script>` tag on your content text in TinyMCE you can disable this security feature.
 
 .. warning::
 
         If you don't trust all of your site editors, then this will open your site for an attack.
 
-In the Zope Management Interface, at the top of a Plone site, go to portal_transforms and open the safe_html transform. Disable it by putting a "1" in the disable_transform field and saving.
-
-A site restart might be required depending on Plone version.
-
 More info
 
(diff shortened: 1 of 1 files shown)


polyester:
If you're on the current version of Plone, only step 1 and 3 are necessary. So I'd much rather document those, and leave the scary step 2 out completely.

polyester:
so maybe copy the essentials from David's blog, like:

Step 1: Turn off Plone’s safe_html transform. Go to /portal_transforms/safe_html in the ZMI, and enter a 1 in the ‘disable_transform’ box. This prevents Plone from removing tags and attributes while rendering rich text.

Now you can edit a page, open TinyMCE’s HTML dialog, enter whatever you want, save, and it will get saved to the database and be shown on subsequent views of the page. However, if using Safari, Chrome, and IE, some unsafe markup will still not show up on the initial view of the page right after saving. Why not?

These browsers provide automatic reflexive XSS protection. That means that if they detect the same potentially unsafe markup in both a request and its response, they’ll block or ignore it. It’s normally a nice security precaution, but can be unwanted. Fortunately these browsers also provide a way for a site to opt out of the protection using an HTTP response header.

Step 2: Set the "X-XSS-Protection: 0" response header. This can be done in your frontend webserver such as apache or nginx. Depending on usecase, you can also only disable the protection for users who have permission to edit, so add this to the site’s main_template:

tal:define="dummy python:checkPermission('Modify portal content', context) and request.RESPONSE.setHeader('X-XSS-Protection', '0');"
Et voilà. No more filtering, and the editors can enter whatever markup they dream up.



(and then end with "more info")

Schemata:
OK will do...

polyester:
thanks!

PR #247, opened 2015-05-06
Update code examples

saily:
I've upgraded code examples to match our coding style guide:
- use lowercase method names (pep8)
- 2 spaces before classes and method names
- whitespaces around equals
- use `plone.api` instead of `getToolByName`

And finally i removed the link to svn `collective.pdfpeek` repo, which was migrated some years before ;-)


svx:
@saily thanks !

would you mind to push that into the 5 branch, too ?

PR #255, opened 2015-05-14
Bobtemplates p4

svx:
This updates the docs to say bobtemplates.plone is the new way instead of paster/zeopskel.

- update links
- add notes about bobtemplates.plone
- add deprecation messages with links to bobtemplates.plone

Please review this, since I changed a lot of files, maybe something slipped through

### DIFF
--- a/develop/addons/bobtemplates.rst
+++ b/develop/addons/bobtemplates.rst
@@ -1,0 +1,120 @@
+=========================================
+Bootstrapping Plone add-on development
+=========================================
+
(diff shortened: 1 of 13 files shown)


pbauer:
The docs for bobtemplates.plone are outdated. Where did you pull them from? For the right ones (with different options) see https://pypi.python.org/pypi/bobtemplates.plone

svx:
I see ther is still the old file in the docs, you are right that is old, and was there before we configured the pulling, I will remove it.

The new setup is pulling directly from GitHub

svx:
fixed !

PR #256, opened 2015-05-14
renaming images for sharing narrative. Fixes https://github.com/plone…

polyester:
…/documentation/issues/159

It won't fix sharing-11.png issue, but that's lost in time forever I'm afraid.


svx:
Thanks !

PR #272, opened 2015-06-01
Format and link fixes

hvelarde:


hvelarde:
@svx we still have a lot of issues with redirections; where is the configuration file for the web server?

plone-foundation:
Just send a list and we will add them to the list :)

This is a signature

PR #273, opened 2015-06-03
templating on tinymce

bloodbare:


svx:
Thanks we will review asap!

This is a signature

polyester:
I would like to include it not here, but somewhere under https://github.com/plone/documentation/tree/master/adapt-and-extend/config (which, admittedly, is an area needing more work...)

Reasoning: 
- this is something only a site admin will do, and even then not every day
- so it falls more in the 'how to allow IFrames and embeds' category of handbook, whereas this part is aimed directly at content editors and their daily work.

PR #277, opened 2015-06-18
Note that ZEXP isn't safe for intid references.

lentinj:
This is really a bug that needs to be fixed, but worth noting so people are aware.


lentinj:
Actually, hold this for now, there's probably more to this than "don't work".

Issue #280, opened 2015-06-24
Broken tutorial link


avoinea:
Here http://docs.plone.org/develop/plone/views/browserviews.html#more-information

`Zope view tutorial`. link is broken

polyester:
I will replace with links to the newer Mastering Plone training

PR #284, opened 2015-07-01
fix https://github.com/plone/documentation/issues/280

polyester:


svx:
@polyester would youmind to fix that in the 5 branch too ?

polyester:
no problemo. Once merged, I'll cherrypick. (did that with previous merge-of-the-day also)

PR #337, opened 2015-08-03
Add new changelog, prepare august release

svx:
prepare release for august 2015


polyester:
changing the titling, otherwise we would have 3 times 'changelog'

PR #340, opened 2015-08-09
Mrtango plone5 theming

MrTango:


polyester:
thanks a lot! Have added some linebreak fixes, will merge when travis is done. Nice!

Issue #323, opened 2015-07-17
remove lingual plone stuff


svx:
- remove all lingual.plone stuff from Plone 5 docs

polyester:
The following pages need review from @bloodbare or @sneridagh , to state the p.a.m. version of things.

- [ ] https://github.com/plone/documentation/blob/5.0/develop/plone/i18n/cache.rst (it can probably go completely, but double check
- [ ] https://github.com/plone/documentation/blob/5.0/develop/plone/searching_and_indexing/query.rst (mentions LinguaPlone 5 times...)
- [ ] https://github.com/plone/documentation/blob/5.0/develop/plone/i18n/language.rst

rest of mentions are taken care of in https://github.com/plone/documentation/issues/344

sneridagh:
1. https://github.com/plone/documentation/blob/5.0/develop/plone/i18n/cache.rst can go away because it's something that can be done from varnish itself (unset the cookie for some resources).
2. https://github.com/plone/documentation/blob/5.0/develop/plone/searching_and_indexing/query.rst @bloodbare should review it to be sure.
3. https://github.com/plone/documentation/blob/5.0/develop/plone/i18n/language.rst the mentions are valid, but should refer to p.a.m instead of LinguaPlone. It should be reviewed as well since it's possible that some reference in code is wrong.

svx:
@bloodbare could you review https://github.com/plone/documentation/blob/5.0/develop/plone/searching_and_indexing/query.rst if you have a minute or two ? Would be great to have that in the 5.1 docs ! :)

svx:
@polyester @sneridagh Who would be the right person to review: https://github.com/plone/documentation/blob/5.0/develop/plone/i18n/language.rst ?

stevepiercy:
Closing as won't fix for Plone 5. We now have https://6.docs.plone.org/i18n-l10n/index.html and https://6.docs.plone.org/volto/development/i18n.html. We don't have good search docs yet, though, but we have an open issue for it at https://github.com/plone/documentation/issues/1730

PR #344, opened 2015-08-12
Polyester fixvarious

polyester:
sorry for the massive commit, I was on a roll a bit. Should dramatically reduce the number of build warnings and errors. Main ones left are 

- the bogus "image not readable' (they are fake because robotframework creates them)
- ones in /external (should be fixed by authors)
- and some in buildout.coredev section. Not fixing now because we should pull that in as external as well, I have some pull request there to make that better.

review carefully, there are some editorial decisions in here as well, such as dropping the docs for Linguaplone and raptus.multiplone, with a link to the old docs.


svx:
@polyester I guess we also have to check which other tickets are may involved into this PR, I konw for example that we have a ticker for LingualPlone #323 .

So we should check the open tickets to and update or if possible close them.

The rest looks awesome thanks a lot !!

polyester:
Don't merge yet, working on removing more linguaplone stuff (and yakshaving in the process)

svx:
GO GO GO, just tell me when it is time for review/merge !

polyester:
There's three mentions of LinguaPlone that I can't take care of, needs somebody with more knowledge. Documented in https://github.com/plone/documentation/issues/323

polyester:
feel free to review/merge so far; the remaining ones need external review so that could be a while.

svx:
In:

develop/plone/searching_and_indexing/indexing.rst

and
 
develop/plone/i18n/internationalisation.rst

we should fix the way how the links are done, currently they are using the 'list at the bottom style' like:

.. _zope.i18n: https://pypi.python.org/pypi/zope.i18n
.. _zope.i18nmessageid: https://pypi.python.org/pypi/zope.i18nmessageid

We should change that into 'in text links', because they are better/easier for translations.

@polyester what is your opinion, do that now, or some day later ?

polyester:
done for these two. My opinion is about the same as fixing linebreaks: when opportunity arises. One day I will write a shell script to automagically do it. Today is not that day.

Issue #302, opened 2015-07-13
develop/addons/javascript.rst


svx:
Update documentation about how we do JavaScript in Plone 5

ichim-david:
http://docs.plone.org/develop/addons/javascript_standards.html needs to be updated since I don't see jquery-migrate plugin being loaded in Plone 5.
This means that code snippets that use $.browser will break ever since jquery 1.9 and since plone 5 ships with 1.1.11 people will run into javascript errors.

timitos:
Some interesting points for the javascript documentation can be found in plone/Products.CMFPlone#1163

timitos:
Here is a small collection of topics that need to be covered in the javascript documentation:
- Modern Javascript in Plone with requirejs
- Mockup and patternslib
- How to integrate legacy javascript (covered in [1])
- How to setup a dev environment for the javascript story
  - Requirements
  - Installation
  - JSHint
- Usage of the development mode of the resource registry
- Adding a resource or bundle to a request (covered in [1])
- How to add javascript or a pattern to an addon
  - TTW (covered in [2])
  - By Addon (covered in [1])

A few of these things are covered in:
[1] http://docs.plone.org/adapt-and-extend/theming/resourceregistry.html
[2] http://training.plone.org/5/javascript/index.html

timitos:
@frisi @thet maybe you want to join working on this issue.

timitos:
There is some more content to integrate in this document:
https://docs.google.com/document/d/1vg3K98oGk8wLkaT7uBZ-O6ufNw3NSupro4mK96hZrao

tkimnguyen:
I just struggled to find information on how to add JavaScript to an edit form, until I found this thread https://community.plone.org/t/add-custom-javascript-to-the-edit-page/2537, to which I added my own details: https://community.plone.org/t/add-custom-javascript-to-the-edit-page/2537/3. I would like to add it to the https://docs.plone.org/develop/addons/javascript folder somehow.

stevepiercy:
@petschki @MrTango is this issue still valid for Plone 6 for managing JavaScript in add-ons for the backend and Classic UI? 

We're not going to update Plone 5 docs.

petschki:
This is outdated.

We need to look at the existing docs here https://6.docs.plone.org/classic-ui/mockup.html and here https://6.docs.plone.org/classic-ui/module-federation.html how we can improve and extend the info for addon developers.

PR #351, opened 2015-09-03
Fixed link to jsvariables.py, only available on Plone 4.3.

mauritsvanrees:


polyester:
Note: this needs update for Plone 5, see also https://github.com/plone/documentation/issues/302

PR #363, opened 2015-09-16
upgrade doc regarding plone app users

ebrehault:


ebrehault:
bad branch, sorry

tisto:
@polyester @svx it seems I'm not the only one who makes this mistake. :) 

Can't we make the current Plone version the master branch? Or at least make the 5.0 branch the default one? It is really confusing that the default master branch is not used at all.

polyester:
that was going to be my job in Munich ;-)

Yes, we need to cut a 4.x branch from master, and then make 5.0 the default branch. And do the same on github.com/plone/papyrus, and reflect it in their respective readme's. So it's a bit of a shuffle, which I can't do before friday morning.

tisto:
Ok, cool. Thanks for working on this! :)

PR #368, opened 2015-09-18
DO NOT MERGE YET link to build docs

polyester:
don't merge yet, trying to figure out the travis errors as well as fix this one tiny link ;-)


polyester:
mr.developer hates mercurial (again, it has been doing that several times over the last few years). That generates errors which travis doesn't like. Fixed (I hope) by moving sphinx-intl from mercurial to git.

Issue #334, opened 2015-07-22
Unable to build under FreeBSD (make externals: Error expanding embedded variable)


mjollnirpaul:
Can s/o help?  I want to build the docs for offline use like described on the start page on github.
Here's what I did:
[root@paul-lifebook /usr/local/www/Plone/papyrus]# ./bin/python bootstrap-buildout.py
Downloading https://pypi.python.org/packages/source/s/setuptools/setuptools-18.0.1.zip
Extracting in /tmp/tmp3IUMcp
Now working in /tmp/tmp3IUMcp/setuptools-18.0.1
Building a Setuptools egg in /tmp/tmpc7FVRV
/tmp/tmpc7FVRV/setuptools-18.0.1-py2.7.egg
Creating directory '/usr/local/www/Plone/papyrus/develop-eggs'.
Generated script '/usr/local/www/Plone/papyrus/bin/buildout'.
[root@paul-lifebook /usr/local/www/Plone/papyrus]# bin/buildout
mr.developer: Queued 'documentation' for checkout.
mr.developer: Queued 'plone.api' for checkout.
mr.developer: Queued 'sphinx-intl' for checkout.
mr.developer: Queued 'sphinx.themes.plone' for checkout.
mr.developer: Skipped checkout of existing package 'sphinx.themes.plone'.
mr.developer: Skipped checkout of existing package 'documentation'.
mr.developer: Skipped checkout of existing package 'plone.api'.
## The current global buildout threat level is:   GUARDED  

Upgraded:
  zc.buildout version 2.2.4,
  setuptools version 7.0;
restarting.
Generated script '/usr/local/www/Plone/papyrus/bin/buildout'.
mr.developer: Queued 'documentation' for checkout.
mr.developer: Queued 'plone.api' for checkout.
mr.developer: Queued 'sphinx-intl' for checkout.
mr.developer: Queued 'sphinx.themes.plone' for checkout.
mr.developer: Skipped checkout of existing package 'plone.api'.
mr.developer: Skipped checkout of existing package 'sphinx.themes.plone'.
mr.developer: Skipped checkout of existing package 'documentation'.
## The current global buildout threat level is:   GUARDED  

Develop: '/usr/local/www/Plone/papyrus/source/plone.api'
warning: no files found matching 'docs/README.rst'
warning: no previously-included files matching '_.pyc' found anywhere in distribution
Develop: '/usr/local/www/Plone/papyrus/source/sphinx-intl'
[...]

PR #373, opened 2015-09-21
various fixes

polyester:
removing references to Plone3, and other fixes


svx:
Currently building the docs these way is failing on for example FreeBSD, that is the reason why the docs mention on which OS the docs build is working.

Maybe the wording was not OK, but removing that is also not really cool, there is also #334.

svx:
Sorry my mistake I was too fast reading

PR #376, opened 2015-09-24
including non-requirejs scripts with plone

instification:
This is some notes on adding non requirejs javascript to a theme whilst also including the plone bundles.

Raised by https://github.com/plone/Products.CMFPlone/issues/984


davilima6:
Thanks! As a minor suggestion I'd drop the `type` attribute since that's both implicit and best practice for HTML5 (on `script`, `style` and `link` tags). I will open an issue suggesting that for Plone core as well.

davilima6:
Thanks! Here it is: plone/Products.CMFPlone#1078

PR #377, opened 2015-09-25
Add basic docs for frontend/backend theming

instification:
Some very rudimentary documentation on using the backend version of Barceloneta in a theme.

Original issue https://github.com/plone/Products.CMFPlone/issues/465
Relevant pull request https://github.com/plone/plonetheme.barceloneta/pull/40

### DIFF
--- a/adapt-and-extend/theming/resources.rst
+++ b/adapt-and-extend/theming/resources.rst
@@ -371,19 +371,23 @@ Example::
         <xi:include href="++theme++barceloneta/backend.xml" />
 
         <!-- Only theme front end pages -->
-        <rules css:if-content="body.viewpermission-view,body.viewpermission-none">
+        <rules css:if-content="body.frontend" css:if-content="#visual-portal-wrapper">
 
-            <rules css:if-content="#visual-portal-wrapper">
+            <theme href="index.html" />
 
-                <theme href="index.html" />
+            <!-- Include basic plone/toolbar bundles -->
+            <after theme-children="/html/head" content="/html/head/link[@data-bundle='basic' or @data-bundle='plone' or @data-bundle='plone-logged-in' or @data-bundle='diazo']" />
+            <after theme-children="/html/head" content="/html/head/script[@data-bundle='basic' or @data-bundle='plone' or @data-bundle='plone-logged-in' or @data-bundle='diazo']" />
 
-                <!-- Your diazo front end rules go here -->
+            <!-- Insert the toolbar -->
+            <before css:theme-children="body" css:content-children="#edit-bar" css:if-not-content=".ajax_load" css:if-content=".userrole-authenticated" />
 
-            </rules>
+            <!-- Your diazo front end rules go here -->
+
         </rules>
     </rules>
 
-TODO - including the toolbar in the front-end
+You can define your own diazo bundle in your manifest.cfg (by using development-js, production-js and css options). This diazo bundle will not be included in the backend theme.
 
 
 Browser Page bundle
--- a/adapt-and-extend/theming/resources.rst
+++ b/adapt-and-extend/theming/resources.rst
@@ -371,13 +371,13 @@ Example::
         <xi:include href="++theme++barceloneta/backend.xml" />
 
         <!-- Only theme front end pages -->
-        <rules css:if-content="body.frontend" css:if-content="#visual-portal-wrapper">
+        <rules css:if-content="body.frontend#visual-portal-wrapper">
 
             <theme href="index.html" />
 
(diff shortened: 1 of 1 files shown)


adapt-and-extend/theming/resources.rst
```diff
@@ -356,9 +356,34 @@ The options are :
...
+        xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+        xmlns:xi="http://www.w3.org/2001/XInclude">
+
+        <!-- Include the backend theme -->
+        <xi:include href="++theme++barceloneta/backend.xml" />
+
+        <!-- Only theme front end pages -->
+        <rules css:if-content="body.viewpermission-view,body.viewpermission-none">
+
+            <rules css:if-content="#visual-portal-wrapper">
```
djay: You can simplify that to 

```
        <rules css:if-content="body.frontend" css:if-content="#visual-portal-wrapper">
```

adapt-and-extend/theming/resources.rst
```diff
@@ -356,9 +356,34 @@ The options are :
...
+
+                <theme href="index.html" />
+
+                <!-- Your diazo front end rules go here -->
+
```
djay, line 386: Need to mention you will need to include a bundle that includes support for the toolbar.

adapt-and-extend/theming/resources.rst
davilima6: You can simplify even more with:

```
<rules css:if-content="body.frontend, #visual-portal-wrapper">
```

djay: No, that means OR, not AND.

davilima6: You´re right, I'm sorry. Only way to achieve AND would be with [nested conditions](http://docs.diazo.org/en/latest/advanced.html#condition-grouping-and-nesting).

polyester:
@instification is this ready to be merged now?

jensens:
very useful!

instification:
@polyester yes I think so. I've tested the example locally and it works.

adapt-and-extend/theming/resources.rst
```diff
@@ -356,9 +356,38 @@ The options are :
...
+            <theme href="index.html" />
+
+            <!-- Include basic plone/toolbar bundles -->
+            <after css:theme-children="head" css:content="head link[data-bundle='basic'], head link[data-bundle='plone'], head link[data-bundle='plone-logged-in'], head link[data-bundle='diazo']" />
+            <after css:theme-children="head" css:content="head script[data-bundle='basic'], head script[data-bundle='plone'], head script[data-bundle='plone-logged-in'], script link[data-bundle='diazo']" />
+
+            <!-- Insert the toolbar -->
+            <before css:theme-children="body" css:content-children="#edit-bar" css:if-not-content=".ajax_load" css:if-content=".userrole-authenticated" />
+
```
djay, line 384: The backend.cfg already inserts the toolbar. Perhaps it shouldn't if thats confusing?

instification, line 384: It seems odd to include the toolbar without the necessary js/css.

I think it would make more sense to let the developer choose to add the toolbar to the frontend (using the example here) and just include the toolbar for non-frontend pages by default.

adapt-and-extend/theming/resources.rst
```diff
@@ -356,9 +356,38 @@ The options are :
 Diazo frontend - barceloneta backend
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Using diazo rules you can define a frontend and a backend separatelly defining which bundles you want to load.
+Using diazo rules you can theme the frontend of your site how you like, and use the default Barceloneta theme for the backend.
```
tomgross, line 359: There is a limitation with this construct: http://stackoverflow.com/questions/17202134/diazo-xsltemplate-not-applying-when-inside-secondary-rules-file
You can't unfortunately place Inline XSL directives into a condition. @djay any ideas how to work around this?

instification, line 359: Would the `<replace/>` workaround in the accepted answer not work for a lot of (possibly all) scenarios where you would need to use inline xsl?

Is this also not a similar problem regardless of whether you are using this backend template? In the [theming guide](https://plone-theming-with-diazo.readthedocs.org/en/latest/snippets_diazo/rules/index.html) we see:

```
 <!-- theme wrapper, to prevent theming ZMI -->
 <rules css:if-content="#visual-portal-wrapper">
```

so inline xsl directives would also need to be defined outside the nested `<rules>`, potentially causing issues with the zmi.

Issue #384, opened 2015-09-30
Update http://docs.plone.org/manage/installing/installation.html


svx:
- remove wip note
- adjust the mix up between 5.0 and rc1, both are mentioned on this site

PR #388, opened 2015-10-01
Fix for #384

svx:
- remove to do note
- fix Plone version

this document still needs improvements, some parts are/could be confusing but at least for now this small parts are fixed


jensens:
meant as WIP or ready for merge?

svx:
Ready for merge ! 

Sorry that this was not clear, I assumed different because I did not use the label wip or said so in the title.

PR #393, opened 2015-10-02
Change Travis to test instead of build

datakurre:
Let's wait for Travis. May be issues.


datakurre:
Wow, so many unknown directives.

@polyester I still prefer to make pybot work instead of making sphinx-build to fail hard. I'll figure out a way to make it support Sphinx-directives when sphinx exists to eventually make this pull pass...

datakurre:
@polyester @svx I managed to take another look and it seems that missing directives (those that get registered for sphinx-build but not for pybot) don't prevent the tests being run. So, after all, one could simply `bin/pybot source` to run just the screenshot scripts as tests and fail hard when they fail.

Docutils could be silenced with `./docutils.conf`:

```
[general]
report_level = 5
```

datakurre:
I also see that builds are getting pretty slow (a lot of sandbox setup and teardown for screenshots happening there). I seem to be getting an idea how to fix that. I try to make new plone.app.robotframework before the conf so that you'll get faster builds there.

datakurre:
https://github.com/plone/plone.app.robotframework/commit/890c27d4ab10379a2501f2def0bc22e1de4d2f21

datakurre:
@svx @polyester Also trying to fix egg caching at Travis to get the build faster. Yet, if you have time, please try to add new extension `plone.app.robotframework.server` into papyrus/conf/conf.py. It should speedup RF builds with sphinx.

datakurre:
@svx @polyester 
- travis egg cache is now working again (buildout is instant)
- plone.app.robotframework >= 0.9.13 works, but for some reason I had to pin it in buldout cmd (version from plone 5 versions.cfg sets it 0.9.13 otherwise)
- build time for pybot test was 7 min (but there was one failure and we don't have S3 test report uploads configured)

I continue to add Sphinx-build without screenshots after pybot run to get also the doc "tested".

datakurre:
Done. Required one more plone.app.robotframework release (>= 0.9.14).

To summarize:
- Faster Travis-CI -builds by caching buildout eggs
- Faster screenshot generation with new p.a.robotframework features to share sandbox between serial rst documents with the same sandbox configuration (requires `--listener plone.app.robotframework.server.LazyStop` for pybot or `plone.app.robotframework.server` extension for Sphinx).
- Travis to run first Robot Framework tests for docs and then Sphinx

Example build: https://travis-ci.org/plone/documentation/builds/84713268

This should get even faster once merged with the phantomjs branch.

PS. During the upcoming sprints / trainings / etc..., remember that you can also skip Travis-CI -build by adding `[ci skip]` into commit message. /cc @pbauer 

PPS. I didn't know that we have already that much generated screenshots. Awesome work! And as a bonus, the documentation is slowly becoming comprehensive acceptance test suite for Plone :-)

polyester:
Tried it locally as well, seems to work fine. Maybe I'll try to look into two things:
- see if there's a less heavy-handed way of excluding plone.app.robotframework docs than rm -rf, since that also makes them unavailable in the subsequent sphinx-build, thus generating extra warning. Can't we just give pybot an ignore-pattern?
- work it into the standard papyrus makefile as well this way, so you would have

"make screenshots" that just creates the screenshots using lazyloading
"make html" which does not use robotframework
"make full" which does make clean, make screenshots, make html

does that seem like a workable idea?

datakurre:
Sounds perfect. You mentioned yourself looking, so I don't interfere more :)

It would be nice to also optimize test browser window openings, but phantomjs probably gives the same saving.

polyester kirjoitti su lokakuuta 11 13:39:14 2015 GMT+0300:

polyester:
well, @datakurre, it seems pybot has no exclude pattern for files, but can do --exclude  for tags. That would mean tagging the plone.app.robotframework docs/tests with some unique tag. And since you are maintaining those, I at least need your permission. Or maybe they already all have a tag, you never know ;-)

https://github.com/robotframework/robotframework/blob/master/doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst

datakurre:
The best would be that I'd fix those tests run, but cannot do it this week.

If you can tag them, maybe use some generic tag like "incomplete" or "wip".

polyester:
I've added a pull-request to tag them with "wip-not_in_docs". https://github.com/plone/plone.app.robotframework/pull/48

together with https://github.com/plone/papyrus/pull/73 we can re-write travis.cfg to use the same syntax. I'll do that tomorrow.

svx:
What is the status of this PR ?

datakurre:
It used to be ready. I don't know how it is now.

sven kirjoitti to joulukuuta 3 09:08:04 2015 GMT+0200:

polyester:
for clarity, it should be slightly rewritten, 'script' can now become 
- make screenshots
- make html

which makes for a more readable file, and also means we keep the actual commands only in one place so they don't diverge.

and update --setuptools-version and --version to be in line with what papyrus now uses.
I'll fix on the weekend.

polyester:
It's still not **actually** testing, since we're pulling in documentation from git https://github.com/plone/documentation, which in end-effect is testing the _previous_ commit. But that's not part of this PR, so I'll merge this and see if I can be cleverer in another one.

PR #397, opened 2015-10-06
resources.rst - enhancements and wording

thet:
Follow up of: https://github.com/plone/documentation/pull/182


polyester:
great work, tnx @thet ! 

just added some .. code-block:: xml statements to get prettier colors.
Note to self: do _not_ add .. code-block:: javascript to incomplete/invalid/example javascript, because the parser is strict and will not show them.

PR #400, opened 2015-10-09
Plone resource registry - completly reworked

thet:
I've completly reworked the resource registry as part of the JavaScript training.

I think the current version is more consistent. I've also moved a theming-only chapter to the barceloneta documentation.


thet:
I have also renamed the file from a bit misleading name resources.rst to resourceregistry.rst. Makes it a bit harder to compare with the previous version.

PR #401, opened 2015-10-10
Phantomjs changes 

polyester:
it seems phantomjs has difficulty in using CSS children selectors in overlays, such as css=#someID a

by picking a unitary selector it works, while running with Firefox also works


jensens:
i'am curious why it worked before w/o a wait...

polyester:
@jensens because phantomjs is a lot faster...

PR #402, opened 2015-10-11
warn about mod_cache

tschorr:
Even though the "Apache and Plone guide" is already marked as old, I think there should be some extra warning about mod_cache's default settings. Adding CacheIgnoreHeaders Set-Cookie to the configuration should also obsolete the section about "Unsetting language cookie for media content".


polyester:
thanks!

PR #415, opened 2015-10-27
Add note about add-ons

svx:
- add a note, telling that not all add-ons are compatible with Plone 5 , yet

### DIFF
--- a/_templates/version_warning.html
+++ b/_templates/version_warning.html
@@ -3,6 +3,7 @@
   <div class="col-sm-12 admonition warning version_warning" >
     <p class="first admonition-title">Warning</p>
    <p class="last">This is the documentation for the just-released Plone 5, and is updated regularly.<br />
+    Please note, not all add-ons are working with Plone 5, yet !<br />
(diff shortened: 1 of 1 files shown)


polyester:
hmm, do we really want this to appear above every single page, or does the message then become so long that people just ignore it? Also it then targets audiences like content-editors that can't actually do anything about it; they cannot even install add-ons anyway.

Maybe a big fat warning at http://docs.plone.org/adapt-and-extend/install_add_ons.html would be more targeted towards the audience of 'integrators' and 'site managers'.

Not sure, just opening up for discussion.

svx:
Fine with me too, I just think we should mention that till we have more add-ons working with Plone 5, seeing the messages on irc/community, we should really try to be a bit more clear about that, even it it should be obvious that you check first.
Apparently people are not checking if add-ons are updated, yet, they just install them without version check.
So imho we should mention that.

polyester:
OK, will make a clear note tomorrow.

https://github.com/plone/documentation/issues/416

PR #419, opened 2015-10-29
Apache load balancing: use pending request counting algorithm

tschorr:
I'd like to suggest a small change for the Apache Load Balancing section:
Using pending request counting instead of Apache's default request counting works well with Plone and will result in a better overall performance of a site. This PR also fixes some small typos/whitespace.


polyester:
thanks!

Issue #421, opened 2015-10-31
Add docs about collective


svx:
Currently we have:
http://docs.plone.org/develop/index.html#developing-for-plone-core

which is nice !!

But we still miss the part about https://collective.github.io/ .

IMHO we should add that to the docs, too - usually people start with stuff in collective, before the hack on the core.

Issue #416, opened 2015-10-27
make clear to check add-ons for which version of Plone


polyester:
we need to make clear that not all add-ons work on Plone 5 yet, and instruct people on how to check for that.

polyester:
also consider making a sticky note on community.plone.org for the moment

PR #422, opened 2015-11-02
unify add-on information

polyester:
contains a fix for https://github.com/plone/documentation/issues/416 
and initial inclusion of https://github.com/plone/documentation/issues/421

the add-on documentation was repeated in two places, now unified and updated with version checking information.


svx:
I will ignore this failing test, because the erros has nothing to do with docs :)

polyester:
"The current global buildout threat level is:   HIGH "

svx:
Yes ! I saw that, too ! That is why I am saying it has nothing to do with the docs :) It is just that buildout makes your tests kind of unreliable, but this is another story :)

PR #424, opened 2015-11-06
Plone 5 upgrade guide tips

gforcada:
I basically wanted to write it down somewhere.

Should I create a subsection within the upgrade guide for this small changes?

### DIFF
--- a/manage/upgrading/version_specific_migration/p4x_to_p5x_upgrade.rst
+++ b/manage/upgrading/version_specific_migration/p4x_to_p5x_upgrade.rst
@@ -620,3 +620,21 @@ Now it should look like this::
   # NEW in 5.0
   language_tool = api.portal.get_tool('portal_languages')
   language_tool.getDefaultLanguage()
+
+
+Tests changes
+=============
+
+In Plone 4.x a date or date time widget used to be rendered as a set of input fields::
+
+  # OLD 4.x approach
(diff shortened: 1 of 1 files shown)


gforcada:
@svx @polyester any suggestion on the formatting? I would like to prettify this so it can be merged before it gets too outdated/I forgot what was all about.

btw if you don't mind I created a @plone/documentation-team so anyone can get the attention of the docs team without having to ping any individual alone, add all other team members though, I just added you two.

svx:
Have a look at http://sphinx-doc.org/markup/para.html 

I use for this kind of stuff:
- .. note::
- .. warning::
- .. versionadded::
- .. versionchanged::
- .. deprecated::
- .. seealso

:)

polyester:
I've done some quick-and-dirty formatting

svx:
Yes, that is already a begin ! But besides the code we should also format the rest properly :)

For example in line 678 a ..note:: would make sense

gforcada:
@plone/documentation-team I reformatted it a bit more and now it does not look so bad anymore. Probably the first section which are side effects of one PLIP it should be moved to the top once we have the proper LinguaPlone -> plone.app.multilingual docs as well.

gforcada:
@plone/documentation-team anyone up to review and possibly merge? :-)

PR #429, opened 2015-11-12
Update to documentation for Plone 5

danboggins:
Second attempt at a pull request


svx:
I would remove the Plone 4 comment and instead add this to the docs branch of Plone 4, this also keeps the docs cleaner and with a better separation between Plone 4 and 5

danboggins:
Removed the Plone 4 line @svx, I'll take a look for the Plone 4 docs branch now.

Issue #430, opened 2015-11-16
gzip option has been removed from plone.app.caching


tkimnguyen:
- [x] update doc at http://docs.plone.org/manage/deploying/stack.html 
- [x] update docs for Nginx and Apache

tkimnguyen:
closed in favour of #431

PR #431, opened 2015-11-17
gzip setting has been removed from plone.app.caching

tkimnguyen:
for https://github.com/plone/documentation/issues/430

### DIFF
--- a/manage/troubleshooting/buildout.rst
+++ b/manage/troubleshooting/buildout.rst
@@ -528,3 +528,15 @@ Buildout download cache is corrupted. Run ``bin/buildout -vvv`` for more info. T
       # Corrupted .tar.gz download
       rm /Users/mikko/code/buildout-cache/downloads/dist/lxml-2.3.6.tar.gz
 
+
+Mac OS X Error: Couldn't install: lxml 3.4.4
+============================================
+
+...using the Plone 5.0 unified installer on Mac OS X 10.11.1 El Capitan
+
+As per http://stackoverflow.com/questions/19548011/cannot-install-lxml-on-mac-os-x-10-9 you should update your Xcode command line tools::
+
+    xcode-select --install
+
+then re-run ``bin/buildout``. 
+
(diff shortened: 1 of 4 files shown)


svx:
In:
 /manage/deploying/front-end/nginx.rst

I would remove:
- http://httpd.apache.org/docs/2.2/mod/mod_deflate.html

You have this link already in the apache docs and having this under nginx could be confusing

tkimnguyen:
Thanks for catching that - fixed.

manage/troubleshooting/buildout.rst
```diff
@@ -528,3 +528,15 @@ Buildout download cache is corrupted. Run ``bin/buildout -vvv`` for more info. T
...
+============================================
+
+...using the Plone 5.0 unified installer on Mac OS X 10.11.1 El Capitan
+
+As per http://stackoverflow.com/questions/19548011/cannot-install-lxml-on-mac-os-x-10-9 you should update your Xcode command line tools::
+
+    xcode-select --install
+
+then re-run ``bin/buildout``. 
+
```
gforcada, line 542: This chunk seems totally unrelated to gzip, no idea if it's on purpose or by accident, just asking...

svx:
@gforcada I guess so too, and this makes the title of this pr a bit confusing :)

On a further note I would change the syntax:

"...using"

should be replaced with:

```
.. note::
```

or something, so that it looks a bit better :) and instead of the full link the usage of proper rst formatting would be better :) 

Example:
External hyperlinks, like `Python <http://www.python.org/>`_.

tkimnguyen:
Ah yeah sorry I made the PR originally as per the subject line but other subsequent docs changes I was committing got automatically lumped in with this PR. How could I keep them separate?

tkimnguyen:
Going to close this PR and make a new one that doesn't include the commit about lxml.

tkimnguyen:
https://github.com/plone/documentation/pull/439 and https://github.com/plone/documentation/pull/440

PR #433, opened 2015-11-17
Do not use direct assignment to attribute in order to set default page

rafaelbco:
Assigning directly to the default_page attribute works, but it can mess with the property manager (getProperty, setProperty, etc) and then you'll be unable to change the default view using the Plone UI.


gforcada:
Looks good to me, merging.

PR #435, opened 2015-11-17
Degrokkify plone.directives.form

### LINKED ISSUE #434: plone.directives (grok alert?)
http://docs.plone.org/develop/plone/forms/schemas.html#multi-choice-example

This example does a `from plone.directives import form`, but `plone.directives.form` depends on `five.grok`.

My question:  is `five.grok` (and all the `grokcore` namespace packages) acceptable because it's not the _real Grok_, or did we miss a big chunk of degrokkification?

(See #296)

fulv:
Fixes #434.

### DIFF
--- a/develop/plone/forms/schemas.rst
+++ b/develop/plone/forms/schemas.rst
@@ -492,15 +492,16 @@ Each value is atomically constrained by *value_type* schema field.
 Example::
 
     from zope import schema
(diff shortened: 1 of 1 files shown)


svx:
Thanks @fulv !!

ale-rt:
This is nice indeed!

Issue #457, opened 2015-12-02
Should add plone-legacy js and css to custom Diazo based theme example


ivanteoh:
I noticed if i did not add plone-legacy js and css, some js are not found. It happened when I add collective.z3cform.datagridfield.

ivanteoh:
Not an issue. Should use @fulv PR #447 instead. It is also mentioned in https://github.com/plone/plonetheme.barceloneta/blob/master/plonetheme/barceloneta/theme/backend.xml comment.

```
<rules xmlns="http://namespaces.plone.org/diazo"
       xmlns:css="http://namespaces.plone.org/diazo/css"
       xmlns:xhtml="http://www.w3.org/1999/xhtml"
       xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
       xmlns:xi="http://www.w3.org/2001/XInclude">
    <xi:include href="++theme++barceloneta/backend.xml" />
    <rules css:if-content="body.viewpermission-view, body.viewpermission-none">
       <theme href="index.html" />
    </rules>
</rules>
```

PR #447, opened 2015-11-27
Fixes #64

fulv:
@sneridagh, @bloodbare, please weigh in.

PR #458, opened 2015-12-02
Add plone-legacy css and js to custom diazo rule. Fix for #457

ivanteoh:
Updated the example for #457 issue. I also think it is `head script[data-bundle='diazo']` and not `script link[data-bundle='diazo']`, although I did not see any scripts that have data-bundle='diazo'.

### DIFF
--- a/adapt-and-extend/theming/barceloneta.rst
+++ b/adapt-and-extend/theming/barceloneta.rst
@@ -140,11 +140,11 @@ You can develop a custom Diazo based theme and use the Barceloneta theme only fo
         <!-- Include basic plone/toolbar bundles -->
         <after
             css:theme-children="head"
-            css:content="head link[data-bundle='basic'], head link[data-bundle='plone'], head link[data-bundle='plone-logged-in'], head link[data-bundle='diazo']"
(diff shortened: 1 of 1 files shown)


fulv:
My PR #447 goes the opposite way.  Notice that if you include `++theme++barceloneta/backend.xml` the css, js and toolbar are already included.

fulv:
But yes, the `script link[data-bundle='diazo']` is clearly wrong.

svx:
@fulv @ivanteoh 

What is the status of this PR ? Do you know ? :)

ivanteoh:
@fulv is right. It is also mentioned in https://github.com/plone/plonetheme.barceloneta/blob/master/plonetheme/barceloneta/theme/backend.xml comment. @svx please ignore this pull. I will close this. Please use PR #447 instead. Thanks.

```
<rules xmlns="http://namespaces.plone.org/diazo"
       xmlns:css="http://namespaces.plone.org/diazo/css"
       xmlns:xhtml="http://www.w3.org/1999/xhtml"
       xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
       xmlns:xi="http://www.w3.org/2001/XInclude">
    <xi:include href="++theme++barceloneta/backend.xml" />
    <rules css:if-content="body.viewpermission-view, body.viewpermission-none">
       <theme href="index.html" />
    </rules>
</rules>
```

PR #459, opened 2015-12-02
Remove RPM - docs are based on Ubuntu

svx:
- Remove RPM, docs are base on Ubuntu, we do not want to confuse user

### DIFF
--- a/manage/deploying/caching/varnish3.rst
+++ b/manage/deploying/caching/varnish3.rst
@@ -24,6 +24,7 @@ To use Varnish with Plone
 
 * Add Plone virtual hosting rule to the default varnish configuration
 
+
 .. note ::
 
     Some of these examples were written for Varnish 2.x.
@@ -34,7 +35,7 @@ Installation
 
 The suggest method to install Varnish is to use your OS package manager.
 
-* You can install using packages (RPM/DEB) - consult your operating system instructions.
+* You can install using packages (DEB) - consult your operating system instructions.
 
 * For more up to date packages for Debian you could check: https://www.varnish-cache.org/installation/debian
 
--- a/manage/deploying/caching/varnish4.rst
+++ b/manage/deploying/caching/varnish4.rst
@@ -39,8 +39,6 @@ The suggest method to install Varnish is to use your OS package manager.
 
 * For more up to date packages for Debian you could check: https://www.varnish-cache.org/installation/debian
 
-* For more up to date packages for RedHat (RPM Based) you could check: https://www.varnish-cache.org/installation/redhat
-
 * You can install backports
 
 * You can install using buildout
@@ -53,7 +51,7 @@ Management console
 ==================
 
 ``varnishadm``
---------------------------------------------
+--------------
 
 You can access Varnish admin console on your server by::
 
--- a/manage/deploying/caching/varnish4.rst
+++ b/manage/deploying/caching/varnish4.rst
@@ -35,7 +35,7 @@ Installation
 
 The suggest method to install Varnish is to use your OS package manager.
 
-* You can install using packages (RPM/DEB) - consult your operating system instructions.
+* You can install using packages (DEB) - consult your operating system instructions.
 
 * For more up to date packages for Debian you could check: https://www.varnish-cache.org/installation/debian
 


polyester:
Needs discussion, I remember smcmahon adding stuff to the installer docs under the clear motto "treat RPM based distributions as first class citizen".

We do not want to scare away non-DEB based users, removing all mention of that may have that unintended effect instead of 'not confusing'....

gforcada:
As I was pointing on https://github.com/plone/documentation/issues/461 if we are directing users towards ansible and such, then it does not matter rpm/deb and what not, only the package names...

(my third cent :-)

svx:
ahem no really....

the only thing that we have in the installer docs are the basics dependencies to install plone on various linux platforms, we removed on other places of the docs already the mentions of rpm, osx and further we name this only is they are really needed for them.

We clearly stated also in: http://docs.plone.org/manage/installing/installation.html#how-to-install-plone

that the docs are based on ubuntu .....

svx:
@gforcada but that is for the full stack not if you have already plone installed, telling ppl that you 'have' to use ansible is IMHO not OK

svx:
on a further note:

@polyester that would mean to add not only rpm but also other package formats, to avoid that we decided a long time ago, to base the docs on Ubuntu LTS.
I removed it also because we do not mention it elsewhere in the docs, not in the nginx or apache part or somewhere else, everything is ubuntu/buildout.
Having it than only under varnish confuses ppl because we do not have it elsewhere ....

So yes, maybe that is not perfect but at least it is more consistent with the rest of the docs !

polyester:
@svx but it is inconsistent here also; in varnish4.rst you want to remove RPM, leave DEB, but then two lines later it says "For more up to date packages for RedHat (RPM Based) you could check: https://www.varnish-cache.org/installation/redhat"

Which I think is perfectly fine, and actually a help to the user. So I would propose to re-write it 'distribution-neutral' where possible, but with links that help the user. So, remove **both** (RPM/DEB) so that it says "preferebly, install with , but leave the links to the varnish docs.

So something like

The suggest method to install Varnish is to use your OS package manager - consult your operating system instructions.
- For more up to date packages for Debian/Ubuntu, Redhat, BSD and others check: https://www.varnish-cache.org/docs 
- You can install using buildout

(UPDATED: link should be https://www.varnish-cache.org/docs, not https://www.varnish-cache.org/install as that doesn't work on it's own)

svx:
@polyester like i said before this is not consistent with all the other docs, and we made this decision long time before you were even on the docs team, if you want to re-discuss fine, but this should be a discussion in general and not only about one tiny part of the docs.

So for now I go with the old decision because that is on what the whole docs are based atm

svx:
OK, I am merging that, because that is the right thing to do.

Reason:

This change follows the current setup of the docs. meaning based on Ubuntu LTS, and for this it actually improves the docs, because it is removing a possible confusing note. 
This makes the docs more consistent !

@polyester if you think basing the docs on Ubuntu should be discussed again, please open a discussion for that !
But this discussion should include the whole documentation and not like I already said above, one tiny part, mention different package formats on different random places is confusing.
So for this I am merging, because for now this PR improves the docs !

polyester:
@svx I did not want to discuss basing the docs on Ubuntu LTS, but there is a difference between
- try to write the docs distribution-neutral where possible, when distribution-specific things are mentioned base them on Ubuntu LTS 

and
- not mentioning other OS's at all

In this case, I think my suggestion would have achieved that better. But since I'm a newcomer, whatever.

svx:
@polyester still this needs to be a discussion about the whole docs in general.

That is why I stated open up a discussion.

Just changing one tiny part and have this part different will not help at all, so yes I stand to my merge because it follows what we have **currently**

svx:
@polyester 
and btw:

<quote>
For more up to date packages for Debian/Ubuntu, Redhat, BSD
</quote>

This is from your suggestion but based on that I would like to ask why you just mention these ones, what is with Fedora server, what is with Arch, Gentoo, Nix or other ones ?

In this case I think you suggestion was not the better way, sorry

It may be a valid point to discuss to change the whole docs to be more 'OS neutral' but please do that properly !

There are many different points which you have to think about, but this should be done first,

thet:
+1 for distribution neutrality where possible.

saying "install using packages (RPM/DEB)" sounds like these are the only supported package formats. this is even stronger when removing the RPM part. mentioning a package format adds not much information anyways in this context.

i'm currently a fedora, thus RPM user.

svx:
I am not against it, but I am saying first talk about that in a own ticket or on community, please :)

This has impact in lots of parts of the docs, and for this we need to think how to solve that in a proper way, starting from installer, lost of examples and so on and so on.

We need to find good solutions for that first, it is IMHO not the best idea to start change  one tiny part of the docs, before thinking about the docs as the 'whole' docs and all possible challenges first.

thet:
I basically agree - I don't think we should have distribution neutrality for installation documentation - IIRC we even have a section for vagrant for people don't have the discussed distribution or are on windows.
But where it's unspecific anyways, we could stay distribution neutral.
+1 for moving this discussion, if there is need for further talk.

svx:
https://community.plone.org/t/should-be-the-docs-distribution-neutrality/1285

PR #473, opened 2015-12-10
restructure: testing and debugging doc from deployment to dev, cachin…

thet:
…g to caching. Fixes: #468

PR #474, opened 2015-12-11
fix links

polyester:
various links had to be changed due to https://github.com/plone/documentation/pull/473


thet:
Ups-totally forgot to check links. Sorry for that!

PR #476, opened 2015-12-14
commit every 100 transactions

tkimnguyen:
correction to if condition


thet:
thanks for the correction!

PR #483, opened 2016-01-02
Remove redundancy

gforcada:
Currently when going to http://docs.plone.org/develop/index.html the left column looks like this:

toc

There's quite a lot of duplication, I'm not sure where to cut it from to make it more meaningful, either on the subfolder titles or on the parent index.

I thought that it make more sense to remove it on the parent so that subfolders are a complete unit without need of external information.

Not sure how does it look like once the titles are removed thought.

This is mostly WIP and open to discussion.

Happy new year!

### DIFF
--- a/develop/index.rst
+++ b/develop/index.rst
@@ -2,25 +2,15 @@ Developing for Plone
 ====================
 
 
-Developing add-ons
-------------------
-
 .. toctree::
    :maxdepth: 2
 
    addons/index
 
-Programming with Plone
-----------------------
-
 .. toctree::
    :maxdepth: 2
 
    plone/index
-
-
-Debugging Plone
----------------
 
 .. toctree::
    :maxdepth: 2
@@ -28,17 +18,11 @@ Debugging Plone
    debugging/index
 
 
-Testing Plone
--------------
-
 .. toctree::
    :maxdepth: 2
 
    testing/index
 
-
-Developing for Plone Core
--------------------------
 
 .. toctree::
    :maxdepth: 4


gforcada:
@plone/documentation-team could any of you provide some feedback? :-)

polyester:
it would probably be better to at the same time change the spelling/wording of the 'lower' ones.

So:
- Develop Plone Add ons -> Developing Plone add-ons
- Programming Plone -> Programming with Plone
- Process for Plone core's development -> Developing for Plone Core
- Styleguides -> Plone Style Guides

(having sensible wording was the reason I put extra headings in the first place, but indeed it's better to change the wording on the underlying title, to avoid duplication)

thet:
+1 on both suggestions.

@gforcada please add the styleguide and importing content from other systems too (last one might be even a better fit in  the "manage" section instead of "develop"?)

But the first entry will then look not strong enough.
@gforcada @plone/documentation-team is there a way to make the first entry look like a heading? Another approach would be to keep the headings in this document, but exclude the first entry with some toctree option: http://www.sphinx-doc.org/en/stable/markup/toctree.html

polyester:
@thet, oh yes I do mean keep the 'styleguide' and 'importing content from other systems' as well, but their titles don't have to be changed.

svx:
@gforcada @polyester what is the status of this ?

gforcada:
@svx I guess titles need to be adapted I think, but I have my hands full enough already

svx:
np ! I just thought I ask :)

svx:
This is fixed by: #580

Issue #491, opened 2016-01-21
Remove mention of PIL update install docs


svx:
Should we update: http://docs.plone.org/develop/plone/images/ ?
- removing PIL
- update Pillow docs, mention about installation of certain packages: libjpeg62-dev libfreetype6 [Ubuntu] and libjpeg-turbo-devel freetype-devel [Fedora/CentOS]

We already have  libjpeg62-dev mentioned in: http://docs.plone.org/manage/installing/installation, should we update that with libfreetype6, too ?

polyester:
+1 on removing PIL, I would probably suggest having the install instruction on Pillow relatively simple (although including libfreetype6 might be a candidate) and have it point to https://pillow.readthedocs.org/en/latest/installation.html for all the details, which is quite up to date, with specific instructions for various operating systems.

svx:
Please review https://github.com/plone/documentation/tree/fix_%23491
- rename pil to pillow
- update pillow.rst
- update installation.rst with mention about pillow dependencies
- update requirements with note about pillow dependencies

The wording is not perfect yet, but IMHO it fixes the confusion and also makes clear [ I hope] that the user may need to install some dependencies for pillow which are not dependencies for plone itself.
@polyester, @smcmahon, @do3cc please review and tell if that is OK, please do not merge immediately, I want to make sure that we do it right this time :)

smcmahon:
How about these edits to the particular sections of pillow.rst:

Plone uses Pillow, a fork of the Python Imaging Library (PIL) for low-level image manipulation and decoding, particularly for GIF, JPEG and PNG image formats.

...

# Installing Pillow

For instructions how to install Pillow and its dependencies, please read:

https://pillow.readthedocs.org/en/latest/installation.html

Note particularly the dependencies on the development packages of libjpeg and libz. Other image format support may be required for special projects, but libjpeg and libz are required for GIF, JPEG and PNG support.

svx:
Awesome, will fix that later tonight ! Thanks a lot !

svx:
Done !

PR #493, opened 2016-01-21
Feat(install) Add Freetype dependency

do3cc:
For some reason the docs did not mention the freetype dep
Also, sorted the changed lists alphabetically
https://community.plone.org/t/pillow-1-7-8-not-installing-creating-a-problem/1447

### DIFF
--- a/manage/deploying/preparing.rst
+++ b/manage/deploying/preparing.rst
@@ -58,11 +58,12 @@ System libraries
 
 For any install, the development versions of:
 
-* libssl
-* libz
+* libfreetype6
 * libjpeg
-* readline
+* libssl
 * libxml2/libxslt
+* libz
+* readline
 
 If you're using the System Python, add:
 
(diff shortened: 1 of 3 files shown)


svx:
This is also discussed on #491 

Also this 'only' updates the Debian/Ubuntu part in: /prepare we should update this for other OS's mentioned on the site too

svx:
@do3cc thanks for your work on that !

svx:
@polyester why did you merged that ? AT least /prepare should be updatet in the Centos/Fedroa section **before** it was got to go.

svx:
oh and also suse and the other ones we mention in that section !

svx:
@polyester please revert or send a new PR to fix that, the last merge left the docs in a [IMHO] inconsistent and thus broken state.

do3cc:
@svx in this case a force push of the commit before this might have been a better solution, now I need to make a whole new PR :-( But I agree on the non merge

polyester:
whoops sorry, I just created a new pull request that includes freetype(x)-(dev|devel) for the various distros, but that does not include the original changes somehow.

svx:
@do3cc no need for that, I already did it, just waiting for travis :)

svx:
@polyester all this is fixed in my PR

PR #494, opened 2016-01-21
Feat(install) Add Freetype dependency

do3cc:
For some reason the docs did not mention the freetype dep
Also, sorted the changed lists alphabetically
https://community.plone.org/t/pillow-1-7-8-not-installing-creating-a-problem/1447


do3cc:
See also #493

PR #498, opened 2016-01-21
fix freetype and different OS

svx:
This should fix it
- update library info
- update OSX part
- update install part
- fix rst syntax

### DIFF
--- a/manage/deploying/preparing.rst
+++ b/manage/deploying/preparing.rst
@@ -63,6 +63,7 @@ For any install, the development versions of:
 * libjpeg
 * readline
 * libxml2/libxslt
+* freetype
 
 If you're using the System Python, add:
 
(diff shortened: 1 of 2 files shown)


svx:
The branch for 4 still needs fixing

smcmahon:
As far as I know, Pillow supports freetype, but has no freetype dependency. AFAIK, Plone gains nothing from the availability of freetype.

Stop the madness! Let's not create dependencies out of thin air. Let's not encourage installation of packages that add no functionality. Every additional package is an added maintenance/security burden (though perhaps slight).

-1 for merge.

svx:
@smcmahon and all

Hi,

please can we try, to not over-rush things, currently this is just a PR, it is not merged yet, and I reverted the merge from @polyester already yesterday.

Currently we do not mention that in the documentation.

Reading the docs of Pillow [https://pillow.readthedocs.org/en/3.1.x/installation.html] the clearly saying
Under buillding
Ubuntu 14

```
$ sudo apt-get install libtiff5-dev libjpeg8-dev zlib1g-dev \
    libfreetype6-dev liblcms2-dev libwebp-dev tcl8.6-dev tk8.6-dev python-tk
```

and for Fedora

```
$ sudo yum install libtiff-devel libjpeg-devel libzip-devel freetype-devel \
    lcms2-devel libwebp-devel tcl-devel tk-devel
```

That is why I want to evaluate my first idea in #491.

We do not out it under requirements for Plone, but we update out own docs about Pillow, linking to the original Pillow docs with some short explanation.

Thank you all for your comments, ideas and input !

smcmahon:
Sven, thanks for researching it, and sorry if my comment sounded panicky.

jensens:
what about using labels?

svx:
old and already fixed

PR #500, opened 2016-01-25
Update styleguide with regards to import sorting

do3cc:
This is the end after a long discussion from here:
https://github.com/plone/plone.api/issues/187

### DIFF
--- a/develop/styleguide/python.rst
+++ b/develop/styleguide/python.rst
@@ -310,7 +310,7 @@ Grouping and sorting
 --------------------
 
 Since Plone has such a huge code base, we don't want to lose developer time figuring out into which group some import goes (standard lib?, external package?, etc.).
-So we just sort everything alphabetically, case insensitive and insert one blank line between ``from foo import bar`` and ``import baz`` blocks.
+So we just sort everything alphabetically, case sensitive and insert one blank line between ``from foo import bar`` and ``import baz`` blocks.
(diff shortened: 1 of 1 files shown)


do3cc:
Oh, for the record, the discussion did not end in universal agreement and no formal voting.
It looks like the majority of people are in favor of the pull request in the discussion. But more importantly the arguments for case insensitive sorting weren't refuted while the arguments for case sensitive sorting were.

develop/styleguide/python.rst
```diff
@@ -309,23 +309,21 @@ About imports
...
 
     # GOOD
     from __future__ import division
     from Acquisition import aq_inner
+    from Products.CMFCore.WorkflowCore import WorkflowException
+    from Products.CMFCore.interfaces import ISiteRoot
     from plone.api import portal
     from plone.api.exc import MissingParameterError
-    from Products.CMFCore.interfaces import ISiteRoot
-    from Products.CMFCore.WorkflowCore import WorkflowException
```
gforcada, line 328: Aren't you sorting here as case sensitive while above saying it should  be case insensitive? :-) so the example was already ok.

do3cc, line 328: Got me, I always mix it up. But if you followed the discussion, you know what I meant ;-)

do3cc:
I'll drop this one

PR #505, opened 2016-01-28
Add info on docker installation

demarant:
Almost any new modern CMS or system offers now a docker image. So I added here an example of docker-based installation we have so far.

### DIFF
--- a/manage/installing/installation.rst
+++ b/manage/installing/installation.rst
@@ -316,6 +316,21 @@ LibXML2/LibXSLT Versions
 
 Don't worry about this if you're using an installer.
 
+Docker-based installation (platform independent)
+------------------------------------------------
+
+Prerequisites: `Docke engine <https://docs.docker.com/engine/installation/>` installed on your host(s).
+
+There is still no official Plone docker Image however you can use the community-based `eeacms/plone <https://hub.docker.com/r/eeacms/plone/>` which is flexible enough to be configured for all kind of setups, even cluster multi-host setups. It automates and facilitate the entire installation process via single commands, since the docker image is prebuild.
+
+For a basic installation:
+
+$ docker run -p 8080:8080 eeacms/plone
+
+The above will first download the images (first time) and than exposing plone on the host port 8080. Now go to http://<yourserverip>:8080 to see Plone in action.
+
+For more information and configuration options see `README file on guthub <https://github.com/eea/eea.docker.plone>`.
+
 Entering debug mode after installation
 ======================================
 
--- a/manage/installing/installation.rst
+++ b/manage/installing/installation.rst
@@ -321,7 +321,7 @@ Docker-based installation (platform independent)
 
 Prerequisites: `Docke engine <https://docs.docker.com/engine/installation/>` installed on your host(s).
 
-There is still no official Plone docker Image however you can use the community-based `eeacms/plone <https://hub.docker.com/r/eeacms/plone/>` which is flexible enough to be configured for all kind of setups, even cluster multi-host setups. It automates and facilitate the entire installation process via single commands, since the docker image is prebuild.
+There is still no official Plone docker Image however you can use the community-based `eeacms/plone <https://hub.docker.com/r/eeacms/plone/>`_ which is flexible enough to be configured for all kind of setups, even cluster multi-host setups. It automates and facilitate the entire installation process via single commands, since the docker image is prebuild.
 
 For a basic installation:
 
@@ -329,7 +329,7 @@ $ docker run -p 8080:8080 eeacms/plone
 
 The above will first download the images (first time) and than exposing plone on the host port 8080. Now go to http://<yourserverip>:8080 to see Plone in action.
 
-For more information and configuration options see `README file on guthub <https://github.com/eea/eea.docker.plone>`.
+For more information and configuration options see `README file on github <https://github.com/eea/eea.docker.plone>`_.
 
 Entering debug mode after installation
 ======================================


svx:
Thanks for that, but :)
we have a docker image which is in early testing state, some ppl are already using it, but it is still WIP, we still have to decide about certain things and polish it a bit :)

Second, the thing is that the eea image is kind of different from what we have, starting for example with the OS which is used and the setup.

As soon as we decide that the official plone docker image is stable enough we will pronounce is more into the public and we will also update the docs.

But this will take some more time, because we also want to make sure that we have better docs about that.

demarant:
@svx 
exactly why I added the eeacms image. we have been waiting already 1 year for the official one ;)

demarant:
@svx:
btw EEA is a strong contributor to Plone I believe, full part of this community, so I am not sure what "official" means, since there is no big large company behind Plone, I am a bit confused on how the decision making regarding this is done.  :)

svx:
@demarant 

https://hub.docker.com/r/plone/plone/

The eea image is not official it is the one of eea, we also do not link to all the other plone docker images there are lots of them and all of them are different.
I am also not linking to my own ones and there are reasons for that :)

It is the same with other installers, lots of ppl install plone in different ways, with different buildouts and so on, but we also try to stick with 'one official' with 'best practices' .

The reason for that is that is easier to help ppl and it is easier for documentation.

demarant:
@svx see my previous comment.
we could become the maintainer of the plone official docker image, if you understand me.

svx:
@demarant the community has a installer-team, if you or eea want to be part of this, feel free to join !

For the rest the my comments above, why your image and not the one from other community members ?

Do not get me wrong I get your point about docker, but for a good product and good docs there needs to be more done, than 'just'a link to one image.

demarant:
@svx: why the eea one? because it is simply the best production ready so far out there. 
We have 2 years docker experience and we run [most of our applications on docker on production](https://github.com/eea?utf8=%E2%9C%93&query=docker) for 1 year now.
If you guys want to catch-up on the docker train, well we can contribute to it easily.

svx:
Currently we do not link in the docs to 'third party' installer, we also do not link to bitnami or something else, if ppl want that, this is open for discussion.

The docs are based on the UI-Installer, currently. If you want to change that, please file a ticket, but be prepared that this will involve a lot of work and talking.

svx:
@demarant other ppl have this experience, too it is not about that but you build your images in your way, I a not saying that is bad, but like I said before everyone has another way of doing images.

So in any case this should be discussed anyway, before any further action and as for the docs, please see my last comment.

demarant:
@svx: i understand the choices on not linking to third party installers. However not having a single line on Docker in the documentation looks like we are in "Stone age". Maybe we could at least write smth that an official is on the making...

avoinea:
I think this can be safely added to http://docs.plone.org/quickstart/index.html#alternative-ways-to-deploy

svx:
@demarant @avoinea Sounds like a valid point, I will do that later today :)

demarant:
@svx: thank you! regarding the plone official image, I just realized you have a repo https://github.com/plone/plone.docker finally.  we are here to support that. we will comment and contribute on that one.

svx:
Great !

smcmahon:
Just to add a little to the discussion:

To me, "official" means that the installer team has committed to maintaining the install kit and the documentation team has committed to maintaining its documentation. Both of those teams change over time and with the participation of interested individuals. And, what's "official" changes over time.

I think "official" also means that a kit is generic Plone unless there's been some sort of general discussion about differences like add-ons included. We don't want to trap a future framework team into having to make plans around particular add-on compatibility because of what was in an installer kit or documentation.

An example, we just had a discussion about whether or not we should include freetype in the listed packages required/suggested for Plone installation in the documentation. I argued against it because it is does not add functionality for generic Plone and I didn't want anyone to think they could rely on freetype availability.

djay:
Would love to see eea join the effort for an offical docker image!!! That would be a win all around. We use our own docker images but the eea ones are well thought out and very nice. please join the installer team @demarant :)

But do think that not providing third party installer information in the documentation is a mistake, in particular where there is no offical solution such as cloud installers like bitnani or heroku. We should be providing all the easiest solutions for getting plone up, regardless of if its official, regardless if it points to businesses that might make money later down line IMO. At least in the short term until such time as there are complaints favouring one commercial solution over another. IMO cloud solutions should be offered, not just in documentation but as a top level option right next to the download plone button plone.org. This allows us to complete with the likes of wordpress.com or squarespace. You can see the kind of confusion in recent community posts caused by a "download plone to try it" approach. We offer no easy way to transition from that to hosted solution and most would just give up.

svx:
@djay I disagree at least partly :)

Firt of all such information should not go into the docs, if there is made the decision to tell about 'third party' installers they should go on a 'special page' on docs.plone.org download.

We have/had this discussion since a long time know, and yes I can see you point of that, and I agree partly, but there is also a possible downside/danger with that ?

For example do you ever took the time and try and check the bitnami installer ?

The setup is different, the buildout is different and so on, we base out docs on out [plone community] best practice and out installers. We had it a couple of times that ppl installed for example the bitnami one, and when they had problems/questions they were 'complaining' that  our docs are not covering that.

Well we do not cover bitnaimis setup.

Next point, some of these 'third party' installer are out for a year, and then they got forgotten and never updated, so someone of our community would spent time to run down, which of the installer we host on out site is still up to date.

We tried to speak with bitnami a couple of times for a better exchange on information and so on, but their were not really interested.

These are now only two short 'issues' I could go on for some time ... :)

So yeah, I guess I am trying to say, if the community decides to do that, this should be thought through there are different parts which should considered careful. :)

djay:
@svx True I can see your point. I guess docker is possibly the good way out of that problem as it could offer in the near future almost a one click cloud install on a variety of hosting providers. I don't think its there yet, but I could see the possibility to cut and paste a docker-compose into AWS/rackspace etc and have a working cluster but using docker images from the installer team so it would be easier for the community to support. Similar to http://blog.jazkarta.com/2014/06/03/scalable-python-application-deployments-on-aws-opsworks/. I gave this a go. It was more or less just cut and paste. But docker would be a way more portal way of doing that. IMO Vagrant images to download just aren't that useful as it doesn't solve the problem of how you host it. 
If we could get to a blessed solution where there is a single button each for AWS. azure, googlecloud, heroku, etc on https://plone.org/products/plone/releases/5.0, esp when there are free tiers, that would be a big win for plone (or a "cut and paste this" solution).

demarant:
@djay we are unofficially already part of the installer team by providing the best docker image so far ;)
we will make the eea plone docker images small, microservices-ready, secure and as configurable as possible for all possible use case scenarios :) this is our official goal. We are also planning to make the official Plone catalog entry for the [open source Rancher](http://rancher.com/rancher/), we already have something useful there (presented at the last plonecon 2015), one-click cloud install for all, independent of underlying IaaS and OS. Vagrant and the likes are just becoming obsolete technologies, although maybe still relevant in some legacy setups... it is time to move to the better.
So the future looks bright to us!

demarant:
@svx 
I will close this PR now, as it is not needed anymore. Some reference to docker installation can be added in the [alternative ways](http://docs.plone.org/quickstart/index.html#alternative-ways-to-deploy) as @avoinea suggested.

as agreed, EEA will contribute to the official plone docker image by moving/merging the work done in [eea.docker.plone](https://github.com/eea/eea.docker.plone) into the existing [plone.docker](https://github.com/plone/plone.docker). 

This will essentially create the first production ready official plone docker image and at that point we will do a new PR to update the documentation. Further Issues and discussions will take place on the [specific repo](https://github.com/plone/plone.docker). We are open to any suggestion/contributions for improvements of course.

In the longer term, we should also consider to have the plone docker image submitted to the [official   docker images](https://github.com/docker-library) if found necessary.

thanks

PR #506, opened 2016-01-29
Replace tutorial link with a current working one

staeff:
The link to the heroku-tutorial does not work anymore. Replaced it with a slightly different solution, which provides similar results.

### DIFF
--- a/quickstart/index.rst
+++ b/quickstart/index.rst
@@ -13,8 +13,7 @@ Online demo sites
 If you're just curious to see how Plone looks, without any modifications, there are a few online demo sites that can help you.
 Both `plone5demo.quintagroup.com <http://plone5demo.quintagroup.com/>`_ and `plone5.veit-schiele.de/ <https://plone5.veit-schiele.de/>`_ provide you with a Plone 5 site that is re-set every day.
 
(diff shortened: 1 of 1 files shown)


svx:
@staeff Thanks for your effort !!

That has now nothing to do with your work, so please do not be offended ! I just saw this now, I am in general not sure if that is the right place for it.

Playing with a demo is different than telling ppl:

> nother interesting approach to try out Plone is to install it in a cloud service. That way, you can set up your own demo (or even very-light-weight production)

We are talking here about installation

> (or even very-light-weight production)

I am feeling not comfortable at all to place stuff like that here.

Basically that means we are talking about install and again like in https://github.com/plone/documentation/pull/505 we are talking about 'third party' the heroku and the cloud 9 stuff is slightly different.

We should try to find a sane way, how we want to handle that, we are starting already to mention various different install/production setups in different places of the docs, this is not ideal !

@smcmahon what do you think ?

Should we create/extends ONE place in the docs to mention all these ones, or should we do it under plone.org/download ?

I would like to fix that once and for all :)

I can see the point that ppl want to see that, but also I completely agree with your comment https://github.com/plone/documentation/pull/505 .

staeff:
Hey @svx I am by no means offended. I was just changing or updating what was already there. So no problem with a decision that might remove or replace this content.

jensens:
@svx I'am +1 merging this and then lets create a new ticket in order to not loose the task. Otherwise we wont improve if we try perfectionism from every single change on - its also a lot stop energy in this.

svx:
Merged for now as temp. workaround, this will be re-located in the future

svx:
see #527

PR #509, opened 2016-01-29
Update styleguide regarding sorting imports

gforcada:
Follow up of #500  (nice number one @do3cc ) and https://github.com/plone/plone.api/issues/187.

At [Alpine City Sprint](http://www.coactivate.org/projects/alpinecitysprint-2016/project-home) @jensens @do3cc and me discussed about the issue and end up deciding that tooling is more important than editors and thus, we follow suit what [isort](http://pypi.python.org/pypi/isort) provides us, i.e. alphabetically case-insensitive sorting and one line between `from X import Y` and `import Z`.

My personal opinion to whoever wants to change it: update [isort](https://github.com/timothycrosley/isort) so that it sorts how you like and **only after that** we can open, yet again, the discussion.

As a matter of fact, currently our stack compromises +8500 python files, who is gonna update them manually without a tool to do so and keep them checked?

### DIFF
--- a/develop/styleguide/python.rst
+++ b/develop/styleguide/python.rst
@@ -311,7 +311,7 @@ Grouping and sorting
 
 Since Plone has such a huge code base,
 we don't want to lose developer time figuring out into which group some import goes (standard lib?, external package?, etc.).
-So we just sort everything alphabetically and insert one blank line between ``from foo import bar`` and ``import baz`` blocks. Conditional imports come last. Again, we *do not* distinguish between what is standard lib,
+So we just sort everything alphabetically case insensitive and insert one blank line between ``from foo import bar`` and ``import baz`` blocks. Conditional imports come last. Again, we *do not* distinguish between what is standard lib,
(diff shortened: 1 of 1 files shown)


thet:
+1. isort works great in vim.
your provided isort configuration also expands multiline imports to one line per import. I'll create a pull requests with an update mentioning the multiline import thing.

do3cc:
I have mixed results with the Isort plugin.
This one: https://github.com/fisadev/vim-isort#installation does not work sometimes.
I never figured out why, but I suspect that vim caches the python interpreter to use and if you change your virtualenv, you might end up in a virtualenv where isort is installed, but your vim uses a cached python without isort installed.
For a month now I use the second suggestion from the isort wiki:
command! -range=% Isort :<line1>,<line2>! isort - - just add this to your ~/.vimrc
and this works.

thet:
I also use the second suggestion and mapped it to a keymap. Since this is all I expect from a vim integration of isort, I prefer this approach.

PR #510, opened 2016-02-01
statement about multi-line import

thet:
@gforcada please review
especially this commit: https://github.com/plone/documentation/commit/fb9c20ba00d37c07a2035510603d6c55cdbd2b93
the first commit is only reformating lines to have one line per sentence and two empty lines between sections in all styleguide files.

### DIFF
--- a/develop/styleguide/python.rst
+++ b/develop/styleguide/python.rst
@@ -6,8 +6,7 @@ Python styleguide
 Introduction
 ==============
 
-We've modeled the following rules and recommendations based on the following
-documents:
+We've modeled the following rules and recommendations based on the following documents:
 
  * `PEP8 <http://www.python.org/dev/peps/pep-0008>`__
  * `PEP257 <http://www.python.org/dev/peps/pep-0257>`_
@@ -15,6 +14,7 @@ documents:
  * `Google Style Guide <http://google-styleguide.googlecode.com/svn/trunk/pyguide.html>`_
  * `Pylons Coding Style <http://docs.pylonsproject.org/en/latest/community/codestyle.html>`_
  * `Tim Pope on Git commit messages <http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html>`__
+
 
 Line length
 ===========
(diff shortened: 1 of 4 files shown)


develop/styleguide/python.rst
```diff
@@ -66,11 +65,16 @@ Based on code we love to look at (Pyramid, Requests, etc.), we allow the followi
...
- * The same style applies to dicts, lists, return calls, etc.
+* Arguments on first line, directly after the opening parenthesis are forbidden when breaking lines.
+
+* The last argument line needs to have a trailing comma (to be nice to the next developer coming in to add something as an argument and minimize VCS diffs in these cases).
+
+* The closing parenthesis or bracket needs to have the same indentation level as the first line.
+
+* Each line can only contain a single argument.
+
+* The same style applies to dicts, lists, return calls, etc.
```
gforcada, line 76: There's no need to add an empty line between them, if not instead of creating a single list with X items you are creating X lists with only one item on each

thet, line 76: Most of the listings have an extra line in between.
And it helps with reformating - one line per sentence, if you have an editor command to do that.

gforcada:
@thet looks good, although the line breaks should be not only at full stops but also at commas as well: http://docs.plone.org/about/rst-styleguide.html#line-length-translations

polyester:
@gforcada, 
do not take that statement about commas too serious,
you know,
because,
well,
it can be annoying as well. 
;-)

(in other words: only when there's really long sentences. For short parts of sentence, having commas is fine.

gforcada:
ok,
I will be more relaxed about it :)

thet:
@gforcada plz re-review and merge if appropriate. this is the content-change: https://github.com/plone/documentation/commit/fb9c20ba00d37c07a2035510603d6c55cdbd2b93
the others are only line formatings.

gforcada:
@thet actually that's already stated a few lines above [About imports, second point](http://docs.plone.org/develop/styleguide/python.html#about-imports).

Although restating doesn't hurt I guess.

thet:
Oh! I oversaw that.
But ok, restating doesn't hurt here.

PR #511, opened 2016-02-01
Fix warning

gforcada:
And headers on the way.


gforcada:
@svx that was quick! thanks!

PR #482, opened 2015-12-21
improve resource registry documentation

frisi:
add an example of how to build an optimized bundle that combines resources of different addons
(see https://github.com/plone/Products.CMFPlone/issues/1163 for a discussion)

WIP - see questions marked with XXX

PR #518, opened 2016-02-09
Resource Registry documentation improvements

cewing:
The documentation of the new Plone 5 resource registry is tough to understand.  This PR aims to improve the documentation.  

This is a Work In Progress and should not be merged until it has been completed.  I am looking specifically for fact-checking and input from @thet, @frapell, @bloodbare, and @ebrehault for Resource Registry concepts and instructions,  and from @svx and @polyester for style.

### DIFF
--- a/adapt-and-extend/theming/resourceregistry.rst
+++ b/adapt-and-extend/theming/resourceregistry.rst
@@ -4,11 +4,11 @@ Plone 5 Resource Registry
 
 .. admonition:: Description
 
-    For JavaScript and CSS development, Plone 5 makes extensive use of tools like Bower, Grunt, RequireJS and Less for an optimized development process.
-    The JavaScript and CSS resources are managed in the Resource Registry.
-    The Resource Registry was completly rewritten in Plone 5 to support the new dependency based RequireJS approach.
-    It also allows us to build Less and RequireJS bundles Through-The-Web for a low entry barrier.
-    This chapter should give you a good understanding of the new Resource Registry.
+    Plone 5 has modernized the JavaScript and CSS development experience by incorporating tools like Bower, Grunt, RequireJS and Less.
+    JavaScript and CSS resources for core Plone and add-on packages are managed in the new Plone 5 Resource Registry.
+    The Resource Registry was completely rewritten in Plone 5 to support a new dependency-based approach built on RequireJS.
+    It also allows developers to create JavaScript and CSS bundles Through-The-Web for a low barrier to entry.
+    This chapter will help you to gain a basic understanding of the new Resource Registry.
 
 .. contents:: :local:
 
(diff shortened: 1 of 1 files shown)


frisi:
@cewing  did you take a look at #482 (which was my attempt of explaining the process of defining an own bundle)?

please take the time to read my changes and if there is anything that could be taken over to your branch - please do so.
after that, my PR can be closed ;-)

cewing:
@frisi I'll look that over.  In talking to @bloodbare it seems that perhaps it's better for the time being for us not to suggest defining our own bundles to replace the Plone bundle.  My impression is that the process is still unsettled and potentially dangerous. Until we have a good story I'd rather leave it out of the docs entirely.  

Any opinion on that?

frisi:
the example i added is not about overwriting plone and plone-logged-in but create a separate 
mysite-plone and mysite-plone-logged-in.

the whole idea was (plip1277 being non-existent) to not end up with 5 different bundles (= 5 requests), just because you installed 3 addons.
addons need to ship a separate bundle to work out of the box w/o compiling resources manually.
even worse: different bundles can possibly contain the same resources (such as jquery, select2 or other stuff they list as required dependencies).

adapt-and-extend/theming/resourceregistry.rst
```diff
@@ -124,71 +241,233 @@ In ``mockup/patterns/structure/js/views/actionmenu.js``:
...
 
     <plone:static
         directory="static"
         type="plone"
-        name="static"
+        name="myresources"
         />
 
-Now we can access the contents within the "static" folder by using the URL part ``++plone++myresources/`` and appending the path to the resource under "static".
+Now we can access the contents of the "static" folder in your pakcage by using a URL that starts with ``++plone++myresources/``.
```
seanupton: s/pakcage/package

adapt-and-extend/theming/resourceregistry.rst
```diff
@@ -124,71 +241,233 @@ In ``mockup/patterns/structure/js/views/actionmenu.js``:
...
+
+In development mode, each bundle loads all of its resources in the rendered site as individually, with individual requests.
+Moreover, in development mode modifications to resources are immediately available.
+You do not need to compile any bundles beforehand.
+You should be aware that this feature can lead to a lot of requests and slow response times, even though RequireJS loads dependencies asynchronously.
+
+In production environments you will compile your bundles to combine and minify all the necessary resources into a single JavaScript and CSS file.
+Since the dependencies of each resource in the bundle are all now well-defined, they can all be included in these files.
+Compiling bundles minimizes the number of web requests and the payload of data sent over the network.
+In Production mode, only one or two files are included in the output: a JavaScript and a CSS file.
```
seanupton: ...included _, for each bundle,_ in the output... ?

cewing: very good point.  Will clarify

adapt-and-extend/theming/resourceregistry.rst
cewing: thanks, will fix.

cewing:
@frisi: having looked at your changes, I am hesitant to put them in place until some of the open questions in them have been answered.  If we do have a solid, tested, and functional story for creating a bundle that could be used _in place of_ the plone and plone-logged-in bundles (as I believe your changes would point to), then we should include documentation for how to do that.

However, at this point it seems to me that the story is not solid, has yet to be tested, and may result in some fragility, as site owners would have to re-check their substitute bundles for accuracy and completeness after any upgrade to Plone.

I think the changes in [plip1277](/plone/Products.CMFPlone#1300) probably represent a better, more flexible and safer route forward than suggesting that Plone admins or integrators create their own custom bundles intended to replace `plone` and `plone-logged-in`.

Any thoughts on this, @ebrehault or @bloodbare?

cewing:
@bloodbare, @ebrehault, @svx, @frisi and any other interested parties.

I believe that my work here is ready for review.

I have also updated the `plip-1277-meta-bundles-generation` branch of this same repository with some language improvements that match better with the changes I've made here.  Merging this PR with the changes made there will require a bit of conflict resolution, which I am happy to undertake when the time comes.

Please let me know if there is anything in this updated documentation which requires fixing, or anything I've missed which should be added.  I would like to add more example bundle records, for example, to show the use of various of the options available for bundle registration. But I am reasonably satisfied with these changes.

At some point, attention should also be paid to this [somewhat parallel documentation](http://docs.plone.org/develop/addons/javascript/intro.html).  It would be good to reduce duplication by pulling some of the information from one into the other and then using more cross references to tie them together.  But that's a discussion for another time, perhaps.

Thanks very much for your time and assistance.

svx:
There are a couple of .rst related minor issues :)

The most important one is about links, we have that covered in out style-guide :) -> http://docs.plone.org/about/rst-styleguide.html#links

Currently a mix of that and the the `.._` reference is used for links in the document, from the point of rst syntax there is nothing wrong with `.._` but we do not use that, because it is bad for accessibility [screen-reader for example] and also for translations :).

Line 690: `described here` is one other issue which we should improve, that is hard for accessibility  and SEO, too

Further more I spotted one or two remaining typos, like in line 9 

As far is goes for syntax and so on, besides that it looks awesome to me :) 

Thanks a lot !

adapt-and-extend/theming/resourceregistry.rst
```diff
@@ -18,47 +18,75 @@ Introduction to Plone 5 resources
 
 Plone 5 introduces new concepts for working with JavaScript and CSS in Plone.
 
-We finally have a solution to define JavaScript modules and their dependencies - in contrast to the previous approach, where all dependencies had to be loaded in order.
-We achieve that with the tool RequireJS, which follows the Asynchronous Module Definition (AMD) approach.
-We chose AMD over other module loading implementations (like CommonJS) because AMD can be used in non-compiled form in the browser and also compiled Through-The-Web.
-This way we can also support a "development mode" like in Plone versions prior to 5, where changes in the JavaScript sources are instantly reflected (actually they are made available a few seconds after initial browser load).
+JavaScript and AMD
+------------------
+
```
ebrehault, line 23: @cewing We should state somewhere that we can write non-AMD JS in an add-on and registrering it in a bundle and that's 100% valid.
AMD is a good practice, and it is provided by default in Plone 5, so that's really easy to use in our context, but it is not mandatory.
A bundle containing something like:

```
$(document).ready(function() {console.log("Hey, I am not AMD!!!");})
```

will work fine.
I think it is important to tell it explicitly so people who are afraid of frontend feel more confortable.

cewing, line 23: @ebrehault I'd be happy to do so.  Can you find an example that does so so that I can point to a clean usage pattern?  Right now all the examples I've seen that do not use the AMD pattern are simply defaulting to using `jsregistry.xml` to get into the `plone-legacy` bundle

cewing:
@svx I think I've now addressed your comments.  I've unified all external links using the recommended pattern.  I've fixed the typo on line 9 and I've changed up the link text on what is now line 722 to make it more clear where it will take you.  

I do want to include something about the newly-added option for bundles (`stub_js_modules`), but it's not yet released, right?  Do we have a way of adding text to documentation that is hidden so that we can un-hide it when that feature is released publicly?  Alternatively, could I mark a give feature as "added in Plone 5.x.y" so that it is clear that it isn't available before that time?

svx:
@cewing no currently there is no way to hide stuff as in text, you can 'hide'stuff from/in the toctree but that is currently all.

As for versions there is http://www.sphinx-doc.org/en/stable/markup/para.html?highlight=version#directive-versionadded quite handy, and I would like to see it more often used, if we change/add things in minor versions, I guess that is what you asking for ?

Thanks btw, for all your work on it ! :)

adapt-and-extend/theming/resourceregistry.rst
ebrehault, line 23: Here:
https://github.com/collective/example.p4p5/blob/master/src/example/p4p5/profiles/plone5/registry.xml
https://github.com/collective/example.p4p5/blob/master/src/example/p4p5/browser/static/chart.js

cewing, line 23: @ebrehault Perfect!  I'll add a mention of this up here at the top, then an example in the bundle registry.  

As a question, the JS in that example is dependent on `$` being in the window namespace, right?  Is that something that is always true?  Even under the new resource registry approach?  We reliably have a reference to jQuery at `$` in the javascript global namespace?

cewing:
@svx, that directive is _exactly_ what I'm looking for.  All I need to know now is which version of `Products.CMFPlone` will contain the new `stub_js_modules`.  Then I can add that documentation in anticipation of it coming out.  @esteele can you comment on that?  The feature in question is on trunk now in `Products/CMFPlone/interfaces/resources`.  Do you have an idea which future Plone release will contain it?

adapt-and-extend/theming/resourceregistry.rst
ebrehault, line 23: Yes, $ and require are always in the global namespace.

adapt-and-extend/theming/resourceregistry.rst
```diff
@@ -18,47 +18,72 @@ Introduction to Plone 5 resources
...
+As the use of JavaScript in front-end development expanded, this model proved insufficient to solve complex dependency management issues.
+At the same time, developments in the JavaScript community led to a new approach to handling complex dependencies.
+
+Starting in Plone 5, we use a new `module <http://requirejs.org/docs/why.html>`_-based solution on the `Asynchronous Module Definition <http://requirejs.org/docs/whyamd.html>`_ (AMD) approach.
+The new Plone 5 Resource Registry uses `RequireJS <http://requirejs.org/>`_ to allow developers to define individual JavaScript modules and the dependencies they will require.
+Because RequireJS uses the AMD approach, these modules and their dependencies can be served in an uncompiled form during development and then compiled and minified Through-The-Web for use in production.
+This allows a "development mode" where changes in JavaScript source files are reflected in the browser in near real-time.
+
+It is also possible in the new Resource Registry to provide bundles with simple Javascript that does not make use of AMD.
+See :ref:`bundle_example_non_amd` below for an example of such a bundle.
```
cewing, line 35: @ebrehault this paragraph is added to address [your comment above](https://github.com/plone/documentation/pull/518/files#r52873699).

adapt-and-extend/theming/resourceregistry.rst
```diff
@@ -291,50 +495,175 @@ Resources which are registered into ``portal_javascripts`` or ``portal_css`` reg
...
+      <value key="compile">False</value>
+      <value key="depends">plone</value>
+    </records>
 
-- ``remove_bundle_on_request(request, bundle)``: Removes a bundle to the current request by specifying its name.
+.. _bundle_example_non_amd:
 
-- ``add_resource_on_request(request, bundle)``: Adds an individual resource to the current request by specifying its name.
+Non-AMD Bundles
+***************
```
cewing, line 594: @ebrehault and this code example is derived from [your pointers above](https://github.com/plone/documentation/pull/518/files#r52873699).  Could you read this and verify that it is both true and accurate?

ebrehault, line 594: It is both true and accurate :)

seanupton:
@cewing @svx reading changelogs, seems like `stub_js_modules` support was released in Products.CMFPlone 5.0.1, so should be in current (but not yet advertised as current on Plone.org downloads) Plone 5.0.2.  If so, I would hope it safe to document now without disclaimer?

adapt-and-extend/theming/resourceregistry.rst
```diff
@@ -124,166 +236,258 @@ In ``mockup/patterns/structure/js/views/actionmenu.js``:
...
+    Even if you do this, however, you should still provide a custom bundle containing only your code.
+    Then users can see the JavaScript and CSS for your package in action without having to revisit the resource registry and press the "Build" button.
 
-Bundle Definition
------------------
+Development vs. Production Mode
+-------------------------------
 
-Example based on Plone's standard bundles defined in ``Products/CMFPlone/profiles/dependencies/registry.xml``
+In development mode, each bundle loads all of its resources in the rendered site as individually, with individual requests.
```
jensens: `... as individually, with individual requests` maybe my english is not sufficient, but i dont get what that means exactly.

adapt-and-extend/theming/resourceregistry.rst
```diff
@@ -124,166 +236,258 @@ In ``mockup/patterns/structure/js/views/actionmenu.js``:
...
+    eggs = Products.CMFPlone
+    scripts = plone-compile-resources
 
+Once the script has been created you may invoke it.
+You will need to provide options indicating the ID of the Plone site in which your package is installed, and the name of the bundle you wish to compile:
+
+    ./bin/plone-compile-resources --site-id=myplonesite --bundle=mybundle
+
+This script will start up your Plone site, extract the required information and compile the bundle.
+Because of this, if you are not using ZEO, you will need to stop your instance before running this script.
```
jensens: `if you are not using ZEO` is confusing, you need to stop one instance anyway.

jensens:
Great work and a major improvement!

adapt-and-extend/theming/resourceregistry.rst
cewing: It's a typo, @jensens.  The sentence should be this:

"In development mode each bundle loads all its resources individually."

Thanks for catching it.  I'll fix it asap.

adapt-and-extend/theming/resourceregistry.rst
cewing: @jensens Is this true?  It isn't like `bin/instance adduser` where you can run it against a client that is running?  

If that's the case, then yes.  I'll re-word this.  I mis-understood.  Can you clarify for certain that you have to have a stopped plone client to run this script?  And if you are running zeo, how does it figure out which client to use?

svx:
@cewing @jensens do guys had time to have a look at https://github.com/plone/documentation/pull/518#discussion_r53104424 yet ?

I really would like to see that merged soon, since it is **huge** improvement to the docs !

Thanks again, to all ppl who took some time to work/review/help to improve it !

cewing:
@svx, I've fixed the problems @jensens noted, and adjusted the wording for the `plone-compile-resources` script to clarify that you need a stopped instance.  I think that's non-controversial and should not hold up merging.

I've also added documentation for the `stub_js_modules` option for a bundle record and noted that it was added in 5.0.1, so we should be good to go on merging.  

thanks, all

svx:
Just to make sure one last time, everyone is fine with merging ?

I ask because this part of Plone is really not my part of wisdom, so to say

cewing:
I believe we are ready.

jensens:
i reviewed it and its so much much better now, please merge and release soon :)

gforcada:
Can we get some commits squashed/fixup'ed before? :-) https://docs.plone.org/develop/coredev/docs/git.html

svx:
@cewing do you want to squash ? :)

cewing:
@svx, will do.  Gimme a few minutes here :)

cewing:
@svx, this is squashed and ready to merge.  Fire away!

svx:
Yes ! Thanks ! Beer in Boston !

PR #528, opened 2016-02-21
Fixed wrong imports and old references to plone.multilingual

dmunicio:


svx:
Can someone with deeper knowledge review this ?, I do not use multilingual :)

@dmunicio thanks !

jensens:
@svx everythings fine. Does @dmunicio need a CLA for this change? I cant find the name in the list of contributors.

svx:
@dmunicio @jensens 

I also can't find @dmunicio in the list, @dmunicio did you signed the CLA already?

@jensens good question, afaik it depends on the contribution, for minor ones you do not need to sign, at least that is what I learned on the last writethedocs meetup and that is how kde and others are doing it, but just to be sure, we should ask @polyester.

@polyester could you check if @dmunicio already signed the CLA and if we need that for 'minor' changes ?

dmunicio:
I have not signed anything. I have no problem to sign whatever is necessary
but I think it is overkill for some doc fixes...

polyester:
We don't need it for minor fixes, just as long as @dmunicio is fine with submitting them under Creative Commons  (as the docs are,  Creative Commons Attribution 4.0 International license to be exact). 

It is, of course, always nice to welcome new developers, and signing a CLA would be a nice thing, but it is not technically needed.

svx:
OK will merge then

PR #531, opened 2016-02-22
the beginning of a deprecation guide

gforcada:
@jensens any reason not to pull request this? I spent quite a lot of time to find it! :-/


jensens:
hmm, sorry, forgot about this one. its not very complete now, but better than nothing and could be a good seed for further improvements.

thet:
Looks good! We should merge. @gforcada ?

thet:
Btw. I added some info about deprecating methods and properties and added this document to the index

PR #533, opened 2016-02-26
more information and backgrounds about deprecations in Zope, Plone and Python

jensens:
I just found my old writings about deprecation and merged parts of it into here. I also added some new information about how to deprecate GenericSetup profiles.

This part is now more complete.

/cc @gforcada


gforcada:
Awesome!

PR #536, opened 2016-03-03
typo in import.

thet:
@svx i was quite tempted to commit this directly :)


svx:
@thet I know that feeling :)

Thanks for not doing it !!!!

PR #540, opened 2016-03-07
Unify style for console usage

svx:
- unify 'look and feel' for console usage

### DIFF
--- a/manage/installing/installation.rst
+++ b/manage/installing/installation.rst
@@ -93,7 +93,7 @@ Install the operating system software and libraries needed to run Plone
 
 .. code-block:: console
 
-    sudo apt-get install python-setuptools python-dev build-essential libssl-dev libxml2-dev libxslt1-dev libbz2-dev libjpeg62-dev
+    $ sudo apt-get install python-setuptools python-dev build-essential libssl-dev libxml2-dev libxslt1-dev libbz2-dev libjpeg62-dev
 
 .. note::
 
@@ -105,7 +105,7 @@ You will probably also want these optional system packages for handling of PDF a
 
 .. code-block:: console
 
-    sudo apt-get install libreadline-dev wv poppler-utils
+    $ sudo apt-get install libreadline-dev wv poppler-utils
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


jensens:
In my opinion the leading `$` is not very copy-paste friendly and I always hated it when I saw that used in documentation.

svx:
@jensens I see you point, even if I think different, we have currently a mix in our docs.
I want to get rid of this confusing mix.

Even if it is not perfect, this is still one common way in documentation to make clear that this part of docs is a command.

This is not something I came up with, I even asked lots of friends from wtd, if its sane to use. 

Even if it may less user-friendly for copy and paste, it still has IMHO a value for the visual factor of docs, which is as important, because it helps people to understand.

jensens:
Maybe I'am overall wrong and readability is more important.
But at least it makes copy-paste an annoying experience. 
So maybe we have here a +1 in readability and a -1 in usability.
Just my opinion as a user, I'am not a good doc-writer.

polyester:
I have a question on this as well. It feels wrong to manually add a $ to a code-block; the whole point of unifying on a single recommendation for   `.. code-block:: console` is that Sphinx and CSS then take care of all visual clues that this is a command. Even if that means putting a "$" as background sprite before the command.

If that is not clear enough, then our CSS needs updating, but manually adding characters in front of a command is probably not the way to go, it mixes content with layout.

svx:
https://pypi.python.org/pypi/sphinx-prompt/1.0.0 and other add-ons changing the background and adding a $ out of the box.

As I mentioned before I 'just' tried to unify the docs, because we use with $ and without $ through all the docs and that is even worse.

Sometimes it is really hard to stay motivated ... just saying .....

jensens:
yeah, sorry if I brought in stop energy here :/

polyester:
@svx, the effort is _hugely_ appreciated. And, as said, I'm all in favour of big visual clues to readers that something is a shell command. So, changing the background, and adding a (non-selectable) prompt is **good**. 

Just in this PR you are adding it in the _source_ RST files 'hard', where I think it should be the job of addons like sphinx-prompt or CSS to add those visual clues once it finds the `.. code-block:: console` directive.

So, no stop energy intended, but just trying to see if we do it the best way.

PR #551, opened 2016-03-10
How to create a new theme based on Barceloneta

ebrehault:
based on Asko's article http://datakurre.pandala.org/2015/05/customize-plone-5-default-theme-on-fly.html


svx:
Thanks !

PR #553, opened 2016-03-11
Update to Plone 4.3.8

svx:
- Update to latest Plone version

### DIFF
--- a/manage/installing/installation.rst
+++ b/manage/installing/installation.rst
@@ -28,9 +28,8 @@ Download Plone
 
 Plone is available for Microsoft Windows, Mac OSX X, Linux and BSD operating systems.
 
-`Download the latest Plone release <https://plone.org/products/plone/latest_release>`_.
+`Download the latest Plone 4 release <https://launchpad.net/plone/4.3>`_.
 
-Binary installers are available for Windows and OS X.
 Installation on Linux, BSD and other Unix workalikes requires a source code installation, made easy by our Unified Installer. "Unified" refers to its ability to install on most Unix workalikes.
 
 Plone installation requirements
--- a/manage/installing/installation.rst
+++ b/manage/installing/installation.rst
@@ -125,11 +125,11 @@ Download the latest Plone unified installer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Download from `the plone.org download page <https://plone.org/download>`_ to your server using wget command. Curl also works.
-Substitute the latest version number for 4.3.6 in the instructions below.
+Substitute the latest version number for 4.3.8 in the instructions below.
 
 .. code-block:: console
 
-    wget --no-check-certificate https://launchpad.net/plone/4.3/4.3.6/+download/Plone-4.3.6-UnifiedInstaller.tgz
+    wget --no-check-certificate https://launchpad.net/plone/4.3/4.3.8/+download/Plone-4.3.8-UnifiedInstaller.tgz
 
 Run the Plone installer in standalone mode
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


manage/installing/installation.rst
```diff
@@ -125,11 +125,11 @@ Download the latest Plone unified installer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Download from `the plone.org download page <https://plone.org/download>`_ to your server using wget command. Curl also works.
```
jensens, line 126: FYI: https://plone.org/download redirects to 5.0.2 and its not easy to get to the 4.3.8 release from there. So we may want to point directly to https://plone.org/products/plone/releases/4.3.8

svx, line 126: @jensens Got catch ! May I offer a other solution ? 

The reason is I want to avoid that we have to fix all this tiny links all the time, yes there is also a plugin to do so, but still not perfect :)
We will face this more often and also with other releases like Plone 5, for example.

Since some people are not always able to use the latest version, maybe we could add a link with a big disclaimer to https://plone.org/download ?
Something like "you always should use the latest release, but if for some reason you need a older version, please go to https://plone.org/download/old and choose the version you need" or so which is a link to https://launchpad.net/plone/+download ?

@esteele @smcmahon @tkimnguyen @3dogMcNeill @polyester what do you think ? I included all of you, because if we do so that also should go on to new and old plone.org

jensens, line 126: I like the overall direction of your idea, but I would then make it more generic, like so:
- https://plone.org/download -> overview (folder_summary_view or something like that) OR redirect to latest (both ways each has it pros and cons)
- https://plone.org/download/latest -> always redirects to most current release
- https://plone.org/download/5-latest -> the link to be used in the 5.x series documentation
- https://plone.org/download/4-latest -> the link to be used in the 4.x series documentation
- https://plone.org/download/3-latest -> the link to be used in the 3.x series documentation (only if really still needed)
- https://plone.org/download/old -> https://launchpad.net/plone/+download

sneridagh, line 126: We still have to decide how to deal with the "downloads" page on the new plone.org. Since we deprecated PSC we need to design and implement a brand new one. I agree with @jensens however, it would be great if it's the least human dependant possible. Let's say that we have some way to detect which are the last versions and update links and content accordingly.

How we could achieve that? Any ideas?

jensens, line 126: It should be easy to extract from the PyPI information we can get as json from https://pypi.python.org/pypi/Products.CMFPlone/json
So we just need a view using `requests`, some caching + some logic (we have to define first) and some html. This is imo beginner level Plone coding.

tkimnguyen, line 126: How about something simpler like: a portlet that includes links to a list of all releases, in reverse chronological order? The main /download page would still point directly at the latest release. That way, no extra code, and no human maintenance required.

svx:
I changed the link

```
 Download from `the plone.org download page <.....>`_
```

to point to the overview of Plone 4 releases on Launchpad, still not perfect, but IMHO a bit less confusing than plone.org/downloads where you can't find Plone 4 easily.

Thus still not perfect but slightly better ?!

manage/installing/installation.rst
svx, line 126: Should we open a own ticket for this issue ?

@tkimnguyen it would make sense to also add links to the versions, like 4.3.8 -> https://launchpad.net/plone/...

otherwise ppl would not be able to find the download of the version they want to use, right ?

tkimnguyen, line 126: Yes, let's open a new ticket for this. 

@svx whatever is easiest (no coding) and requires zero maintenance.

jensens, line 126: +1 -> new ticket , merge this one

PR #556, opened 2016-03-17
typo in the installation manual 

reyha:
removes a typo in the installation manual


gforcada:
@reyha thanks for the fix!

@plone/documentation-team a fix like that does not need CLA right? Revert if needed.

PR #576, opened 2016-04-02
wait for FF on slower machines

polyester:
Fixes: #

Improves: makes robotscreenshot work on slower machines, where some forms take a long time to render. Solution: check if the button you want to click, is visible.


svx:
thanks !

PR #600, opened 2016-04-21
Fix a typo and give better explanation

staeff:
Fixes: A typo

Improves: Explanation on invoking the bin/debug command


svx:
Thanks !

PR #611, opened 2016-05-17
Replace grok/plone.directives and small formal fixes

staeff:
Improves: 
- replace mentions of plone.directives from fieldsets.rst
- several small formal fixes, like missing spaces and wording

### DIFF
--- a/develop/addons/schema-driven-forms/customising-form-behaviour/actions.rst
+++ b/develop/addons/schema-driven-forms/customising-form-behaviour/actions.rst
@@ -27,7 +27,7 @@ in our pizza order form:
         def handleCancel(self, action):
             ...
 
-The name of the method is not particularly important, so long as it is
+The name of the method is not particularly important, but it needs to be
 unique. The body of the handler function may react to the button however
 is appropriate for the form’s use case. It may perform a redirect or
 update form properties prior to re-rendering of the form. It should not
@@ -87,7 +87,7 @@ Updating button properties
 As with regular widgets, it is sometimes useful to set properties on
 buttons after they have been instantiated by *z3c.form*. One common
 requirement is to add a CSS class to the button. The standard edit form
-in*plone.directives.form* does this, for example, to add Plone’s
+in *plone.directives.form* does this, for example, to add Plone’s
 standard CSS classes. The usual approach is to override
 *updateActions()*, which is called during the form update cycle:
 
--- a/develop/addons/schema-driven-forms/customising-form-behaviour/fieldsets.rst
+++ b/develop/addons/schema-driven-forms/customising-form-behaviour/fieldsets.rst
@@ -30,13 +30,10 @@ purely for aesthetic effect.
 
 ::
 
-    from five import grok
-    from plone.supermodel import model
-    from plone.directives import form
-
-    from zope import schema
-
     from example.dexterityforms.interfaces import MessageFactory as _
+    from plone.autoform import directives as form
+    from plone.supermodel import model
+    from zope import schema
 
     ...
 
--- a/develop/addons/schema-driven-forms/customising-form-behaviour/vocabularies.rst
+++ b/develop/addons/schema-driven-forms/customising-form-behaviour/vocabularies.rst
@@ -183,8 +183,8 @@ Here is an example that returns our pizza types:
 
         return SimpleVocabulary(terms)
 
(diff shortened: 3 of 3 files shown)


develop/addons/schema-driven-forms/customising-form-behaviour/actions.rst
```diff
@@ -27,7 +27,7 @@ in our pizza order form:
         def handleCancel(self, action):
             ...
 
-The name of the method is not particularly important, so long as it is
+The name of the method is not particularly important, it only needs to be
```
svx: can we remove 'only' that could give user a wrong impression

staeff: What about 'but it needs to be unique.'?

plone-foundation: On 05/17/2016 04:05 PM, Stephan Klinger wrote:

> In
> develop/addons/schema-driven-forms/customising-form-behaviour/actions.rst
> https://github.com/plone/documentation/pull/611#discussion_r63527537:
> 
> > @@ -27,7 +27,7 @@ in our pizza order form:
> >          def handleCancel(self, action):
> >              ...
> > 
> > -The name of the method is not particularly important, so long as it is
> > +The name of the method is not particularly important, it only needs to be
> 
> What about 'but it needs to be unique.'?
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> https://github.com/plone/documentation/pull/611/files/8268ac35ed2a073e7d4e11b152e4ea2f8952691d#r63527537
> 
> This is fine ! :)

We are want to avoid words like: just, only, easy, ... because the could
give user a wrong impression, what may easy for you is maybe not that
easy for me :)

If you are person ho tries to get all this stuff, such words can be a
huge burden, that is why we want to avoid them.

svx:
Thanks a lot !

develop/addons/schema-driven-forms/customising-form-behaviour/actions.rst
```diff
@@ -87,7 +87,7 @@ Updating button properties
 As with regular widgets, it is sometimes useful to set properties on
 buttons after they have been instantiated by *z3c.form*. One common
 requirement is to add a CSS class to the button. The standard edit form
-in*plone.directives.form* does this, for example, to add Plone’s
+in *plone.directives.form* does this, for example, to add Plone’s
```
gforcada, line 90: @staeff plone.directives.form needs to be replaced by something else here

develop/addons/schema-driven-forms/customising-form-behaviour/actions.rst
staeff: @svx thanks for the heads up. I agree with the goal to avoid these words in context where they could make something sound easy which not necessarily is easy for everybody. 
I will keep this in mind while working with the docs at the berlin-sprint. 
Even though I want to add, that the use in the sentence above is in my opinion a false positive (and I feel there are correct uses of the above words).

It was saying: _The name of the method is not particularly important, it only needs to be uniqe._ - I think, thats a different use than saying _You only have to import this 25 modules._

develop/addons/schema-driven-forms/customising-form-behaviour/actions.rst
staeff, line 90: @gforcada thanks for the heads up. I missed that while rather looking for rst-problems, I will look at the content side today.

davisagli, line 90: should be `plone.dexterity` here

svx, line 90: @staeff would you mind to check ? If we can update this with plone.dexterity we can include that into our next release, which would be super awesome ! :)

Thanks  a lot !

staeff, line 90: @svx I made the replacement here https://github.com/plone/documentation/pull/633. What will be the  date for the next release? I am still going through the whole section as stated here https://community.plone.org/t/clean-up-documentation-section-schema-driven-forms/2163 (there is much more stuff, that needs to be replaced).

svx:
@staeff I a targeting upcoming Monday 6 of June for the next release :)

PR #617, opened 2016-05-24
Update GenericSetup documentation

mauritsvanrees:
Improves: GenericSetup documentation.

- Added lots of examples of GenericSetup xml files, especially with uninstall examples. Includes notes on whether (and how) the `remove` and keyword is supported in each file. Not all possible xml files are handled, but I have been adding uninstall profiles for core Plone packages the past few months and have updated this documentation based on the files I encountered.
- Updated GenericSetup best practices and added examples for the `upgradeDepends` directive.
- General cleanup on this file, which should bring it more in line with the documentation best practices.

I am confident about the content. But please have a look at styling, documentation best practices, wrong references to other parts of the documentation.

I am happy to squash these commits into one if it helps.

### DIFF
--- a/develop/addons/components/genericsetup.rst
+++ b/develop/addons/components/genericsetup.rst
@@ -1126,6 +1126,51 @@ When you search for ``metadata.xml`` in the documentation, you will find more in
     But for dependencies no ``purge`` or ``remove`` keyword is supported.
 
 
+portal_atct.xml
+---------------
+
+This is deprecated.
+It was used mostly for Archetypes Topics, the old Collections.
+Since Plone 4.2 you can use new style Collections.
+Please use those.
+
+.. code-block:: xml
+
+    <?xml version="1.0"?>
+    <atcttool
+        xmlns:i18n="http://xml.zope.org/namespaces/i18n">
+     <topic_indexes i18n:domain="plone">
+      <index name="allowedRolesAndUsers"
+         description="The roles and users with View permission on an item"
+         enabled="False" friendlyName="Internal Security"
+         i18n:attributes="description; friendlyName" />
+      <index name="created" description="The time and date an item was created"
+         enabled="True" friendlyName="Creation Date"
+         i18n:attributes="description; friendlyName">
+       <criteria>ATFriendlyDateCriteria</criteria>
+       <criteria>ATDateRangeCriterion</criteria>
+      </index>
+     </topic_indexes>
+     <topic_metadata i18n:domain="plone">
+      <metadata name="created"
+         description="The time and date an item was created"
+         enabled="False" friendlyName="Creation Date"
+         i18n:attributes="description; friendlyName" />
+     </topic_metadata>
+     <property name="title">ATContentTypes Tool</property>
+     <property name="image_types">
+      <element value="Image"/>
+      <element value="News Item"/>
+     </property>
+     <property name="folder_types">
+      <element value="Image"/>
+     </property>
+     <property name="album_batch_size">30</property>
+     <property name="album_image_scale">thumb</property>
+     <property name="single_image_scale">preview</property>
+    </atcttool>
+
+
 portlets.xml
 ------------
 
--- a/develop/addons/components/genericsetup.rst
+++ b/develop/addons/components/genericsetup.rst
@@ -1156,7 +1156,7 @@ Example:
   It can also contain a specific skin, for example ``Plone Default``, ``Sunburst Theme``, ``Plone Classic Theme``.
 
 You can set a few properties on the ``portal_skins`` object.
-``Products.CMFPlone`` set good defaults:
+``Products.CMFPlone`` sets good defaults which you should keep:
 
 .. code-block:: xml
 
--- a/develop/addons/components/genericsetup.rst
+++ b/develop/addons/components/genericsetup.rst
@@ -498,7 +498,7 @@ Upgrade steps
 
 You can define upgrade steps to run code when someone upgrades your product from version *x* to *y*.
 
-As an example, let's say that the new version of YOUR.PRODUCT defines a *price* field on a content type *MyType* to be a string, but previously (version 1.1 and earlier) it was a float.
+As an example, let's say that the new version of your.package defines a *price* field on a content type *MyType* to be a string, but previously (version 1.1 and earlier) it was a float.
 Code that uses this field and assumes it to be a float will break after the upgrade, so you'd like to automatically convert existing values for the field to string.
 
 Obviously, you could do this very quickly in a simple script, but having a GenericSetup upgrade step means non-technical people can do it as well.
@@ -525,16 +525,16 @@ Next we add an upgrade step:
     <configure
         xmlns="http://namespaces.zope.org/zope"
         xmlns:genericsetup="http://namespaces.zope.org/genericsetup"
-        i18n_domain="YOUR.PRODUCT">
+        i18n_domain="your.package">
 
       <genericsetup:upgradeStep
           title="Convert Price to strings"
           description="Price was previously a float field, it should be converted to string"
           source="1000"
           destination="1100"
-          handler="YOUR.PRODUCT.upgrades.convert_price_to_string"
+          handler="your.package.upgrades.convert_price_to_string"
           sortkey="1"
-          profile="YOUR.PRODUCT:default"
+          profile="your.package:default"
           />
 
     </configure>
@@ -555,7 +555,7 @@ The code for the upgrade method itself is best placed in a *upgrades.py* module:
     from plone import api
     import logging
 
-    PROFILE_ID = 'profile-YOUR.PRODUCT:default'
+    PROFILE_ID = 'profile-your.package:default'
 
 
     def convert_price_to_string(context, logger=None):
@@ -569,7 +569,7 @@ The code for the upgrade method itself is best placed in a *upgrades.py* module:
 
         if logger is None:
             # Called as upgrade step: define our own logger.
-            logger = logging.getLogger('YOUR.PRODUCT')
+            logger = logging.getLogger('your.package')
 
         # Run the catalog.xml step as that may have defined new metadata
         # columns.  We could instead add <depends name="catalog"/> to
--- a/develop/addons/components/genericsetup.rst
+++ b/develop/addons/components/genericsetup.rst
@@ -1244,6 +1244,63 @@ Please use those.
     </atcttool>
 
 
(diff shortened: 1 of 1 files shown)


develop/addons/components/genericsetup.rst
```diff
@@ -184,33 +167,51 @@ Unit testing example::
...
+3. CMFQuickInstaller 3.0.9 (Plone 4.3.7 and higher) makes the second part a bit easier.
+   If the uninstall method is not found, it looks for a profile with id ``uninstall`` and applies it.
+
+.. note::
 
-For an example, see the `collective.pdfpeek source code
-<https://github.com/collective/collective.pdfpeek/tree/master/collective/pdfpeek/profiles>`_.
+    Plone 5.1 is scheduled to not use the CMFQuickInstaller, but always apply an ``uninstall`` profile.
+    For more information on this plan,
+    see `PLIP 1340 <https://github.com/plone/Products.CMFPlone/issues/1340>`_.
```
svx, line 230: We will have to update that, when we will release 5.1, seeing that is it handy to write it like that ? Maybe we should already prepare a branch where we have done this for 5.1 ? 
I fear if we do not prepare that now, we will forget it.

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
 .. note ::
 
     This example is for Extensions/install.py, an old Plone 2 way of writing installers and uninstallers.
-    It is still working in Plone 5.0,
-    but will likely go away in Plone 5.1.
+    It is still working in Plone 5.0, but will likely go away in Plone 5.1.
+    In Plone 5.1 you can decide not to create an uninstall profile.
```
svx, line 768: Same here, we are mixing 5.0 and a not yet released 5.1

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
...
+
+Any other categories are not displayed in the overview control panel.
+Note that in Plone 4, only ``Member``, ``Plone`` and ``Products`` were supported.
+For add-ons, the category ``Products`` is recommended.
+
+The ``action_id`` must be unique over all categories.
+
+Only one permission is allowed.
+
+Uninstall example:
```
svx, line 1156: Does the info for Plone 4 has to be here ? We have own docs for Plone 4, so if it has to stay here, we should use .. versionadded:: version

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
...
+jsregistry.xml
+--------------
+
+.. note::
+
+    This is deprecated in Plone 5.
+
+See :ref:`resourceregistries`.
+
+
```
svx, line 1298: If it is deprecated, does it have to stay, if it a Plone 4 thing, move it there and maybe add a link, if we need to do that?

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
...
+kssregistry.xml
+---------------
+
+.. note::
+
+    This is deprecated in Plone 4.3.
+
+See :ref:`resourceregistries`.
+
+
```
svx, line 1306: Same as above

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
...
+portal_atct.xml
+---------------
+
+.. note::
+
+    This is deprecated.
+    It was used mostly for Archetypes Topics, the old Collections.
+    Since Plone 4.2 you can use new style Collections.
+    Please use those.
+
```
svx, line 1353: Same here

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
...
+
+propertiestool.xml
+------------------
+
+.. note::
+
+    This is deprecated in Plone 5.
+    Most properties are now handled in the configuration registry and can be configured in ``registry.xml``.
+
```
svx, line 1464: Same here, since we have this already on a couple of places, maybe we should rewrite the doc in a way, which is mentioning all the depredations once like 'changes in versions' or something, otherwise we should use .. deprecated:: version

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
...
+-------------------
+
+.. note::
+
+    This is deprecated in Plone 5.
+
+Code is in ``Products.ResourceRegistries.exportimport.resourceregistry``.
+
+See: :doc:`Front-end: templates, CSS and Javascript </adapt-and-extend/theming/templates_css/index>`.
+
```
svx: Same here

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
...
+
+tinymce.xml
+-----------
+
+.. note::
+
+    This is deprecated in Plone 5: it is done in ``registry.xml`` instead.
+    The fields are defined in ``Products.CMFPlone.interfaces.controlpanel``.
+
```
svx, line 1831: Same here

develop/addons/components/genericsetup.rst
```diff
@@ -717,203 +742,1393 @@ Example:
...
-In the propertiestool.xml you can change all values of the portal_properties.
+actionicons.xml
+---------------
 
-take a look at: https://plone.org/documentation/manual/developer-manual/generic-setup/reference/properties-ref
+.. note::
+
+    This is deprecated.
+    Plone 5 no longer reads this file.
+    The icons should go in ``actions.xml`` directly.
```
svx, line 928: So if its is Plone 4, we can remove it ?

develop/addons/components/genericsetup.rst
```diff
@@ -603,15 +591,15 @@ The code for the upgrade method itself is best placed in a *upgrades.py* module:
                 count += 1
```
svx, line 612: Can we change Dutch to English, otherwise we mix languages in the document :)

svx:
Awesome, thanks a lot !! This is great !!! Thank you so much !!!

I added some comments, over some parts which we may want to change :)

For Me as a reader, some parts are confusing, because currently we mix different Plone versions here, if it would be possible I would like to move stuff which is only Plone 4 to the docs of Plone 4 for example.
Also we are talking here already about 5.1, which is not released yet, this is tricky, because we possible confuse reader and this has to be updated/re-written after the release :)

Further I added a note, about languages, currently we mix here Dutch and English :)

mauritsvanrees:
Thanks for the review. I made minor changes, at least removing the Dutch variable names.

My colleague Fred has a good remark about the info on different versions. You are right that giving only information on Plone 5.0 is clearer for someone who wants to make an add-on or customer code that only has to run on Plone 5.0.
But what if you want to write an add-on that works on 4.3 and 5.0 and is ready for 5.1? If we split all the info in different branches of the documentation, the add-on creator would need to check three similar pages.

So I would be inclined to keep all info about these versions in one file.
Does this make sense?

More use of `.. versionadded:` and `.. deprecated:` could help. I didn't know those.

svx:
Yeah, that is a valid point ! We never had to deal with that before :)

We have to find a way to deal with that, in a way that we are able to use through the whole docs, I guess we will face that on different locations and it will get more in the future, since we'll get more and faster releases.

I am in favor of using .. versionadded: and .. deprecated: ,but we should make sure that we write the docs and use these tags in a sane way, so that the docs are still understandable.

IMHO the risk we have to take care of, is that the docs or certain parts could get really hard to understand and to follow if you have for example lots of these tags next to each other in one document :)

@polyester what do you thing ?

polyester:
more semantic tagging is always good! 
So yes, use `.. versionadded::` and `.. deprecated::`. We should have them marked up in CSS to be _subtle_ in how they look, and not big jarry blocks like `.. note::`, so that the flow of reading is not interrupted.

svx:
@polyester so do you volunteer to check them, how they look like in css and if necessary fix them too ?

mauritsvanrees:
I have added `deprecated` and `versionadded`.

Note that 'deprecated since version X' sometimes means 'it still works in X, but will be gone in X+1, so you should not use it', and sometimes 'it does not work at all in X'. Hopefully the comments make this difference clear.

svx:
@mauritsvanrees awesome ! Thanks I will check later tonight !

svx:
The only point which we still may have is, that we do not have Plone 5.1 released yet, but we mention it already :)

So strictly spoken I would wait with the merge till 5.1 is released, just to avoid possible confusion, on the other hand this is for developer not for _normal_ end user, meaning that it actually could be useful to merge and publish it already, because it could help.   

@polyester do you have a opinion on that ?

Proselint is giving some suggestions for replacing/changing words like _but_ and _very_ , for now I consider that less important and will add it to the to do list.

Thanks a lot, again !

mauritsvanrees:
In my opinion the info on Plone 5.1 helps developers to prepare for the new version now.
I would not want to wait with merging. But I could remove those bits and put them in a different pull request that we can keep in our pocket, if you want.
Technically this might be better: the relevant plip might not be finished or might be rejected or postponed.

polyester:
The audience for this is definitely developers, and I think it is important they have all the info for preparing for an imminent release, so they can start preparing their add-ons already. That's in the interest of end-users also. They are also informed enough to know how to work with upcoming versions.
So, I'm +1 for merging.

sverbois:
+1 to publish 5.1 features now

svx:
ok lets try it

mauritsvanrees:
Cool! Thanks for review and opinions!

svx:
Thanks for all your work !

svx:
I will push a new build upcoming weekend to docs.plone.org

mauritsvanrees:
It was a pleasure.
I am looking forward to using this information a lot. :-)

polyester:
I'll try to get some nice css in for the semantic classifiers. Once I fix some local build issues, Ubuntu 16.04 and its python are a bit temperamental...

mauritsvanrees:
The GenericSetup page contains some documentation that is pulled in from elsewhere, like `plone.app.portlets` for the section on `portlets.xml`. I have updated the documentation in that package:
https://github.com/plone/plone.app.portlets/commit/3209b1d76f1b4e39e1c0ab2c9b1064dc49c81ce0

Does anything need to happen to get this included in the next docs update, like adding it to `auto-checkouts` or making a release?

mauritsvanrees:
I have just released plone.app.portlets 4.1.1 with the improved documentation. Note that this is only used in the coredev 5.1 branch, not coredev 5.0.

PR #621, opened 2016-05-27
Modfy a documentaion for dexterity AddCoustomView on ZCML with CMFDef…

terapyon:
Fixes: #

Improves:

## 
## 
## 

…ault

Plone 5 doesn't have `Products.CMFDefault`, I think we don't already need `Products.CMFDefault.interfaces.ICMFDefaultSkin`.
Do you think so?

### DIFF
--- a/develop/plone/content/dexterity.rst
+++ b/develop/plone/content/dexterity.rst
@@ -86,8 +86,7 @@ You can register such an adapter in ``configure.zcml``
 
         <adapter
             for="Products.CMFCore.interfaces.IFolderish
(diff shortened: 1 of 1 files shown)


svx:
@terapyon thanks !

@jensens this part of plone is not my strong suit, is that OK to merge ?

jensens:
The old registration does not make sense at all (Browserpage with 3 interfaces?) . 
The change is fine and ok for merge.

Anyway, the whole chapter here has some grok-pollution left and needs some love.

svx:
@jensens thanks ! @terapyon thanks !

PR #628, opened 2016-05-31
Made isort.cfg more readable by adding spaces.

mauritsvanrees:


mauritsvanrees:
And for clarity: yes, the `isort.cfg` works fine with these extra spaces. :-)

PR #633, opened 2016-06-01
replace plone.directives.form

staeff:
Improves: Replace one mention of _plone.directives.form_ with _plone.dexterity_ as discussed here: https://github.com/plone/documentation/pull/611


svx:
Thanks !

PR #652, opened 2016-07-03
Remove notes, they are already fulfilled

gforcada:
Fixes: #

Improves:

- remove a couple of notes mentioning that the docs should talk about bobtemplates.plone rather than zopeskel. The docs had already been updated and the notes left behind :-(

It should be ported to branch 5.0 as well.


gforcada:
@jensens thanks for merging I manually cherry-picked it into 5.0

PR #654, opened 2016-07-03
Update zopeskel references to bobtemplates.plone

gforcada:
Fixes: #

Improves:

- Update link to docs.plone.org and reference bobtemplates.plone

### DIFF
--- a/develop/plone/content/types.rst
+++ b/develop/plone/content/types.rst
@@ -80,10 +80,6 @@ Python instead, here's how::
         # Return (id, title) pairs
         return [ (id, portal_types[id].title) for id in prepared_types ]
 
-.. note:: 
-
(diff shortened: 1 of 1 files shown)


gforcada:
@jensens thanks for merging I manually cherry-picked it into 5.0

@plone/documentation-team what's the workflow for backporting to 5.0/4.3 branches? Just cherry-pick when something gets merged as I did?

polyester:
@gforcada as far as I know, yes. 
It would be good to document this... that becomes kind of meta, but it does help. Maybe you can just press the 'up' arrow on your keyboard a few times and turn that into docs ;-)

gforcada:
@polyester where do you want me to put that? :-)

polyester:
@gforcada I would create a new file under /about/cherrypicking.rst 

and then we link to it from /about/contributing.rst (or not, as it's very specific). For now, leave the link, but an example how to cherry-pick would surely help others.

PR #655, opened 2016-07-04
Add documentation about TinyMCE styles and formats

davilima6:
After https://community.plone.org/t/cant-add-original-style-to-plone5-tinymce/2412

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,13 @@
 Changelog
 =========
 
+
+Unreleased
+==========
+
+- Add documentation about TinyMCE styles and formats [davilima6]
+
+
 20160606
 ========
 
@@ -29,7 +36,6 @@ Changelog
 - Various typo fixes [jianaijun]
 
 
-
 20160409
 ========
 
@@ -37,6 +43,7 @@ Changelog
 - Improve rst style guide [svx]
 - Fix 588 [svx]
 - Fix 582 [svx]
+
 
 20160405
 ========
@@ -46,6 +53,7 @@ Changelog
 - Fix #579 [rnixx] [svx]
 - Fix #583 [polyester]
 - Fix 'Feedback widget' [svx]
+
 
 20160330
 ========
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -2,6 +2,13 @@
 Changelog
 =========
 
+
+Unreleased
+==========
+
+- Add documentation about TinyMCE styles and formats [davilima6]
+
+
```
gforcada, line 11: I think documentation repository is one exception to the rule of always adding a change log entry :-)

develop/plone/forms/wysiwyg.rst
```diff
@@ -120,11 +120,87 @@ You can use TinyMCE body selector make your CSS class have different styles in v
 
         Firefox does not actually support column breaks, so this was useful headaching experience.
 
+
 Customizing TinyMCE options
 ----------------------------
 
+Plone 4 uses TinyMCE 3. Plone 5 upgraded to TinyMCE 4, which works with a new concept called formats and therefore a new syntax for inline styles: `Your Custom Format's Title|custom_format_id|custom_icon_id`.
```
gforcada, line 127: Given that we have docs versioned for each Plone release, does it make sense to explain the differences here?

CHANGES.rst
davilima6, line 11: I agree, however I'm not sure this change is smaller than some from previous release. Would you prefer me to remove it from changelog at this time?

gforcada, line 11: Let's ask the @plone/documentation-team team

develop/plone/forms/wysiwyg.rst
davilima6, line 127: I think so, because it's not really diving into Plone 4 explanations, it's just the subject introduction with a tiny bit of context/comparison on how things evolved. Probably useful for keyword searches too (i.e. because mentioning `tinymce.xml`).

CHANGES.rst
polyester, line 11: no, updating changelog is good. I always forget... Although for the docs we try to also have 'human-readable' overview of what changed as well, but the one-line change notes are a good foundation for that.

PR #661, opened 2016-07-09
WIP: Security Guidelines

vangheem:
Improves: Add programming security guidelines

### DIFF
--- a/develop/plone/security/guidelines.rst
+++ b/develop/plone/security/guidelines.rst
@@ -43,7 +43,7 @@ If you forgot to specify permissions on a persistent traversable object, and it 
 
 Do not:
 
-code-block:: python
+.. code-block:: python
 
     class MyContent(DexterityItem):
         def foobar(self):
@@ -52,7 +52,7 @@ code-block:: python
 
 Do this:
 
-code-block:: python
+.. code-block:: python
 
     class MyContent(DexterityItem):
         def foobar(self):
@@ -61,7 +61,7 @@ code-block:: python
 
 Or this:
 
-code-block:: python
+.. code-block:: python
 
     class MyContent(DexterityItem):
         security = ClassSecurityInfo()
@@ -80,7 +80,7 @@ If you build HTML from user input, make sure to always escape the input.
 
 A common pitfall in jQuery will look like this:
 
-code-block:: javascript
+.. code-block:: javascript
 
     var value = '<script>alert("hi")</script>';
     $('body').append($(value));
@@ -88,7 +88,7 @@ code-block:: javascript
 
 By default, jQuery is not safe. To do the previous example in jQuery, you could:
 
-code-block:: javascript
+.. code-block:: javascript
 
     var $el = $('<div/>');
     var value = '<script>alert("hi")</script>';
@@ -98,13 +98,13 @@ code-block:: javascript
 
 In underscorejs templates make sure to use:
 
-code-block:: javascript
+.. code-block:: javascript
 
     <%- … %>
 
 Do not(underscorejs):
 
-code-block:: javascript
+.. code-block:: javascript
 
     <%= … %>
 
--- a/develop/plone/security/guidelines.rst
+++ b/develop/plone/security/guidelines.rst
@@ -23,7 +23,9 @@ Manual filtering
 
 If you need to use structure with unfiltered input, you can manually run Plone's output filtering engine on arbitrary html.
 
-Example::
+Example:
+
+.. code-block:: python
 
     from plone import api
 
--- a/develop/plone/security/guidelines.rst
+++ b/develop/plone/security/guidelines.rst
@@ -6,10 +6,10 @@ Programming Guidelines
 Use of structure in page templates
 ------------------------------------
 
-Do not use the `structure` page template feature with unfiltered input. This can lead to
-XSS attacks.
+Do not use the `structure` page template feature with unfiltered input.
+This can lead to XSS attacks.
 
-Example potentially dangerous page template snipper::
+Example potentially dangerous page template snippet::
 
   <figcaption tal:content="structure context/image_caption" />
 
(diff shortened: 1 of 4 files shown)


svx:
Great !

Now have have to make sure, to follow: http://docs.plone.org/about/rst-styleguide.html and then we are good to go :)

For example code-blocks :) For python we use

```
.. code-block:: python

    if "foo" == "bar":
        # This is Python code
        pass
```

Thanks for the great work ! Awesome ! Thanks a lot !

svx:
This PR should be merged into 5.1 too !

polyester:
shall I merge and cherry-pick into 5.1?

svx:
If my last commit got reviewed and is OK I would say yes

hvelarde:
don't merge it; squash it!

develop/plone/security/guidelines.rst
```diff
@@ -0,0 +1,116 @@
...
+        def foobar(self):
+            """My doc string. This method is public!"""
+            self.x = 'foobar'  # mutation
+
+Do this:
+
+.. code-block:: python
+
+    class MyContent(DexterityItem):
+        def foobar(self):
```
hvelarde, line 62: I would remove this example in favor of the second: removing the **docstring** is not a good practice for introspection; declaring explicitly access control is good for security.

develop/plone/security/guidelines.rst
```diff
@@ -0,0 +1,116 @@
...
+    $('body').append($el);
+
+
+In underscorejs templates make sure to use:
+
+.. code-block:: javascript
+
+    <%- … %>
+
+Do not(underscorejs):
```
hvelarde, line 109: I don't understand this; is this a typo?

develop/plone/security/guidelines.rst
```diff
@@ -0,0 +1,116 @@
...
+
+
+Manual filtering
+~~~~~~~~~~~~~~~~
+
+If you need to use structure with unfiltered input, you can manually run Plone's output filtering engine on arbitrary html.
+
+Example::
+
+    from plone import api
```
hvelarde, line 30: missing `.. code-block:: python` here

develop/plone/security/guidelines.rst
```diff
@@ -0,0 +1,116 @@
...
+
+Use of structure in page templates
+----------------------------------
+
+Do not use the `structure` page template feature with unfiltered input.
+This can lead to XSS attacks.
+
+Example potentially dangerous page template snippet::
+
+  <figcaption tal:content="structure context/image_caption" />
```
hvelarde: missing `.. code-block:: xml` here

develop/plone/security/guidelines.rst
```diff
@@ -0,0 +1,116 @@
...
+
+    var value = '<script>alert("hi")</script>';
+    $('body').append($(value));
+
+
+By default, jQuery is not safe. To do the previous example in jQuery, you could:
+
+.. code-block:: javascript
+
+    var $el = $('<div/>');
```
hvelarde, line 97: what about this?

``` javascript
var $el = $('<script/>');
var value = 'alert("hi")';
$el.text(value);
$('body').append($el);
```

develop/plone/security/guidelines.rst
```diff
@@ -0,0 +1,116 @@
...
+            # My doc string.
+            self.x = 'foobar'  # mutation
+
+Or this:
+
+.. code-block:: python
+
+    class MyContent(DexterityItem):
+        security = ClassSecurityInfo()
+        security.declarePrivate('foobar')
```
davisagli, line 72: Better to use the decorator (@security.private) because it's not prone to typos or forgetting to change the name passed to `declarePrivate`

davisagli:
This looks good.

Might be good to have something about CSRF, recommending to use forms properly, put the authenticator in the URL, or allow particular objects to be written rather than disabling CSRF checks entirely.

This is more basic, but maybe something about making sure that views have appropriate permissions.

Warn against doing a redirect to a URL that came from a request (e.g. in a `came_from` parameter) without validating that it's known to be within the site. This can be used for phishing.

Couldn't hurt to remind people not to unpickle untrusted input.

Point to the OWASP top ten as another resource?

develop/plone/security/guidelines.rst
vangheem, line 62: The security team would disagree with you. We've explicitly been removing doc strings even when it is probably okay to have in some cases. Less of a chance someone screws up.

hvelarde, line 62: yes, I saw that; I know that any method that has a **docstring** becomes public by default, I just didn't understandd why you removed the **docstrings** instead of declaring them private.

svx:
Question to all, does someone has the time and is in the mood to improve this PR with the suggestions made by @davisagli or should we merge for now ?

Thanks !

svx:
So reminder: Question to all, does someone has the time and is in the mood to improve this PR with the suggestions made by @davisagli or should we merge for now ?

Thanks !

develop/plone/security/guidelines.rst
```diff
@@ -0,0 +1,120 @@
...
+        def foobar(self):
+            """My doc string. This method is public!"""
```
thet, line 74: With the security decorator above, this method is private now.

davisagli: It's still publishable though, so it'll respond with a 401 rather than a 404. I think.

loechel:
@svx merge it now.

If any additional discussion or work is needed we could do another pull request.
It has been to long open, to not merge it.

svx:
OK, does someone has time to rebase and merge ?

gforcada:
@svx I had some time so I rebased (merging nearly all commits into a single one) and merged into 5.0 and cherry-picked into 5.1.

PR #667, opened 2016-07-20
Fixed link to bobtemplates.plone in types documentation.

mauritsvanrees:
Fixes: warning when running `make html` in papyrus:

```
.../papyrus/source/documentation/develop/plone/content/types.rst:197:
WARNING: unknown document: /develop/addons/bobtemplates.plone
```

### DIFF
--- a/develop/plone/content/types.rst
+++ b/develop/plone/content/types.rst
@@ -194,7 +194,7 @@ Dexterity
 
 * Use Dexterity control panel in site setup
 
-* Use :doc:`bobtemplates.plone </develop/addons/bobtemplates.plone>`
+* Use :doc:`bobtemplates.plone </develop/addons/bobtemplates.plone/README>`
 
 Plomino (Archetypes-based add-on)
 ---------------------------------
@@ -286,4 +286,3 @@ The default addable types are the types that are addable when
 ``constrainTypesMode`` is ``0`` (i.e not enabled).
 
 For more information, see **Products/CMFPlone/interfaces/constraints.py**
-


svx:
@mauritsvanrees thanks :) but even if it works for now, this is still wrong :) because we have to fix the build script in papyrus. 

Do not get me wrong I am really thankful for every contribution you rock !!. 

But unfortunately we are linking now to the wrong file, we are linking to the README now, but we do not want to link to READMEs, in case of this we want to link to: http://docs.plone.org/develop/addons/bobtemplates.plone/bobtemplates.plone/docs/.

Yes, the link I just mentioned looks currently that bad, it should be:    http://docs.plone.org/develop/addons/bobtemplates.plone, so we should strip the second bobtemplates.plone and the docs part.

We have this because the script we currently use: https://github.com/plone/papyrus/blob/5.0/assemble-docs.tpl.genshi got merged and put into use without proper testing.

Because of this and because of people not following the docs guidelines we have this situation now, I actually tried to fix that for bobtemplates here: https://github.com/plone/bobtemplates.plone/commit/314f8cfa2b290101dce9c1512ee2c3263671a248

Thanks for the 'dirty' quick fix solution I will try to find some time to fix this mess ones and for all :)

mauritsvanrees:
Related: https://github.com/plone/papyrus/pull/106 which I opened around the same time, but have now closed. And I think you switched me and @fredvd around in a mail you sent him Wednesday related to this. :-)
You seem to have a good idea on how to solve this properly, so I'll let you do your docs rock star thing. ;-)
And we'll catch up during the [Plone planning and organizing spring](http://www.coactivate.org/projects/munich2016/summary) next week.
BTW, I will be offline until then.

PR #696, opened 2016-09-20
Document QI deprecation in 5.1 upgrade guide. PLIP 1340.

mauritsvanrees:
This documents the approved and about-to-be-merged Plone 5.1 PLIP [1340](https://github.com/plone/Products.CMFPlone/issues/1340)

This seems the first upgrade doc fro Plone 5.1, so I added a new document.

In the Plone 5.0 upgrade doc I changed the title to specifically mention Plone 5.0 instead of just Plone 5.


hvelarde:
@mauritsvanrees this is awesome and has to be pointed as an example of how a module must be deprecated.

I still have a couple of doubts:
- when installing code using GS I see must of the time that some steps are imported multiple times during the process, how can we avoid that?
- in almost all of our add-ons we didn't add uninstall code for some of the features we install using GS; can we get a list of those things that will need a XML file? `cssregistry.xml`, `jsregistry.xml`, `layer.xml`, `types.xml`…?

mauritsvanrees:
Steps imported multiple times: I have not seen this, except that when profile A gets installed and it depends on profile B, and they both have a rolemap file, then you will see the step imported twice. And custom import steps (or the pre/post_handlers) may explicitly run some steps again, for example an import handler that explicitly applies the `catalog.xml` before doing its own stuff.

In general, if your default profile has a `foobar.xml` file, then your uninstall profile needs a `foobar.xml` file that undoes the install.

For the stuff that used to be auto-removed by `portal.quickinstaller`, see the [`snapshotPortal`](https://github.com/plone/Products.CMFQuickInstallerTool/blob/master/Products/CMFQuickInstallerTool/QuickInstallerTool.py#L423) method.
The code for removing it is in `InstalledProduct.py` in the [`_cascadeRemove`](https://github.com/plone/Products.CMFQuickInstallerTool/blob/master/Products/CMFQuickInstallerTool/InstalledProduct.py#L305) method.

So that's:
- leftslots and rightslots in the portal root, so old-style portlets. If you are still adding these, then you are doing something wrong. But you can remove them with `properties.xml`.
- content_type_registry predicates. These are in `contenttyperegistry.xml`. I don't think this is used often, in core I only see it in CMFPlone. Looks like `Products/CMFCore/exportimport/contenttyperegistry.py` only supports adding, not removing.
- objects in `portal_types`, use `types.xml` to remove them.
- objects in `portal_skins`, use `skins.xml` to remove them.
- actions in `portal_actions`, use `actions.xml` to remove them.
- objects in `portal_workflow`, use `workflows.xml` to remove them.
- objects in the portal root, use `toolset.xml` to remove them.
- adapters and utilities, use `componentregistry.xml` to remove them.
- resources in `portal_javascripts`, use `jsregistry.xml` to remove them.
- resources in `portal_css`, use `cssregistry.xml` to remove them.

I will add some of this to the doc.

develop/addons/upgrade_to_51.rst
```diff
@@ -0,0 +1,395 @@
...
+
+New code:
+
+.. code-block:: python
+
+    qi.install_product('Products.CMFPlacefulWorkflow')
+
+.. @FWT: When a Product registers its profile in Python code instead of zcml, it might get registered without ``Products.`` in the name.
+   Do we still wish to support this?
+   My vote: no.
```
gforcada: @mauritsvanrees should this be cleaned up before merging?

develop/addons/upgrade_to_51.rst
```diff
@@ -0,0 +1,395 @@
...
+
+New code:
+
+.. code-block:: python
+
+    qi.install_product(product_name)
+
+Note that no keyword arguments are accepted.
+
+.. @FWT: We might keep ``omit_snapshots`` and ``blacklisted_steps`` if we really want.
```
gforcada: ditto

develop/addons/upgrade_to_51.rst
mauritsvanrees: Ah, yes, I thought I had removed all such comments already.
Done now, also for the second one.

svx:
Awesome ! Thanks ! Do you want me to merge ?

mauritsvanrees:
Yes please. Later I will probably think of more little things to add, but for now I would say it is good to go.

PR #704, opened 2016-10-05
Update outdated link

staeff:
Fixes: 
- a broken link

I didn't find a really good replacement for the oudated link, but I think this one is close enough and saves the reader a search.


svx:
@staeff would you mind to create another PR for the 5.1 branch ?
Oh ... and thanks !!

staeff:
@svx a question for the next time: Do both branches need to be updated or should everything go to the 5.1? 
Thanks!

svx:
@staeff it depends :) 
If the content/link/fix/addition is valid for 5.0 and 5.1 it should go into both branches ! If it is even valid for the 4 branch, we also should update it there.
:)

PR #708, opened 2016-10-26
Upgrade: no large version leaps.

mauritsvanrees:
In https://github.com/plone/plone.app.upgrade/pull/91 we removed the Plone 2.5 and Plone 3 upgrade code from plone.app.upgrade.
This pull request reflects this in the documentation: you should only upgrade to one higher major version at a time.
Technically you can still upgrade from 2.5 to 4.3 in one go, but we feel this is not advisable, so I did not mention this in the documentation.

### DIFF
--- a/manage/upgrading/intro.rst
+++ b/manage/upgrading/intro.rst
@@ -36,3 +36,17 @@ Since the Plone 2.5 release, Plone has a policy that increases the version numbe
 
 In addition to the general procedure there are :doc:`version-specific migration guides </manage/upgrading/version_specific_migration/index>`. These guides contain more specific instructions and valuable information that has been collected from real-life migration cases.
 
+
+No large leaps
+==============
+
+It is advisable to not make large leaps in version numbers.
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
I probably don't need to convince anyone, but let me give an example of why it is better to not make large version leaps. Today I tried upgrading from Plone 3.1 to 5.0.6 in one go. Technically that is still possible, and after a few fixes in plone.app.upgrade it worked.

But during the blob migration in 4.0, I got errors for all the images that got migrated, because some plone.i18n code got called, which in the Plone 5 version tries to get the `IRegistry` utility, which fails because in 4.0 this did not exist yet. The migration continued, but it did not feel clean. It could be that the images were still stored in the Data.fs and were not turned into blobs.

A migration to a buildout with 4.3 and then a buildout with 5.0.6 worked fine.

jensens:
@mauritsvanrees may you please apply this to 5.1 as well?

gforcada:
@jensens @mauritsvanrees I already did that, was just one commit after all :)

PR #710, opened 2016-10-31
minor overhaul of z3c.form validators section

jensens:
Improve docs about z3c.form validators.

Could be even better, but that for later. 

This here is enough to not confuse people more than needed ;)

### DIFF
--- a/develop/plone/forms/z3c.form.rst
+++ b/develop/plone/forms/z3c.form.rst
@@ -2169,8 +2169,6 @@ Validators
 Introduction
 ------------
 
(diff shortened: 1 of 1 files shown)


jensens:
Travis says: `Error: Couldn't install: urllib3 1.18.1`

I dont think this is because of my change. Is there something broken?

/cc @plone/documentation-team

svx:
No, it is not because of your change, it is because of old version pins in papyrus, I am working currently on the new search integration, merging the PR with the fix and other updates is on the list too, I'll hope to push all the fixes upcoming weekend.

Thanks for your contribution to the docs !

PR #724, opened 2016-11-16
Add: portal properties to site properties by specifying a missing key

svx:
Fixes: #

Improves: Docs about upgrading

- update docs about portal properties to site properties by specifying a missing key

-

-


polyester:
cherrypicked into 5.0

PR #741, opened 2016-12-17
Update askforhelp.rst

tkimnguyen:
wording tweaks, genericize chat, include Gitter, update links, add tips

Fixes: #

Improves:

-

-

-

### DIFF
--- a/askforhelp.rst
+++ b/askforhelp.rst
@@ -70,6 +70,8 @@ Asking for help in online chat
 
 The Plone community uses `online chat <https://plone.org/support/chat>`_, specifically `Gitter <https://gitter.im/plone/public>`_ and traditional Plone IRC chat at the  `#plone <irc://irc.freenode.net/#plone>`_ IRC channel on the freenode IRC network.
 
+(To use Gitter you must log in using a free GitHub or Twitter account, which allows you to receive notifications when there is a response to your question. IRC does not require you to log in, but you will receive notifications of responses only while you remain connected).
+
 Here are useful hints for using online chat:
 
 - Remember that chat participants are volunteers; they are not paid to provide support.
--- a/askforhelp.rst
+++ b/askforhelp.rst
@@ -6,21 +6,21 @@ Asking for help
 Guidelines and Examples
 =======================
 
-If you need help with an error or problem: before asking the question, please take a few minutes to read the guidelines below. It is important to know how to state questions, because once you learn it, answers will come much faster.
+If you need help with an error or problem: before asking the question, please take a few minutes to read the guidelines below. It is important to know how to state questions, because once you learn it, you will get better answers back more quickly.
 
 
 Asking help on support and discussion forums
 --------------------------------------------
 
-By stating a **well-phrased question** you increase the likelihood of fast and helpful response to your question.
+By stating a **well-phrased question** you increase the likelihood of fast and helpful responses to your question.
 
 Here are some general key rules users need to follow in creating a new topic.
 
-- ALWAYS start with searching, that means, before you create a topic. Most of your questions were probably already answered by someone else in the past. Save your and our time by searching the web first.
+- ALWAYS start with searching before you ask a question. Most of your questions were probably already answered by someone else in the past. Save your and our time by searching the web first.
 
 Where to search:
 
-- Google - Before asking for help, make a Google search with related keywords. Pick meaningful keywords from the log entry.
+- Google - Before asking for help, make a Google search with related keywords. Pick meaningful keywords from the log entry. Sometimes searching for the entire error message works!
 
 - `Plone Community Forums <https://community.plone.org/>`_ - help and discussion forums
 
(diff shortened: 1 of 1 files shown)


askforhelp.rst
```diff
@@ -63,17 +63,16 @@ The big picture - An opening sentence should state the general problem that you
...
-**Steps to reproduce the issue** - Give information about your ideas of how this error appeared, what caused it or anything that could lead to reproducing the error on another computer. The expected result .
+**Steps to reproduce the issue** - Give information about your ideas of how this error appeared, what caused it or anything that could lead to reproducing the error on another computer, including your buildout.cfg, versions.cfg, version numbers of installed add-ons, detailed command lines, the complete error message stack. Mention your expected result.
 
-Asking help in real time chat (IRC)
------------------------------------
+Asking for help in online chat
+------------------------------
 
-Plone community real time chat is at the  `#plone <irc://irc.freenode.net/#plone>`_ IRC channel on the freenode IRC network.
+The Plone community uses `online chat <https://plone.org/support/chat>`_, specifically `Gitter <https://gitter.im/plone/public>`_ and traditional Plone IRC chat at the  `#plone <irc://irc.freenode.net/#plone>`_ IRC channel on the freenode IRC network.
```
svx, line 72: IMHO if we want to mention Gitter, there should be more information about it, like that you will need a GitHub or Twitter account to sign in.

svx, line 72: IMHO if we want to mention Gitter, there should be more information about it, like that you will need a GitHub or Twitter account to sign in.

svx: Next question why Gitter, there are also people on Slack so what is the idea behind adding Gitter but not Slack ?

IMHO I would be really careful with adding more and more options here, we should make it 'easy' for user and thsi means add the option(s) which the most people are using.
Otherwise you could end up in one channel which is not active at all.

So yeah I am not against it, but like I said we should be careful with splitting our own resources here.
It should not about what you or me personally like, it should be about which one is 'easy' and used by the most.

stevepiercy:
FWIW, Pyramid and other projects under the Pylons Project went through the same thing. We debated over issues such as open vs. closed source, account required or not, spreading ourselves thinly. Ultimately we opted for email via Google Groups and IRC as our two official support channels, both of which are open and archived.

askforhelp.rst
tkimnguyen: It does say that in the room's welcome message. The canonical info about Gitter is at plone.org/support/chat so I didn't just want to repeat the info and risk having it get outdated eventually.

tkimnguyen: I agree and understand. We went through this discussion and ended up deciding to leave Slack fans on their own and able to use Slack (they continue to do so), but for official chat we leave IRC for fans of that and we introduce Gitter because it's friendly to open source projects, doesn't have the problems Slack has caused for open source projects, and it authenticates with GitHub, which makes that match plone.org and our code repos and issue trackers. Also, Gitter's requirement for authentication lets us respond to questions & notify people after they'd normally have disconnected from IRC (something that drives me bananas not just personally but also from the community perception perspective).

In other words, we preserve choice and freedom but steer newcomers to Gitter if they don't like or find IRC unapproachable (which we've heard a lot of).

tkimnguyen:
thx @stevepiercy - our equivalent to Google Groups is https://community.plone.org where we have search and archiving

askforhelp.rst
svx: Would you mind to point me to the place where "we" made the decision? It seems I missed that, I only can find postings on community where people discus that.

Besides from my own opinion why this is *really bad* I will try to explain it in a hopefully different way, sorry that is getting 'longish' now :).

Please try to think not as Kim the developer but in place of that as Kim the webmaster who is new to Plone and never used it before.
So you are a webmaster, you do not have programmer knowledge, you do not write code, your work environment is your browser and Libre Office.

You have a question about a setting in the CMF, you can't find the answer in the docs (bad bad) but hey there is this chat mentioned on docs.plone.org/askforhelp. Hopefully there people will able to answer me my question, yeah !
Your click on the link and you end up on the Gitter page, nice, but uh eh I can see there is a chat and I can see people talking but I can not ask a question, first I have to login ...

Ok lets click on sign in ... uh they are asking for a GitHub or Twitter login .. What is GitHub never heard about it, lest check oh for developer and nerds, ok do not need nor have that .... Twitter? Hm .. I will not use my personally account to login for work related stuff, and I am not allowed to use a company account ....

So hey, looks nice but I am not able to use that in anyway ...

Thanks for staying with me !

What I am trying to say, Gitter maybe handy for you and I can see the point of have all the nice GitHub hooks, but I doubt that is that useful for 'normal' people.
So IMHO you make it harder for people to ask for help. If you are a 'developer',a person who has already a GitHub account, yes it maybe helpful, great but for others not at all, you may even push them away, which is bad for Plone.

That is exactly the reason why we thought about this deep and long for docs.plone.org and went with a really 'simple' widget, which people can click and just can write us or attach pictures, no need for a login somewhere.
Again ... if you want to be serious about helping people you also should make it easy and in a way that it is working for them, this may involve slightly more work you you, but like I said it is not about you it is about user.
Going further this is also about marketing, if you make it hard for people you will loose them to other Solutions and you will get knows as hard to use and impossibly to get help.

If I went for all stuff what and how I would like to do docs.plone.org they would be completely different today, I did not do that because it would be 'harder' for people to contribute, so I did not do that, yet ......(you never know), hehe
I you feel that you get 'bananas' the way how IRC works and people disconnect, maybe the discussion should be more 'do we want a chat as support' at all ?

So in short I am serious concerned and against merging that in the current state. If this would get re-written in a way, where it is better and more explained that Gitter requires to login and also the reason why 'we' choose for that, I may fine, but sorry not how it is now.

@tkimnguyen please do not take it personally, this is **not meant** to insult you at all ! I am trying to make it better for less technical people because I am concerned that we 'forget' them again and I am concerned about the way Plone is moving .....

svx: see my long answer above :)

tkimnguyen:
@svx I don't really care to argue this further in here. We want to improve the current state of chat, we have done that, and now we need to update docs to reflect that. If you don't want to merge this then don't but be aware that the page as it stands has a bunch of other issues.

svx:
@tkimnguyen your current PR is IMHO not improving the situation, that is what I am trying to say, currently you will leave the person who is looking for help with various chat possibilities [IRC and Gitter] without giving or telling a clear direction and because of this this is confusing which one should the user choose, like I tried to explain above, if you really like to add Gitter you should make sure to explain why this could be an advantage over IRC, like *if you already using GitHub* bla bla bla.
But there should be also explained that you need to login with GitHub/Twitter.

So maybe it would be better for all of us [community] to define first what kind of support/chat we want to provide to/for which audience, this is what I am still missing, we as community have currently, at least so it looks to me no idea about who/what/how ...

The point if using Gitter is an improvement is a completely different topic... I am still serious concerned about Gitter and the reasons why and how the decision was made ....

tkimnguyen:
@svx I hope I've clarified it based on your suggestions. I appreciate and can act on specific suggestions.

svx:
Needs to cherry-picked into 5.1

svx:
Thank you !

svx:
Merged into 5.1 !

PR #558, opened 2016-03-21
More context on backend.xml in barceloneta so themers understand when to use it

djay:
Fixes #.

Improves:

Changes proposed in this pull request:
## 
## 
##

PR #746, opened 2017-01-03
Give more information about backend.cfg

gforcada:
Follow up pull request from https://github.com/plone/documentation/pull/558

It should be cherry-picked to 5.1 as well.


gforcada:
@svx so is it ready for merge?

svx:
Yeah, guess so, still not perfect but well if we wait till prefect I fear we will never merge it :)

So yeah, I say lets merge !

gforcada:
Pushed to 5.1 as well!

PR #749, opened 2017-01-04
Remove references to Varnish 3

hvelarde:
On April 10th 2015, Varnish Cache 3.0 reached end of life (EOL) status.

https://www.varnish-cache.org/lists/pipermail/varnish-announce/2015-April/000702.html

Makes no sense to keep the documentation on it here.


gforcada:
Pushed to 5.1 branch as well.

hvelarde:
@gforcada you're too fast :-)

I pushed a new commit; the documentation is a mess and is not easy to organize it.

I'll try to do so, little by little.

svx:
I guess this also needs to be done for the 4 docs ?

PR #751, opened 2017-01-06
Fix language check. Resolves #537

### LINKED ISSUE #537: Bypassing language check seems wrong
From https://github.com/plone/documentation/blob/5.0/develop/plone/searching_and_indexing/query.rst#bypassing-language-check 

> All portal_catalog() queries are limited to the selected language of the current user. You need to explicitly bypass the language check if you want to do multilingual queries.
> 
> Example of how to bypass language check:
> 
> all_content_brains = portal_catalog(Language="")

This seems wrong with Plone.app.multilingual, perhaps its true with LinguaPlone?  From my experiment below, `catalog.searchResults({'Language': ''})` just shows you content with no language set.  You'll have to take my word on that of, but you _can_ see that a catalog search with no parameters returns ALL content in the site (the 3 languages below are the only ones enabled).

```
(Pdb) len(catalog.searchResults({'Language': ''}))
30
(Pdb) len(catalog.searchResults({'Language': 'en'}))
107
(Pdb) len(catalog.searchResults({'Language': 'de'}))
17
(Pdb) len(catalog.searchResults({'Language': 'pt'}))
4
(Pdb) len(catalog.searchResults())
158
(Pdb) 
```

What DO we want to say here?

djowett:
First offering on fixing the above issue.  This needs review for technical accuracy - in particular that the Language parameter works for LinguaPlone.  I thought it should be the following:

    portal_catalog(Language="all")

But since I wasn't able to prove either that or Language='' works, I'm leaving the LinguaPlone part as it was.

@erral @jensens

Fixes: #537 corrects "Bypassing Language Check" in to consider plone.app.multilingual in addition to LinguaPlone

Improves:

-

-

-

### DIFF
--- a/develop/plone/searching_and_indexing/query.rst
+++ b/develop/plone/searching_and_indexing/query.rst
@@ -436,15 +436,32 @@ the objects without triggering a security exception::
 Bypassing language check
 ========================
 
-.. note::
+.. note ::
 
         All portal_catalog() queries are limited to the selected language of
         the current user. You need to explicitly bypass the language check if you
         want to do multilingual queries.
 
-Example of how to bypass language check::
+Language is only a factor when a multilingual product is installed - which 
+basically comes down to one of the venerable ``LinguaPlone`` or the more modern 
+``plone.app.multilingual``. Bypassing the language check depends on which of 
+these you are using.
+
+In LinguaPlone and plone.app.multilingual 1.x (what you would probably use in 
+versions 4.3 or earlier of Plone), a patch is applied to the portal_catalog.  
+To bypass this add the parameter ``Language='all'`` to your catalog query like
+so::
 
     all_content_brains = portal_catalog(Language="")
+
+``plone.app.multilingual`` creates Root Language Folders for each of your site's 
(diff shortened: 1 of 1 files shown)


erral:
That's right, the Language='all' is needed in LP and PAM 1.x

https://github.com/plone/Products.LinguaPlone/blob/master/Products/LinguaPlone/patches.py#L30

https://github.com/plone/plone.app.multilingual/blob/1.x/src/plone/app/multilingual/catalog.py#L26

Although the language folders are also marked to be INavigationRoot, in LP the language of the content is not enforced inside the language folder (in PAM there's a subscriber that moves the content to the appropriate folder)

djowett:
Thanks @erral. I thought your last paragraph was worth adding too, of you don't mind?

Ready to go IMO.

erral:
:+1:

svx:
@djowett if that commit is valid for Plone 5.1, please cherry-pick it into the 5.1 branch !

Thanks !

djowett:
@svx done! :-)

PR #754, opened 2017-01-24
Strengthen recommendation against using bare zope http for production

smcmahon:
Make it clear that using bare zope http for production is not an accepted practice.


svx:
@smcmahon Thanks a lot !!!  Awesome !! Would you mind to cherry-pick that into the 5.1 branch ?

svx:
This needs to be picket into 5.1

smcmahon:
Cherry picked into 5.1 and 4.3

svx:
Thanks a lot !!!

Issue #690, opened 2016-08-31
for Plone version upgrades, recommend deactivating unused add-ons


tkimnguyen:
...if they're not going to be used in the upgraded sites, because after the upgrade it will be impossible to deactivate them

svx:
I guess this: http://docs.plone.org/manage/upgrading/addon_upgrade.html would be a good place to put it ?

PR #761, opened 2017-01-31
Fix: #690

svx:
Fixes: #690 

Improves: Docs on upgrading

- Fix wording

- Improve reST

- Add info, suggested by @tkimnguyen


polyester:
and cherry-picked into 5.0

tkimnguyen:
thx but I don't recall now what I might have suggested :) - ah, ok I see now!

Issue #772, opened 2017-02-07
Error in: Overriding a view class in a product


lukebrannon:
The Plone 5 documentation for _Overriding a view class in a product_ <http://docs.plone.org/develop/plone/views/browserviews.html#overriding-a-view-class-in-a-product> results in a TypeError:  page() takes at least 4 arguments (5 given).

The error is caused by the layer attribute in the zcml for browser:page
Referencing older packages zcml should be?:

<configure
        xmlns="http://namespaces.zope.org/zope"
        xmlns:browser="http://namespaces.zope.org/browser"
        i18n_domain="example.policy">
      <browser:page
          name="register"
          class=".customregistration.CustomRegistrationForm"
          permission="zope2.View"
          for="Products.CMFPlone.Portal.PloneSite"
          />
    </configure>

> zope.configuration.xmlconfig.ZopeXMLConfigurationError: File "/home/someone/Plone/plone51/baseInstance/parts/instance/etc/site.zcml", line 16.2-16.23
    ZopeXMLConfigurationError: File "/home/someone/Plone/plone51/buildout-cache/eggs/Products.ATContentTypes-2.2.13-py2.7.egg/Products/ATContentTypes/configure.zcml", line 18.2-18.44
    ZopeXMLConfigurationError: File "/home/someone/Plone/plone51/buildout-cache/eggs/plone.app.collection-1.2.1-py2.7.egg/plone/app/collection/configure.zcml", line 9.2-9.45
    ZopeXMLConfigurationError: File "/home/someone/Plone/plone51/buildout-cache/eggs/plone.app.querystring-1.4-py2.7.egg/plone/app/querystring/configure.zcml", line 11.2-11.42
    ZopeXMLConfigurationError: File "/home/someone/Plone/plone51/buildout-cache/eggs/plone.app.registry-1.3.12-py2.7.egg/plone/app/registry/configure.zcml", line 12.4-12.34
    ZopeXMLConfigurationError: File "/home/someone/Plone/plone51/buildout-cache/eggs/plone.app.registry-1.3.12-py2.7.egg/plone/app/registry/browser/configure.zcml", line 6.4-6.43
    ZopeXMLConfigurationError: File "/home/someone/Plone/plone51/buildout-cache/eggs/plone.app.z3cform-2.1-py2.7.egg/plone/app/z3cform/configure.zcml", line 10.2-10.41
[...]

PR #773, opened 2017-02-07
update zcml to use for attribute rather than layer

lukebrannon:
Fixes: #772 

Improves:
Updates documentation with correct zcml xml.

- use for attribute in browser:page instead of layer when overriding class view

-

-

### DIFF
--- a/develop/plone/views/browserviews.rst
+++ b/develop/plone/views/browserviews.rst
@@ -590,7 +590,7 @@ original.
           />
     </configure>
 
-.. Note::
+Note::
 
         We've retained the permissions and marker interface of the original view.
         You may provide a specific permission or marker interface in lieu of these
--- a/develop/plone/views/browserviews.rst
+++ b/develop/plone/views/browserviews.rst
@@ -585,7 +585,7 @@ original.
           name="register"
           class=".customregistration.CustomRegistrationForm"
           permission="zope2.View"
-          layer="..interfaces.IExamplePolicy"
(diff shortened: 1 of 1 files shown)


thet:
Defining a layer should be possible for ``browser:view`` as well as ``browser:page`` until zope.components 4.0 (https://github.com/zopefoundation/zope.component/blob/master/CHANGES.rst).
What version of zope.component are you using?

MrTango:
https://training.plone.org/5/mastering_plone/views_2.html?highlight=layer#the-default-view

develop/plone/views/browserviews.rst
```diff
@@ -584,11 +584,18 @@ original.
...
+          permission="cmf.AddPortalMember"
+          for="plone.app.layout.navigation.interfaces.INavigationRoot"
+          layer="example.policy.interfaces.IExamplePolicy"
           />
     </configure>
 
+Note::
+
+        We've retained the permissions and marker interface of the original view.
+        You may provide a specific permission or marker interface in lieu of these
```
svx: Can we replace *lieu* with *instead of* ?

*Lieu* is not wrong but *instead of* may be easier to understand, we have to keep in mind that our audience is from all over. :)

stevepiercy: What do you have against French? 😜 

"instead of" is more common than the fancy pants "in lieu of", so I would agree.

lukebrannon: @svx updated :)

PR #781, opened 2017-02-16
Document Retina PLIP #1483.

mauritsvanrees:
This documents [PLIP #1483](https://github.com/plone/Products.CMFPlone/issues/1483) for Retina images. The PLIP has been accepted, but has not been merged yet, so hold off on merging this documentation change.


fgrcon:
As far as I know, *Retina display* is very Apple specific (maybe even a registered tm ?).  Would'nt it be more neutrally better called HiDPI  or the like?

PR #791, opened 2017-03-05
fix link and note markup

tkimnguyen:
Fixes: #

Improves:

- The old note wasn't displaying the external link correctly, so changed note according to the ReST style guide at https://docs.plone.org/about/rst-styleguide.html

- Fixed link to mockup repo


svx:
Thanks !

Issue #726, opened 2016-11-23
Add Ubuntu 16.04, remove 12.04 from the docs


svx:
Currently we still mention Ubuntu 12.04 in the installation docs, since 12.04 will be eol in April 2017, and 16.04 is now working, I would like to:

- remove 12.04 from the docs
- add 16.04 next to 14.04

gforcada:
One of the big changes on 16.04 is that python 2.7 is no longer installed by default and that /usr/bin/python may point to python3.5 as well.

svx:
OK, but this is not a problem, we add that, we have currently docs for 12.04 and 14.04 which are also different.

But anyway thanks for the reminder ! :)

svx:
done !

PR #800, opened 2017-03-14
Fix: #726

svx:
Fixes: #726 

Improves: Updates installation by adding Ubuntu 16.04

- change 'apt-get' to 'apt' [v2] which is working already since 14.04 ootb

- add packages for 16.04, including python2.7 and python2.7 dev

Since I am not using ubuntu as OS, it would be nice if someone who does could look over it, I did a quick test in a VM but maybe I forgot something :)

### DIFF
--- a/manage/deploying/preparing.rst
+++ b/manage/deploying/preparing.rst
@@ -1,8 +1,9 @@
+===================
 Preparing the Server
 ====================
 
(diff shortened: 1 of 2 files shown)


svx:
Again enchant ... great, time to change our travis setup.

polyester (approved):

PR #809, opened 2017-03-19
adding documentation for thumbs/icons ....

fgrcon:
**WIP dont merge yet!**
(branched from 5.1)

Fixes: documentation for https://github.com/plone/Products.CMFPlone/issues/1734 and 
                                         https://github.com/plone/Products.CMFPlone/issues/1226

**also changed related docu in:**

https://github.com/plone/plone.app.dexterity/pull/240
https://github.com/plone/plone.app.contentlisting/pull/19
-

-

-


svx:
Hi there !
You are aware of #781 ?

I did not checked yet, but I wanted to ask to make sure that people do not work twice on related or same parts :)

On another note, why do you use ``figure`` for images ?

You can use ``image`` -> https://docs.plone.org/about/rst-styleguide.html#images

:)

fgrcon:
@svx  cc: @mauritsvanrees , @jensens, @ebrehault 
 yes I absolutely  am aware of  #781 .  At first glance both plips should not interfere (thumb-icon ... is a matter of better ui means to control where icons and/or thumbs should be shown and if  thumbs are shown in which size. The retina (oops thats sounds apple...)  plip focusses on the pixel resolution of shown images only. 

But unfortunately  both  plips will need some merging related to the final rendering  of the images (using
 ... scale ... instead of simple  '... /@@images/image/'+thumb_size_list|table|portlet ..').

Contributing to documentation seems a little cumbersome to me - restructured text in a wysiwyg-world,  furthermore I am often not quite sure about the structure:  I usually prefer different levels of docu for different stakeholders, so at least general overview, user documentation ( could be refined into contributors, admins, etc) and developers/maintainers. This levels are often not easy to locate in the current plone documentation. 

I started  with general overwiew and will try to add info for other levels and I  will keep an eye on retina/HiDPI plip as well  :) (btw: this also  could do with some general explanations) 

Why i used ``figure`` ? Its simple: I just copied tags from the beginning of the file i edited (git blames @polyester for that :-) ) - Does this make much difference ?

By the way feel free to comment or even  better improve  and restructure as needed  :-).

polyester:
@fgrcon hey I was young and needed the money ;-)

on a slightly more serious note: `figure` is more suited to proper scientific diagrams. `image` is the way forward for simple illustrations and screenshots.

fulv:
OTOH `figure` allows a subordinate `figcaption` element.  I am not aware of a similar element for `image` specifically for adding a caption.

fgrcon:
please see top

PR #825, opened 2017-04-20
Updated add-ons/downloads link

sgrepos:
Fixes: #

Improves:

- Downloads link was linking to https://plone.org/products/plone, which redirects to https://plone.org/download. Updated to a direct link, since it's faster.

- Addons link was pointing to downloads page, so I updated it to the direct page for add-ons at https://plone.org/download/add-ons.


svx:
Thanks !

PR #857, opened 2017-04-27
Add deployment on Digital Ocean tutorial

malloxpb:
Improves: Add the deployment on Digital Ocean tutorial right after deployment on Ubuntu VPS to illustrate a deployment method for users. I have fixed the indentation by following the ReST style guide in the Documentation. If there is any other mistakes, please let me know so I can continue working on fixing them. 
In addition, I think this is necessary since I think a lot of people who just start to use Plone might struggle with deployment (testing env is important for both developers and clients). A suggestion might be a good solution to that.
If there is any other better location for this tutorial, I won't hesitate to start working on migrating this PR :).


manage/deploying/production/digitalocean_deployment.rst
```diff
@@ -0,0 +1,151 @@
...
+When you finish step 5, you should have your Plone app running at [your-host-name].com.
+
+
+Common errors
+=============
+
+Cannot install lxml==3.5.0
+--------------------------
+
+**Traceback**::
```
hvelarde, line 117: you can use `.. code-block:: console` here.

hvelarde (dismissed):
LGTM

svx:
I am still not sure abut this at all :)

This may seems now as a great idea, but could be really bad in the long term.

Let me explain:

- The first part if this PR is already great and long explained by DO itself in their own docs.
This is IMHO not part of Plone.
- The rest ofthe docs in this PR is already covered in our own installation docs.

If we decide to merge that, it will IMHO increase our own workload, since we need to keep that document up to date, too.

For example say DO is changing their UI or something else.

There is as far as I can see nothing in these PR what we do not have already in our docs.
If we start with adding it for DO, what is coming next, a custom manual for *providerX* ?.

IMHO it would be better to have a good look what we (may) need to improve in our already documentation, than add almost the same docs we already have.

Keeping all these small pieces in sync is really time consuming !

Better less but working and up to date docs, than more and possible fast outdated.

Please also consider how people use docs. Having this for DO but not for others can be lead to confusions.
A IMHO way better solution is to add to our already existing docs a line that all this works on DO, Linode, Rackspace, etc.
If really needed we could use here our ``os-example`` add-on (Sphinx) to highlight and explain the possible differences.  

@polyester what is your take on that ?

svx:
BTW: This PR is not following our style-guide at all ! :)

hvelarde:
so maybe this can be included on DO documentation instead :)

svx:
@hvelarde If that would end up in the official DO docs, yes that would be great !!!

polyester:
If it can be 'outsourced' to Digital Ocean docs, that would be good. 

Otherwise, I agree with @svx, this will get outdated fast, and is very specific to one particular provider.

On the other hand, it is a _tutorial_, which in and of itself has value, and has a different goal. The main benefit being, that people can go through it in one go.

**If** we take it in, I would not put it under /manage/deploying/production, but rather under a (new?) heading "tutorials". Then we could put a clear intro paragraph saying this is valid as of November 2017, and is presented _'as is'._

Haven't completely thought this through, but in short:

- I think it should not go into the main 'deploying' docs
- but it may have value as a 'beginner tutorial', and then should be put under an appropriate section.

all open for discussion of course

svx:
Excuse my impoliteness, *just* something to take in mind :)

IMHO this should be not forgotten if people bring up the idea about tutorials again :)

We once included tutorials and that was/is in the docs of Plone 4, for example:   https://docs.plone.org/4/en/external/tutorial.todoapp/docs/

This is still beating us and creating headaches ! :)

- No one ever took care of making sure it is still working and following best practices
- No upgrade to Plone 5
- Tutorials are written in a different tone of voice, usually even with a different style-guide  :)

People still sending mails (via freshdesk) that the tutorial is not working :).

Because of this and a couple of other reasons, we decided for the docs of Plone 5 to *not* include tutorials or even *new code examples*.

Instead we link to the awesome training documentation, which is updated at least once a year :)

Please, my intention is not to offend anyone, I *just* wanted to remind that we been there before and it may be worth to evaluate carefully.

Thanks ! :)

polyester:
Ok I'm convinced

stevepiercy:
The [Pyramid Community Cookbook](https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/) is another example of what @svx describes. We got great documentation or a tutorial that is valid for a while, then the provider changes their implementation and no one maintains it. The Cookbook is slowly becoming a collection of [links to blog posts](https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/auth/auth_tutorial.html) and other documentation.

We have only supported tutorials and documentation in our official documentation, which we found to be a good line in the sand.

tkimnguyen (commented):
I've read over the comment thread and I have to defer to Sven and Paul on whether this fits with the current docs structure. I do see great value in having this sort of how-to, especially as it helps with Plone adoption. The question is: how could we commit to maintaining this type of installation how-to so it remains up to date for changes to DO and to Plone? For which other deployment platforms should we have such an installation how-to? Linode, Amazon, etc.

tkimnguyen:
Can we recommend or create another place for @nctl144 's contribution to live?

svx:
Ok sorry for repeating myself lets look at it again:

- First it explains how you create a VM on DO.

This is not part of Plone !
Every provider, be it DO, Linode, Amazon etc, have into the deep and up to date docs how you create VM on their platform.
They keep them also up to date, because this is their business.

Starting with this kind of docs on docs.plone.org is like opening pandora's box
- It adds lots of extra work for to little people 
- Who decides which provider we write docs about ?

Looking further at it from a 'marketing' or 'communication' point of view:

I do not see the value, I see unclearness and user confusion, with that we 'just' would add another place with some parts of installation docs, which makes it really hard for user looking for information.
It messes up SEO, may gives you different info than other install docs.

The second part of this pr is like already written above repeating our already existing docs about installation.

Like I also already wrote above, if there is any info in this pr which we are emissing about the install  it would be IMHO better to just include this little parts into our existing docs.
ANd if we really thing it is helpful we also can link to the docs of various could providers about 'VM creation'.
This would be way better and smarter than writing the stuff on your own.

tkimnguyen:
Maybe if Sven doesn't want this in the official docs, you can add it to the Digital Ocean how-to articles if they have Linode's equivalents at https://www.linode.com/docs

hvelarde:
I think is clear now that this must not be in the documentation; fill free to reopen if I'm wrong.

PR #868, opened 2017-05-08
Document ImportError on case unaware file systems.

mauritsvanrees:
See https://community.plone.org/t/plone-5-0-7-unifiedinstaller-importerror-issue/3986


svx:
Thanks !

PR #869, opened 2017-05-08
Reword the reindexing section and add more information

hvelarde:
Reword the whole section and add a link to https://community.plone.org/t/best-practices-on-reindexing-the-catalog/4157

### DIFF
--- a/develop/plone/searching_and_indexing/indexing.rst
+++ b/develop/plone/searching_and_indexing/indexing.rst
@@ -281,64 +281,53 @@ The difference with metadata is that
 
 * You store always a value copy as is
 
-To create metadata colums in your ``catalog.xml`` add::
+To create metadata colums in your ``catalog.xml`` add:
 
-	<?xml version="1.0"?>
-	<object name="portal_catalog" meta_type="Plone Catalog Tool">
+.. code-block:: xml
 
-		<!-- Add a new metadata column which will read from context.getSignificant() function -->
-		<column value="getSignificant"/>
-
-	</object>
+    <?xml version="1.0"?>
+    <object name="portal_catalog" meta_type="Plone Catalog Tool">
+      <!-- Add a new metadata column which will read from context.getSignificant() function -->
+      <column value="getSignificant"/>
+    </object>
 
 
 When indexing happens and how to reindex manually
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


develop/plone/searching_and_indexing/indexing.rst
```diff
@@ -281,64 +281,53 @@ The difference with metadata is that
...
 
 * The object is modified by the user using the standard edit forms
 
 * portal_catalog rebuild is run (from *Advanced* tab)
 
-* If you add a new
-  index you need to run :doc:`Rebuild catalog </develop/plone/searching_and_indexing/catalog>`
-  to get the existing values from content objects to new index.
-
-* You might also want to call :doc:`reindexObject()
```
hvelarde, line 310: I hate the fact our documentation is full of corner cases that are mostly useless for newbies.

tkimnguyen: They may be useless to newbies but they're good to have in the documentation so people can find them when they search...

tkimnguyen: I'm uncomfortable deleting material; maybe you could move what you consider to be edge case documentation to a catch-all section at the end of this page?

I also think it's preferable to summarize the forum thread information and put that summarized info here, rather than making someone read the entire thread and have to digest it to get the final info. There is also the chance that the forum thread might not be there in future (broken links).

tkimnguyen (commented):
Please see my comments but I defer to @svx and @polyester.

develop/plone/searching_and_indexing/indexing.rst
hvelarde: no, that paragraph is completely useless; before saying you're _uncomfortable deleting material_ you have to review if the documentation in question is useful; what's the use case of that? none, zero, nada...

hvelarde: how can an interface from a private module that is not supposed to be used could be useful? I'm not uncomfortable deleting material that is useless, like in here; our documentation is plenty of corner cases that just confuse and scare the newbie.

this is what I can do now; giving the context of the thread is more useful, IMO, that do otherwise.

jensens:
I think the new version is much better/readable/less confusing than the old. So it could be better/more complete/more perfect, but I would just merge the change to get it better now than perfect never.

hvelarde:
do you want this same PR on previous versions of the documentation?

PR #873, opened 2017-05-17
improve syntax and grammar

stevepiercy:


svx:
Please create smaller pull requests, as said earlier, one per file would be nice !

Thanks !

There some 'issues' here which are strictly sproken not wrong, and yes that is partly our fault, cause we did not merged the latest style-guide and best practices yet, cause of missing time.

But, since I prefer to not touch multiple times if I can avoid it here some pointers, and yes soory again that these are not yet in the style-guide:

- One sentence per line

Headers:

Every word in a header starts with a capital letter 

Example:

Old: Hello world

New: Hello World

Thanks !

svx:
I will merge that for now, still there are various fixes needed here.
If you want to contribute in the future, please consider to crate a branch in github.com/plone/documentation, by doing this other contributors would be able to update/improve Pull Requests which would makes it easier and also faster to merge.

Thanks

Issue #818, opened 2017-04-18
Inline code formatting


svx:
At the moment we have a interesting mix of various ways how we show inline code like how to run a certain command.

Currently we use techniques like:

"run command"

'run command'

``run command``

**run command**

*run $command*

This is somehow confusing for user. I propose that we settle on one and use that one through the whole documentation.

I would be in favor of using ``run command`` consistent through the whole docs, but this means we have to adjust certain bits where we use `` `` for highlighting stuff, which IMHO would be fine.

gforcada:
Yes please!

svx:
Okay, I checked again and before we start with this we need something for 'highlithing' stuff, for example we use now in some places this:

```
This ability is used by ``zc.buildout``
```
looks like: This ability is used by ``zc.buildout``

If we want to use back ticks for inlne code commands we need something else, should we make highlights bold ?

Meaning

```
 This ability is used by **zc.buildout**
```
looks like: This ability is used by **zc.buildout**

Summary
------------

Example1:

After you updated **buildout.cfg** with your changes, you have to re-run buildout again: ``/bin/buildout -v``

Or if we change it the way around (like Kubernetes does it)

Example2:

After you updated ``buildout.cfg`` with your changes, you have to re-run buildout again: **/bin/buildout -v**

Which one is better as in more clear and easier to understand ?

svx:
@polyester @fredvd any suggestions ?

polyester:
I'd say stay as close to the norm as we can. We use Restructured text, so:
- double backticks means "inline literal". Normally rendered using monospace font, and it would be good for code commands. 
- single backticks are "interpreted text". They are domain- or application- specific, so we can define we want this to mean "Highlight" and/or "particular names". 

So I would re-phrase the example as:
```
After you updated `buildout.cfg` with your changes, you have to re-run buiildout again: ``/bin/buildout -v``
```

Now, of course how that "looks" is a question of CSS, but it does make sense in a semantic way.
So: 
- inline code use double backticks
- names of programs and things you want to highlight, could be rendered as bold with a yellow background or anyway else: use single backticks
- things with heavy emphasis, conventionally rendered in bold: use double asterisk

stevepiercy:
I think it's more important to use semantic markup first, and appearance second.

This in reST:

```
``Hello World!``
```

Becomes this in HTML:

```html
<code class="docutils literal"><span class="pre">Hello,</span> <span class="pre">world!</span></code>
```

```
`Hello World!`
[...]

PR #876, opened 2017-05-17
Improve caching caching: fix syntax, grammar, links

stevepiercy:


svx:
Plese see: https://github.com/plone/documentation/issues/818

'd say stay as close to the norm as we can. We use Restructured text, so:
- double backticks means "inline literal". Normally rendered using monospace font, and it would be good for code commands. 
- single backticks are "interpreted text". They are domain- or application- specific, so we can define we want this to mean "Highlight" and/or "particular names". 

So I would re-phrase the example as:
```
After you updated `buildout.cfg` with your changes, you have to re-run buiildout again: ``/bin/buildout -v``
```

Now, of course how that "looks" is a question of CSS, but it does make sense in a semantic way.
So: 
- inline code use double backticks
- names of programs and things you want to highlight, could be rendered as bold with a yellow background or anyway else: use single backticks
- things with heavy emphasis, conventionally rendered in bold: use double asterisk

And again, please smaller pull requests, this will take now longer too review and so on ..

Thanks !

stevepiercy:
> Plese see: #818

I didn't know it had been decided because the issue is still open. If it has been decided, then the style guide should be updated so there is something to use as reference.

I'll update this PR (and others) accordingly.

stevepiercy:
> And again, please smaller pull requests

I broke them into one PR per page as requested. However, there is a problem with breaking these into even smaller PRs. Unless there is some other way I don't know about (suggestions very welcome!), GitHub requires that the developer create a new branch, commit the minor change, push to new branch, submit PR, all in order to have separate, "digestible" PRs. This is not insignificant overhead, not just from the mechanics, but more from the constant interruption of concentrating on writing docs. All this would impose a significant barrier to developers who want to make overhaul-scale updates. I would be strongly discouraged from making any contributions other than extremely minor changes. Other developers might feel the same way.

I don't know, maybe it's just my way of working that doesn't fit in with the Plone community. When I update docs, I read the whole thing from top to bottom, and make fixes as I go along. Is that weird?

svx:
Here is how we usually work:

For *huge changes* like overhaul whole chapter we create branches in github.com/plone/documentation this enables other contributors to help/improve/fix.

Most of the **awesome** people who work on the docs, doing that even for every PR !

For *smaller* stuff like for example fixing grammar, typos and so on, we prefer one PR per file, better lots of small PR which we can review/merge faster than one big one.

This is proven to work for us very well !

svx:
I will merge for now, but for the next time, please ask before you start so that we can discus/explain first !

Thanks !

PR #877, opened 2017-05-17
Improve frontend apache: fix reST syntax, grammar, links

stevepiercy:


svx:
I will merge that for now, but please would you mind to:

- avoid words like: *very*, *just* or *easy* -> you used *very*
- next time please one PR per file, these here are three changed files.

Yes, this may be differ from the way you work, sorry for that.
But we choose to work like that and for us that is working well, and usually PR are faster reviewed, improved and merged.

Thanks !

PR #882, opened 2017-05-19
correct import to use asdatetime

tkimnguyen:
Fixes the import statement needed to run the code snippet

### DIFF
--- a/develop/plone/misc/datetime.rst
+++ b/develop/plone/misc/datetime.rst
@@ -74,7 +74,7 @@ Since two different datetime object types are used, you need to often convert be
 
 You can convert Zope DateTime objects to datetime objects like so::
 
-        from DateTime import DateTime
+        from DateTime import DateTime, asdatetime
         zope_DT = DateTime() # this is now.
         python_dt = zope_DT.asdatetime()
 


svx:
Thanks !

ale-rt:
This has to be reverted because DataTime module has no asdatetime:
```
>>> from DateTime import asdatetime
Traceback (most recent call last):
  File "<console>", line 1, in <module>
ImportError: cannot import name asdatetime
```

ale-rt:
Sorry @svx, I should have asked @tkimnguyen to close this before :p

svx:
done by #883 Thanks !

tkimnguyen:
Argh, sorry about that. I had run into a problem using asdatetime in some of my own code. I must have gotten confused as to what I ended up doing to get it to work. It might have had something to do with how I tried the sample code... using a normal Python interpreter I get errors, but if I invoke the sample code as a script run like `bin/instance run test.py` then it works ok. Maybe that is something worth clarifying on this page?

```
bash-3.2$ bin/python test.py
Traceback (most recent call last):
  File "test.py", line 6, in <module>
    from DateTime import DateTime
ImportError: No module named DateTime
bash-3.2$

**Comment by ale-rt** [comment]
I think this has something to to with the DateTime module not being in your path.

Probably running:
```
bash-3.2$ bin/zopepy test.py
```
will give you different results.

Anyway I think it will fail neverthless, because `bin/instance run` is going to set up the zope app global and some other stuff.

PR #884, opened 2017-05-24
Fix linx to point to the latest 5.1 versions file

ale-rt:
Fixes a link to a non existing file


svx:
Yeah that was done on purpose :)

Anyway thanks ! :)

svx:
I should really fix that automatic version include in Sphinx ...

PR #904, opened 2017-07-13
Toolbar developer documentation

agitator:
Improves:
Documentation on how to set permissions to show/hide the toolbar and how to extend it.


svx:
@agitator wait let me fix that !

svx:
@agitator I included the docs from hand, this was faster than fixing this PR, also I made some small reST improvements, still not perfect but OK for now I guess :)

PR #905, opened 2017-07-13
update description

agitator:
Improves:
existing toolbar docs

### DIFF
--- a/develop/plone/functionality/toolbar.rst
+++ b/develop/plone/functionality/toolbar.rst
@@ -4,7 +4,7 @@ Toolbar
 
 .. topic:: Description
 
-   How to deal with permissions making your code permission-aware in Plone.
+   How to configure and extend the toolbar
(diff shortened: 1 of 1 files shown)


agitator:
@svx :doc: and :guilabel: look weird, how should that be formatted?

svx (commented):
Should there be a . at the end ?

svx:
@agitator what do you have against :doc: or :guilabel: ?

I mean it depends on the content :guilabel: is used for buttons, meaning that you do not write click on $foo but :guilabel:`$foo`.

But yeah I have to read the toolbat docs again, I am a bit in a hurry, I will check tomorrow :)

:doc: is used to directly link to a document -> http://www.sphinx-doc.org/en/stable/markup/inline.html

agitator:
ping me on #plone tomorrow then ;-)

Issue #822, opened 2017-04-20
Improve docs about 'First Time Contributions' 


svx:
Improve docs to make clear that people who contribute for the first time(s):

- pick a ticket which is labeled 

- if want to write docs about something else, first open a ticket on GitHub, labeled as question and explain what and why.

If you are 'newish' it is not always clear where, why and how we do certain things in a certain way.
To avoid that people invest time and motivation in 'wrong' work we should make this more clear.

stevepiercy:
I think this is already well covered in the repo's [README.rst](https://github.com/plone/documentation/) and [Contributing to the documentation](https://docs.plone.org/about/contributing.html).

Additionally in this case, a discussion originated in the [Plone Community forum](https://community.plone.org/t/documenting-plone-installation-for-digital-ocean-etc/4009). The author was informed of the process.

I can think of only a couple of minor improvements.

1. There exists in the footer of the Plone documentation site the following. The link should remove the `#id6` because it does not exist on the target page.

> About this documentation:
> [Contribute](http://docs.plone.org/about/contributing.html#id6)

2. When you [create a new Issue](https://github.com/plone/documentation/issues/new), you get the friendly yellow bar with a link to [guidelines](https://github.com/plone/documentation/blob/5.0/.github/CONTRIBUTING). However, when one clicks the link, then the page is not rendered as HTML because it lacks an extension of `.md` (preferred on GitHub) or `.rst`.

stevepiercy:
3. Document how to build docs in just the `documentation` repository, instead of via the `papyrus` repository.

svx:
papyrus is not meant for 'normal' people, we have already: https://docs.plone.org/about/helper_tools.html, which will updated soon with more/other solutions

stevepiercy:
For 3:

1. Follow the Papyrus installation instructions, making sure that you can build docs there and the virtual environment is activated.
2. Clone the `documentation` repo.
3. Copy both `Makefile` and `conf.py` from `papyrus` to `documentation`.
4. Edit `Makefile`: `ALLSPHINXOPTS   = -d build/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .`
5. Create a new directory `log`.
6. `make html SPHINXBUILD=/Users/stevepiercy/projects/Plone/papyrus/bin/sphinx-build` (or wherever the path to papyrus' sphinx-build is located)

svx:
[...]

Issue #802, opened 2017-03-15
Improve: Docs about how to contribute


svx:
We should tell:

- send better a couple of small PRs than one huge one
- give nice examples, like a gist with a git-hook or so, and explain people to run local tests, this should be not to heavy, like not use my own setup as example, since We do not want to scare people.

tomgross:
And: Give instructions on how to build the documentation locally.

loechel:
@tomgross I have made a pull request that makes it possible and easy to build the docs locally --> #920 

Might one of you have a look and review?

stevepiercy:
Also requested in https://github.com/plone/documentation/issues/822#issuecomment-295663293

stevepiercy:
See Contributing in Plone 6 docs (temporary URL): https://6.dev-docs.plone.org/contributing/index.html

PR #920, opened 2017-09-26
Make Docs self-buildable 

loechel:
for users it is hard to contribute, if they could not generate the documentation locally, papyrus might be a good idea for deployment and continuous Integration, but not for a local build.

This was also requested in https://github.com/plone/documentation/issues/802#issuecomment-332295192 & https://github.com/plone/documentation/issues/822#issuecomment-295663293.


stevepiercy:
I requested this in https://github.com/plone/documentation/issues/822#issuecomment-295663293, too, and the issue is open. Thanks for moving this forward.

svx (commented):
LGTM

stevepiercy:
This is awesome! I tried it out and it worked. Thank you @loechel!

PR #959, opened 2018-02-15
explain how to find available image scales

tkimnguyen:
Improves: explanation of image scales

- explain how to find available image scales

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changelog
 
 Unreleased
 ==========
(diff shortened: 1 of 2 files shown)


svx:
Thanks a lot !

gforcada:
IMHO, there is a image handling control panel, so, there is no need to go to the registry control panel.

tkimnguyen:
Oh geez - thx @gforcada I had totally missed that! I will adjust.

PR #966, opened 2018-02-22
update Windows installer info

tkimnguyen:
Improves Windows story to include our new Windows installer


jensens:
Remark: In Windows 10 with Linux Subsystem (LXSS) the development is as easy as on Linux or macOS. We just did not manage to get robot tests running. But that is out of scope here.

tkimnguyen:
True but the Windows installer automates a lot of the things that someone would otherwise have to read and do by hand. Thx for merging!

PR #967, opened 2018-02-23
add restaurant metaphor to explain ZEO

tkimnguyen:
Improves: explanation of how ZEO works

as per https://community.plone.org/t/what-does-zeo-stand-for/5888/3


svx:
@tkimnguyen Thanks !

A question do we think that this is the right place to put these in general ?

IMHO this is one part of the docs which should completely re-structured :)

It should go somewhere more general, why we should place it under  /manage/deploying ?

This is not only for deploying, right ?

tkimnguyen:
It seems to me it belongs here because this is where the explanation of ZEO lives.

PR #977, opened 2018-03-19
Improve files download url section

jensens:
Improves: 

- information how to construct a download URL for any dx file field.

- Explain what happens instead of only showing an opinionated example.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,7 @@ Unreleased
 - Improve ReStructuredText Style Guide [svx]
 - Add documentation about TinyMCE styles and formats [davilima6]
 - Improve Editor Guide [tmassman]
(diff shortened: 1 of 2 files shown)


svx (approved):
As a note, personally I would not use ``in order`` we removed or changed that already on lots of locations.

But for now this is not the the style-guide :)

LGTM ! Thanks !

svx:
Oh, yes circle ci is not working anymore, ignore if for now ! I will remove the test.
That is because of circle ci

jensens:
Circle worked and found Javascript, I pushed it.

Well the chapter needs a complete overhaul anyway. It uses `plone.directives.form`, `implements` and other outdated. I'll open an issue, so it does not get lost.

svx:
Yeah but we are using Circle 1 which will be retired in a couple of weeks anyway.
I will setup travis upcoming weekend !

PR #980, opened 2018-03-19
remove duplicate placement of the javascript/ajax section

sunew:
selection_558

remove duplicate placement of the javascript/ajax section, it shows up twice in the contents, but is really the same document


svx:
Thanks !

PR #995, opened 2018-04-25
Update mrbob commands

abosio:
Improves: Updates mrbob commands to be correct for current version

- Old: mrbob -O collective.myaddon bobtemplates:plone_addon

- New: mrbob bobtemplates.plone:addon -O src/collective.myaddon

- Old command produces: `AttributeError: 'module' object has no attribute '__file__'`

### DIFF
--- a/develop/addons/bobtemplates.plone/README.rst
+++ b/develop/addons/bobtemplates.plone/README.rst
@@ -28,11 +28,11 @@ bobtemplates introduction
 
 To create a package like ``collective.myaddon``::
 
-    $ mrbob -O collective.myaddon bobtemplates:plone_addon
+    $ mrbob bobtemplates.plone:addon -O src/collective.myaddon
 
 You can also create a package with nested namespace::
 
-    $ mrbob -O collective.foo.myaddon bobtemplates:plone_addon
+    $ mrbob bobtemplates.plone:addon -O src/collective.foo.myaddon
 
 
 Options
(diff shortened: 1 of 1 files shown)


polyester (approved):

svx:
@polyester we fetch the docs straight from bobtemplates.plone.

The change needs to be done there :)

svx (changes requested):
This change needs to be done in the repo of bobtemplates.plone :)

abosio:
@svx Do you know exactly where that is coming from because it already appears correctly here: https://github.com/plone/bobtemplates.plone/blob/master/README.rst

svx:
@abosio yes I do :)

The README which is still in the docs is from the first times back in the days when we not *auto-included* the docs of bobtemplates.plone.

It got never removed, guess we forgot that.

As soon as we release a new build of the docs the file on docs.plone.org will be updated with the new version out of the bobtemplates.plone repo.

abosio:
Great! Does that happen fairly regularly?

svx:
There we not lots of commits to the docs during the last couple of months and because of this there was no new-build.

I guess we could/should push a new one soon :)

abosio:
That would be good, for this if nothing else.

I was looking at both sets of documentation but did not realize the command was slightly different and I ended up making a unhelpful issue on that repo then making this PR so that someone else, maybe someone new to Plone, would not have the same problem creating an add on.

Thanks for the feedback!

PR #1014, opened 2018-08-03
fix broken link

jugmac00:
The link to Sam Rushing's Medusa page is no longer directly available, so I replaced it with a link from archive.org.

Fixes:  a broken link

Improves: usability

- replace broken link with a working one from archive.org


svx:
thanks !

PR #1026, opened 2018-10-21
Revise styleguide tone and word list

stevepiercy:
Per discussion:
https://github.com/plone/training/pull/336#discussion_r226870182

- Avoid contractions, spell them out.
- Break sentences on to own lines.
- Make use of quote marks consistent.
- Add "js" to word list to avoid.
- Add derivatives to words to avoid (easily, simply).


svx:
Thanks !

About the line length, what is if we say aim to keep your lines to around 130 chars, I use something like that in other guides, because it gives you an idea, but it is no "hard" setting, 140 chars is 150% fine, if it fit the sentence :)

svx:
Thanks !

About the line length, what is if we say aim to keep your lines to around 130 chars, I use something like that in other guides, because it gives you an idea, but it is no "hard" setting, 140 chars is 150% fine, if it fit the sentence :)

svx:
Thanks !

About the line length, what is if we say aim to keep your lines to around 130 chars, I use something like that in other guides, because it gives you an idea, but it is no "hard" setting, 140 chars is 150% fine, if it fit the sentence :)

PR #1029, opened 2018-10-24
use return instead of rewrite for (non-)www redirection

petschki:
according to https://www.nginx.com/resources/wiki/start/topics/tutorials/config_pitfalls/#taxing-rewrites this is much faster


svx:
LGTM ! @cleberjsantos ? :)

jensens:
Lgtm.

PR #1040, opened 2018-12-06
It might be a good idea to activate the virtualenv

bsnizek:
Thank you for your contribution to the Plone Documentation.

Before submitting this PR, please make sure:

- [ ] You are following our [Documentation Styleguide](https://docs.plone.org/about/contributing/documentation_styleguide.html)
- [ ] You are following our [General Writing Guidelines](https://docs.plone.org/about/contributing/rst-styleguide.html)

Fixes: #

Improves:

-

-

-

### DIFF
--- a/manage/installing/installation_minimal_buildout.rst
+++ b/manage/installing/installation_minimal_buildout.rst
@@ -22,6 +22,7 @@ Create a virtual python environment (`virtualenv <https://virtualenv.pypa.io/en/
 
    virtualenv-2.7 zinstance
    cd zinstance
+   source bin/activate
    bin/pip install zc.buildout
 
 Create a buildout.cfg file:


manage/installing/installation_minimal_buildout.rst
```diff
@@ -22,6 +22,7 @@ Create a virtual python environment (`virtualenv <https://virtualenv.pypa.io/en/
 
    virtualenv-2.7 zinstance
    cd zinstance
+   source bin/activate
```
jensens, line 25: since it uses explicit the ``bin/pip`` afterwards it does not need any activation. So, it does not hurt.

jensens:
Well, nobody merged it, but anyway we're now further forward with the docs, I think this can be closed.

PR #1046, opened 2019-02-02
Use localhost instead of yourserver

jaroel:
This changes the installation docs so people will actually see the new Plone install


stevepiercy (approved):

stevepiercy:
@jaroel would you also please submit a PR with this change to the `5.2` branch? That means this change would be carried forward to future versions of Plone. Thank you!

PR #1047, opened 2019-02-02
Use localhost instead of yourserver

jaroel:
Applies #1046 to 5.2


stevepiercy (approved):

stevepiercy:
Thank you for the follow through!

PR #1054, opened 2019-03-02
Revis Line Length section

### LINKED ISSUE #1027: Line Length clarification
There are some conflicts within the [Line Length section](https://docs.plone.org/about/contributing/documentation_styleguide.html#line-length).

> - Use semantic linefeeds when you are editing.
> - (snip)
> - One sentence per line.

These conflict with one another. For the non-professional documentation writer, one sentence per line is a lot easier rule to understand than semantic line feeds. I would much prefer one sentence per line.

Personally I can't _stand_ semantic line feeds because it makes it much harder to read when editing. It's like reading bad adolescent shoe-gazer emo poetry.

> Please do not follow PEP8 maximum line length standard.

Appears twice in this section. I'd suggest deleting the second instance.

> Limiting lines to a maximum of 130 characters.

This conflicts with "one sentence per line". Suggest its removal.

stevepiercy:
Fixes #1027

### DIFF
--- a/about/contributing/documentation_styleguide.rst
+++ b/about/contributing/documentation_styleguide.rst
@@ -45,31 +45,23 @@ Line Length
 -----------
 
 - Please do not follow `PEP8 <https://www.python.org/dev/peps/pep-0008/#maximum-line-length>`_ maximum line length standard.
-
-- Keep sentences short and understandable.
-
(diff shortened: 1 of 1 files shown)


svx (approved):
Awesome! Thanks!

PR #1063, opened 2019-04-01
Styleguide cleanup

stevepiercy:
Clean up the styleguide per the styleguide.

GitHub does not display the diff nicely when deleting a stanza, so it looks like I made more changes than I actually did.


svx:
Thanks!

PR #1064, opened 2019-04-02
capitalize title to match other sections

tkimnguyen:
Other sections in this part of the docs https://docs.plone.org/working-with-content/adding-content/index.html capitalize all words in the heading


stevepiercy:
Thank you!

PR #1086, opened 2019-07-24
Custom object actions per portal type

idgserpro:
Thank you for your contribution to the Plone Documentation.

Before submitting this PR, please make sure:

- [X] You are following our [Documentation Styleguide](https://docs.plone.org/about/contributing/documentation_styleguide.html)
- [X] You are following our [General Writing Guidelines](https://docs.plone.org/about/contributing/rst-styleguide.html)

Improves:

- Documentation about custom object actions per portal type that still uses CMFFormController.

### DIFF
--- a/develop/plone/functionality/actions.rst
+++ b/develop/plone/functionality/actions.rst
@@ -365,6 +365,13 @@ You might want to have different actions for different site sections or folders.
 
 * http://plone.293351.n2.nabble.com/Custom-portal-tabs-per-subsection-tp5747768p5747768.html
 
+Custom object and object_buttons actions per portal type
+--------------------------------------------------------
+
+If you need to override or customize an action from ``object_buttons`` that still uses ``CMFFormController`` for a specific ``portal_type``, check:
+
+* https://stackoverflow.com/questions/11218272/plone-reacting-to-object-removal/11225447#11225447
+
 Copy, cut and paste
 ----------------------
 


develop/plone/functionality/actions.rst
```diff
@@ -365,6 +365,13 @@ You might want to have different actions for different site sections or folders.
 
 * http://plone.293351.n2.nabble.com/Custom-portal-tabs-per-subsection-tp5747768p5747768.html
 
+Custom object and object_buttons actions per portal type
+--------------------------------------------------------
+
+If you need to override/customize an action from `object_buttons` that still uses CMFFormController for a specific `portal_type`, check:
```
stevepiercy: All inline literals must be wrapped in double backticks in reST. Markdown uses single.

Also avoid "/" in narrative text, using the appropriate conjunction.

```
If you need to override or customize an action from ``object_buttons`` that still uses ``CMFFormController`` for a specific ``portal_type``, check:
```

Also for consistency throughout this file, it would be nice to apply double backticks to all other instances of these three inline literals while you are in there, but not required.

Thank you!

stevepiercy (changes requested):

develop/plone/functionality/actions.rst
idgserpro: Thanks for the review. Indeed I was messing with markdown files before editing this rst.

> Also for consistency throughout this file, it would be nice to apply double backticks to all other instances of these three inline literals while you are in there, but not required.

I think it's better to do this in a new PR.

stevepiercy (approved):

stevepiercy:
Thank you!

Also please make the same PR against the 5.2 branch, so that it gets included going forward, else it might get lost.

PR #1099, opened 2019-10-14
Fix implicit link to subsection.

jugmac00:
In reStructuredText it is easily possible to create a link to a
subsection by adding backticks around the same words as the title of
the subsection.

Here, the question mark at the end was missing, resulting in a broken
link.

This has been fixed by this commit.

modified:   manage/upgrading/version_specific_migration/upgrade_zodb_to_python3.rst

### DIFF
--- a/manage/upgrading/version_specific_migration/upgrade_zodb_to_python3.rst
+++ b/manage/upgrading/version_specific_migration/upgrade_zodb_to_python3.rst
@@ -46,8 +46,8 @@ In short you need to follow these steps to migrate your database:
 
 
 
-Why Do I Have To Migrate My Database
-====================================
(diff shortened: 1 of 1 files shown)


stevepiercy (approved):

stevepiercy:
In case that doesn't work, try using an explicit label and reference as in this example: https://docs.pylonsproject.org/projects/docs-style-guide/#cross-referencing-arbitrary-locations

jugmac00:
@stevepiercy Thank you for the review.

Thanks to @polyester I was able to build the documentation locally and can confirm the link now works.

ataoleo:
f


---Original---

PR #1101, opened 2019-11-04
Add info about ContentProviderLookupError

idgserpro:
Although already documented, it's nice to warn here as well about acquisition chain.

Thank you for your contribution to the Plone Documentation.

Before submitting this PR, please make sure:

- [X] You are following our [Documentation Styleguide](https://docs.plone.org/about/contributing/documentation_styleguide.html)
- [X] You are following our [General Writing Guidelines](https://docs.plone.org/about/contributing/rst-styleguide.html)

Fixes: #
N/A
Improves:
Gives hints about errors that may occur when creating viewletmanagers.
Changes proposed in this pull request:

- Add info about `ContentProviderLookupError`

PR #1102, opened 2019-11-05
Fix minor rst formatting

idgserpro:
Inspired by https://github.com/plone/documentation/pull/1101#discussion_r342705326

If anyone is interested in redoing this in another branches, run:

`grep -rl '.. note ::' . | xargs sed -i 's/.. note ::/.. note::/g'`


stevepiercy (approved):
Thanks for catching and fixing all these!

PR #1105, opened 2019-11-21
Prevent content-spoofing in 404 pages

svx:
This PR updates `AllowEncodedSlashes NoDecode` to VirtualHost configs to prevent content-spoofing in Apache's 404 error pages.

More info on the topic:

- [Apache docs](https://httpd.apache.org/docs/2.4/mod/core.html#allowencodedslashes)
- [Blog post](https://ma.ttias.be/prevent-content-spoofing-apache-404-error-pages/)

### DIFF
--- a/manage/deploying/front-end/apache.rst
+++ b/manage/deploying/front-end/apache.rst
@@ -43,6 +43,7 @@ domain name is ``yoursite.com`` (note with or without ``www`` matters, see below
     <VirtualHost *>
         ServerAlias yoursite.com
         ServerSignature On
+        AllowEncodedSlashes NoDecode
 
         Header set X-Frame-Options "SAMEORIGIN"
         Header set Strict-Transport-Security "max-age=15768000; includeSubDomains"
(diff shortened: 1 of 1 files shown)


jensens (commented):
I am not an Apache expert. But reading the docs confuses me, there are two valid options for ``AllowEncodedSlashes``: ``on`` and ``off``. There is no ``NoDecode`` mentioned in the docs. See
https://httpd.apache.org/docs/2.4/de/mod/core.html#allowencodedslashes

@svx May you explain?

svx:
@jensens it is in there :smile: 

https://httpd.apache.org/docs/2.4/mod/core.html#allowencodedslashes

says:

Syntax: `AllowEncodedSlashes On|Off|NoDecode`

Same paragraph (4 lines below the box)

With the value *NoDecode*, such URLs are accepted, but encoded slashes are not decoded but left in their encoded state.

Does this help, or should I provide more info?

PR #1106, opened 2019-11-21
Fix link to six documentation.

jugmac00:
Fixes: broken link


stevepiercy (approved):

stevepiercy:
Thank you! Would you also please make the same PR against `6` branch?

jugmac00:
> Thank you! Would you also please make the same PR against `6` branch?

Unfortunately I cannot find the "manage" section in the 6 branch.

My best guess is a "upgrade to Python3" guide makes no more sense for Plone 6, as Plone 6 probably won't support Python 2 any more?

stevepiercy:
Ah, yes. That makes sense. Thanks for checking. I made a mistake in assuming that Plone 6 docs will be anything like the current docs.

PR #1110, opened 2019-12-06
Fix #1109 Update import statement

### LINKED ISSUE #1109: wrong example execute_under_special_role
the example `execute_under_special_role` has a [wrong import statement](https://docs.plone.org/develop/plone/security/permissions.html#bypassing-permission-checksl)  for Plone 5.2

old:
```
from AccessControl.User import UnrestrictedUser as BaseUnrestrictedUser
```
new:
```
from AccessControl.User import Super as BaseUnrestrictedUser
```

1letter:
Update import statement in example 'Bypassing permission checks'

Fixes: #1109

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,10 +2,11 @@
 Changelog
 =========
 
-Unreleased 
+Unreleased
 ==========
 
 - New documentation how to develop on Windows 10 with WSL [nilshofer]
+- Fix #1109 Update import statement in example 'Bypassing permission checks' [1letter]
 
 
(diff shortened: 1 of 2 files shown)


spereverde:
Thanks for the fix!
I saw that backwards compatibility for the import was added in https://github.com/zopefoundation/AccessControl/blob/master/CHANGES.rst#42-2020-04-20, but the documentation should reflect the new way
We might need to add a note for users using plone < 5.2, since there is no way to differentiate in the live docs between the minor versions

PR #1114, opened 2020-01-30
Fixed 2 broken links on wsgi page

mikejmets:
Fixed broken "Zope 4 and WSGI" and "Plone 5 with WSGI" links.

### DIFF
--- a/manage/deploying/wsgi.rst
+++ b/manage/deploying/wsgi.rst
@@ -17,8 +17,8 @@ Plone and WSGI
 
 Collection of Links about this topic, since this is a very special topic we also link to documentation about WSGI and Plone 4 for reference.
 
-* `Zope 4 and WSGI <https://zope.readthedocs.io/en/latest/wsgi.html>`_
+* `Zope 4 and WSGI <https://zope.readthedocs.io/en/latest/operation.html#using-alternative-wsgi-server-software>`_
 
-* `Plone 4 with WSGI <http://comments.gmane.org/gmane.comp.web.zope.plone.devel/23886>`_
+* `Plone 4 with WSGI <https://comments.gmane.org/gmane.comp.web.zope.plone.devel/23886>`_
 
-* `Plone 5 with WSGI <http://blog.toms-projekte.de/run-plone-with-wsgi.html>`_
+* `Plone 5 with WSGI <https://blog.toms-projekte.de/run-plone-with-wsgi.html>`_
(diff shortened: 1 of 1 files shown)


manage/deploying/wsgi.rst
```diff
@@ -17,8 +17,8 @@ Plone and WSGI
 
 Collection of Links about this topic, since this is a very special topic we also link to documentation about WSGI and Plone 4 for reference.
 
-* `Zope 4 and WSGI <https://zope.readthedocs.io/en/latest/wsgi.html>`_
+* `Zope 4 and WSGI <https://zope.readthedocs.io/en/latest/operation.html#using-alternative-wsgi-server-software>`_
 
-* `Plone 4 with WSGI <http://comments.gmane.org/gmane.comp.web.zope.plone.devel/23886>`_
+* `Plone 4 with WSGI <https://comments.gmane.org/gmane.comp.web.zope.plone.devel/23886>`_
```
stevepiercy: This link never connects. It appears that gmane.org has been out of commission for about a year. So I dug around in the Wayback Machine to find this link.

https://web.archive.org/web/20100702205219/http://permalink.gmane.org:80/gmane.comp.web.zope.plone.devel/23897

Is that relevant? It's before my time.

stevepiercy (commented):
Two of the three changes are good. I don't know about the link for "Plone 4 with WSGI". Someone with more historical knowledge will have to comment. Because Plone 4.3.x is still supported, we should keep this link around, unless we cannot find a suitable replacement.

mikejmets:
If no one comes back with anything we could replace the link with an extract from that conversation. Something like "WSGI is not officially supported by either Zope2 or Plone 4 . There is code in ZPublisher that should work but it has minimal documentation and should be viewed as experimental"

jensens:
Well, this is the 5.2 branch of the docs. I would remove the whole page. There is a 4.3 branch targeting the old release.

stevepiercy:
I think purging the history of Plone 4.x is correct.

However I think we need to keep this page and update the WSGI information. [Plone 5.2 uses Waitress instead of ZServer](https://docs.plone.org/manage/upgrading/version_specific_migration/upgrade_to_52.html#id5).

@jensens does that mean we also need to update the page [Zope Application Server](https://docs.plone.org/manage/deploying/zope.html)? Sorry, I don't know the architecture well.

spereverde:
I know this is a bit old by now, but if all agreed I will dot he following:

4.3 branch:
- update the page with a short note about WSGI in Plone 4 (summary of the gmane info, something like mike suggested)

5.2 branch (added to this PR)
- remove the Plone4 and WSGI link
- add a link to [Plone 5.2 uses Waitress instead of ZServer.](https://docs.plone.org/manage/upgrading/version_specific_migration/upgrade_to_52.html#id5)

spereverde:
adjusted docs on Plone 4.3 branch: https://github.com/plone/documentation/blob/4.3/manage/deploying/wsgi.rst (https://github.com/plone/documentation/commit/55f9ae63759576342684da7f3e0b00fd79e4bbda)

PR #1119, opened 2020-06-25
Alternative to enable captcha after installing pfg

idgserpro:
Ref: https://github.com/smcmahon/Products.PloneFormGen/pull/226/commits/a7563e462d1863cbdcdb82eacc222ae6ebcb51e8

Give an alternative to add captcha support without reinstalling the package. Ref: https://community.plone.org/t/when-was-using-reinstall-in-portal-quickinstaller-considered-harmful/4789

### DIFF
--- a/working-with-content/managing-content/ploneformgen/captcha.rst
+++ b/working-with-content/managing-content/ploneformgen/captcha.rst
@@ -13,7 +13,7 @@ When PFG is installed in a Plone instance via add/remove products, it will look 
 
 If you are using collective.recaptcha, you need to take the additional step of setting your public/private keypair. You get these by setting up an account at recaptcha.net. The account is free. You may specify your keypair in the PFG configlet in your site settings.
 
-If you add a captcha facility *after* installing PFG, to enable captcha support, you will need to add `FormCaptchaField` as an allowed content type to `FormFolder` in `portal_types` or reinstall PFG (via add/remove products).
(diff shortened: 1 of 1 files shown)


working-with-content/managing-content/ploneformgen/captcha.rst
```diff
@@ -13,6 +13,8 @@ When PFG is installed in a Plone instance via add/remove products, it will look
 
 If you are using collective.recaptcha, you need to take the additional step of setting your public/private keypair. You get these by setting up an account at recaptcha.net. The account is free. You may specify your keypair in the PFG configlet in your site settings.
 
+If you add a captcha facility *after* installing PFG, to enable captcha support, you will need to add `FormCaptchaField` as an allowed content type to `FormFolder` in `portal_types` or reinstall PFG (via add/remove products).
```
stevepiercy:
```suggestion
If you add a CAPTCHA facility *after* installing PFG, to enable CAPTCHA support you will need to add ``FormCaptchaField`` as an allowed content type to ``FormFolder`` in ``portal_types`` or reinstall PFG via :menuselection:`Site Setup > Add-ons`.
```

stevepiercy (changes requested):
Minor changes for casing, punctuation, syntax for literal inlines, and use of menuselection.

working-with-content/managing-content/ploneformgen/captcha.rst
stevepiercy: I would also suggest similar changes to the original from which this text was copied.

idgserpro: @stevepiercy The reference was already merged since it's a txt. I commited your suggestions.

stevepiercy (approved):

PR #1197, opened 2022-03-10
Try a new build

sneridagh:


sneridagh:
@stevepiercy I realised that builds are broken since March, 4th.

image

sneridagh:
That commit has no special reason to make things fail...

stevepiercy:
@sneridagh thank you for looking into it. I searched for `broken` in the failed build's log, and I see two links that are broken which will be fixed in an open PR https://github.com/plone/documentation/pull/1195. That PR is waiting for this issue to be resolved: https://github.com/plone/volto/issues/3128

For now it's OK to ignore them, but it does break the build.

sneridagh:
I updated to latest master in Volto, still fails in a valid URL https://razzlejs.org ... 
BTW, not sure if we want to run the link checker in the build test... Not a good CI practice to let your tests rely on third party URLs. Maybe we could separate the check from the test, and issue a warning.

stevepiercy:
> I updated to latest master in Volto, still fails in a valid URL https://razzlejs.org 

I think I need to sync up the `conf.py` settings between the two repos.

> BTW, not sure if we want to run the link checker in the build test... Not a good CI practice to let your tests rely on third party URLs. Maybe we could separate the check from the test, and issue a warning.

Previous experience with the Documentation and Training repos when we did not require `linkcheck` resulted in link rot. No one cared enough to do anything about it. Then you have someone (me) come by every 5 years to run `linkcheck` locally and fix hundreds of links.

We have a solution. If a link is broken and cannot be fixed, then we need to add it to the [`linkcheck_ignore` list](https://github.com/plone/documentation/blob/6-dev/docs/conf.py#L69-L76) or fiddle around with settings. That's what I did in this PR: https://github.com/plone/documentation/pull/1195/files

sneridagh:
@stevepiercy we have green now

stevepiercy (approved):

PR #1206, opened 2022-03-19
Replace deprecated `egrep` with `grep` in `make docs-linkcheckbroken`

stevepiercy:


ksuess (commented):
Would it make sense to filter also by redirect, like in training?
`grep -wi "broken\|redirect"`

stevepiercy:
> Would it make sense to filter also by redirect, like in training? `grep -wi "broken\|redirect"`

Indeed. Good catch. Updated in [6ae99da](https://github.com/plone/documentation/pull/1206/commits/6ae99dafada60d3bd662da631447013bc890aa2d)

ksuess (approved):
LGTM

Issue #1211, opened 2022-03-28
GitHub link for "suggest edit" returns 404 for plone.restapi and volto


stevepiercy:
Visit https://6.dev-docs.plone.org/volto/index.html, click the GitHub Octocat icon, and click "suggest edit". You get a 404 at https://github.com/plone/documentation/edit/main/docs/volto/index.md

The same thing happens with plone.restapi.

The quick fix is to remove the link, but that's not really a good fix.

I don't know if there is a way to override the theme template depending on the path of the file being written.

ksuess:
GitHub link for "suggest edit" overridden for this repo: Impede drive-by edits by linking chapter on how to contribute. Fixed in https://github.com/plone/documentation/pull/1292

PR #1221, opened 2022-04-06
Fix Github button 'suggest edit'  (at least for non-submodule docs)

ksuess:


stevepiercy (approved):

ksuess:
Fixing the "suggest edit" link for submodule docs would mean a massive interference in sphinx-book-theme https://github.com/executablebooks/sphinx-book-theme/blob/526956165f3fda3e5faaf53ffc39081715702f6d/src/sphinx_book_theme/header_buttons/__init__.py#L94-L102

I tend to just hide the link for submodule pages (Volto, plone.restapi, plone.api) via CSS.

stevepiercy:
What do you think of using JavaScript to rewrite the URL for non-main documentation to https://6.dev-docs.plone.org/contributing/index.html ?

Actually, I think I would prefer to use that link site-wide than "suggest an edit" for each page. I don't want to raise expectations of people making "drive by" pull requests without signing the Plone Contributor Agreement. This would direct them toward the authoritative method to submit a PR.

ksuess:
> What do you think of using JavaScript to rewrite the URL for non-main documentation to https://6.dev-docs.plone.org/contributing/index.html ?
> 
> Actually, I think I would prefer to use that link site-wide than "suggest an edit" for each page. I don't want to raise expectations of people making "drive by" pull requests without signing the Plone Contributor Agreement. This would direct them toward the authoritative method to submit a PR.

Yes, I completly aggree. 👍

docs/conf.py
```diff
@@ -214,7 +214,8 @@
     # "google_analytics_id": "UA-1907133-6",
     "path_to_docs": "docs",
     "repository_url": "https://github.com/plone/documentation",
-    "repository_branch": "main",
+    # TODO: Going live: Switch "repository_branch" to 'main' (Github buttons in header)
+    "repository_branch": "6-dev",
     "use_repository_button": True,
     "use_issues_button": True,
     "use_edit_page_button": True,
```
stevepiercy, line 221:
```suggestion
    "use_edit_page_button": False,
```

Then in a template, we can hard code a link to https://6.dev-docs.plone.org/contributing/index.html

stevepiercy (changes requested):

stevepiercy:
See https://github.com/plone/documentation/issues/1211

MrTango:
> What do you think of using JavaScript to rewrite the URL for non-main documentation to https://6.dev-docs.plone.org/contributing/index.html ?
> 
> Actually, I think I would prefer to use that link site-wide than "suggest an edit" for each page. I don't want to raise expectations of people making "drive by" pull requests without signing the Plone Contributor Agreement. This would direct them toward the authoritative method to submit a PR.

For the docs it is not required to sign the controbutor aggrement as far as i know.
With that in mind, i would prefer a direct link to update docs, not just to a document.

stevepiercy:
> For the docs it is not required to sign the controbutor aggrement as far as i know.

See https://6.dev-docs.plone.org/contributing/index.html#granting-permission-to-publish as established by https://docs.plone.org/about/license.html.

> With that in mind, i would prefer a direct link to update docs, not just to a document.

I would, too. However with the complexity of importing submodules from three other repos, this is not easy.

ksuess:
GitHub link for "suggest edit" overridden for this repo: Impede drive-by edits by linking chapter on how to contribute. Fixed in https://github.com/plone/documentation/pull/1292

PR #1238, opened 2022-04-20
Add srcset info to images chapter

MrTango:


docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
 
 The `ImageScaling` view explicitly checks the permissions of the current user.
 To access image scales, which are normally not accessible to the current user, override the `validate_access` method in `plone.namedfile.scaling.ImageScale`.
+
+
+## srcset configuration
```
stevepiercy:
```suggestion
## `srcset` configuration
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+
+In `/@@imaging-controlpanel` Plone allows you to define srcset's.
```
stevepiercy:
```suggestion
In `/@@imaging-controlpanel` Plone allows you to define HTML `srcset` attributes.
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+A srcset can help the Browser to serve the best fitting image for the current users situation.
```
stevepiercy:
```suggestion
A `srcset` can help the browser serve the best fitting image for the current users situation.
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+{
+  "scale": "large",
+  "media": "(max-width:800px)"
+},
+{
+  "scale": "larger"
+}
+```
+
+With this definition the Browser will use the large scale (800px) when the viewport size is max 800px.
```
stevepiercy:
```suggestion
With this definition the browser will use the `large` scale when the viewport's width is 800 pixels or fewer.
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+When the viewport size is greater then 800px, the Browser will use the larger scale (1000px).
```
stevepiercy:
```suggestion
When the viewport width is greater than 800 pixels, the browser will use the larger scale `1000px`.
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+
+
+### Pixel density
+
+Another metric is the pixel density of the users screen.
```
stevepiercy:
```suggestion
Another metric is the pixel density of the user's screen.
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+This metric defines how many DPI per pixel the Screen is using.
```
stevepiercy:
```suggestion
This metric denotes the pixel density, or resolution, of an output device in dots per inch (DPI).
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+{
+  "scale": "huge",
+  "media": "(min-resolution: 192dpi), (-webkit-min-device-pixel-ratio: 2)"
+},
+{
+  "scale": "large"
+}
+```
+
+With this definition the Browser will use the huge scale (1600px) when the screen has a density of 192 DPI per pixel also knows as `2x`.
```
stevepiercy:
```suggestion
With this definition the browser will use the `huge` scale of 1600 pixels when the screen has a density of 192 DPI, also knows as `2x`.
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+We use here two different media queries to also support older Safari versions.
+Mobile devices with Safari like iPhones are still only supporting the old non-standard media query.
```
stevepiercy:
```suggestion
Mobile devices with Safari-like iPhones only support the old non-standard media query.
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+If you don't care about IE support you can also use `min-resolution: 2dppx` which is closer to `2x`.
```
stevepiercy:
```suggestion
If you do not care about IE support, you can use `min-resolution: 2dppx`, which is closer to `2x`.
```

docs/classic-ui/images.md
```diff
@@ -247,3 +247,83 @@ Other options are:
...
+The most common variants are:
+
+- `1x`: 96 DPI per pixel
```
stevepiercy: Remove ` per pixel` after DPI for this and all other instances below.

```suggestion
- `1x`: 96 DPI
```

stevepiercy (changes requested):
Would you please add a Glossary entry in `glossary.md` and make a term for the following?

```
srcset
    The [HTMLImageElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement) property `srcset` is a string which identifies one or more image candidate strings, separated using commas `,`.
    Each image candidate string specifies image resources to display in web pages under given circumstances.

    ```seealso
    https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset
    ```

dots per inch
DPI
    Represents the number of [dots per inch](https://en.wikipedia.org/wiki/Dots_per_inch).
    Screens typically contain 72 or 96 dots per inch.

    ```seealso
    https://developer.mozilla.org/en-US/docs/Web/CSS/resolution#dpi
    ```

stevepiercy (approved):
I pushed a few more commits to save a review cycle. My admonition syntax for the Glossary was off, I added `{term}` to get the links to the Glossary terms, and removed redundant ` per pixel`.

PR #1244, opened 2022-05-04
Add option to redirect

### LINKED ISSUE #1149: Use redirects to avoid link rot with sphinx-reredirects
Configuring the web server for redirects requires an admin.

This Sphinx extension allows the author to add a redirect to the file itself.

See https://github.com/documatt/sphinx-reredirects/issues/3 and open PR https://github.com/documatt/sphinx-reredirects/pull/6.

At this moment we do not have any specific redirects, but we might in the future.

wolbernd:
Fixes #1149


wolbernd:
@stevepiercy At the moment there is no check if the target of a redirect actually exists, so typos can lead to dead links. It should be possible to create a small script to check this but I'm not sure if this is worth the effort...

stevepiercy:
How would you implement that small script? Ideally this extension already checks for the existence of the target file before writing the source page. If it does not, then I would consider this a feature request for the extension.

wolbernd:
@stevepiercy The extension does no consistency check for the target documents. An upstream implementation would make sense because every check-script we would write for this project would probably also work for upstream. 

I'll open an issue in their gitlab.

wolbernd:
Feature request for sphinx-reredirects: https://github.com/documatt/sphinx-reredirects/issues/3 and open PR https://github.com/documatt/sphinx-reredirects/pull/6.

I'll work on implementing this after the sprint since I think it's a bit out of scope

stevepiercy:
Closing as this was previously implemented in #1601.

PR #1252, opened 2022-05-10
Mrtango update responsive images settings docs

MrTango:


mauritsvanrees:
Seems good.

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
 
 
-## `srcset` configuration
+(classic-ui-images-responsive-image-support)=
+## Responsive image support
 
-In `/@@imaging-controlpanel` Plone allows you to define HTML {term}`srcset` attributes.
-A `srcset` can help the browser serve the best fitting image for the current users situation.
-Which image scale is best for the user can be decided on different metrics.
+Plone supports the generation of picture tags with `srcset`'s for image optimization.
```
stevepiercy:
```suggestion
Plone supports the generation of picture tags with `srcset`s for image optimization.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+Additionally you can define `media queries` for {ref}`classic-ui-images-responsive-image-support-art-direction` and further optimization.
```
stevepiercy: The `ref` renders as "Art direction", which is annoying. We can fall back to MyST linking.

```suggestion
Additionally you can define [media queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries) for [art direction](classic-ui-images-responsive-image-support-art-direction) and further optimization.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
 
+The configuration allows to define different picture variants, like `Large`, `Medium`, `Small`.
```
stevepiercy:
```suggestion
The configuration allows you to define different picture variants, such as `Large`, `Medium`, or `Small`.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+Users can choose from them in editors like TinyMCE and Developers can use them in templates.
```
stevepiercy: I'm not sure if Slate is an option, but if it is, then it should be included, else drop it.

```suggestion
Users can choose from them in editors, such as Slate and TinyMCE, and developers can use them in templates.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
 
-### viewport size
 
-The first is metric would be the viewport size.
+(classic-ui-images-responsive-image-support-picture-variants)=
+### Picture variants
+
+In `/@@imaging-controlpanel` Plone allows you to define picture variants with a list of available image scales.
+These are used for HTML {term}`srcset` attributes.
+A `srcset` attribute can help the browser to serve the best fitting image size for the current users situation.
```
stevepiercy:
```suggestion
A `srcset` attribute can help the browser to serve the best fitting image size for the current user's display.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+
+
+### default configuration
```
stevepiercy:
```suggestion
### Default configuration
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+
+The default configuration covers `image size optimization` and will provide the Browser with the needed information to load the optimal image size.
```
stevepiercy: I'm not sure why you put "image size optimization" in backticks. It is not an inline literal. Did you mean to define a Glossary term? That would be good.

```suggestion
The default configuration covers image size optimization and will provide the browser with the needed information to load the optimal image size.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+            "scale": "preview",
+            "additionalScales": ["large", "larger"],
+        ],
+    },
 }
 ```
 
-With this definition the browser will use the `large` scale when the viewport's width is 800 pixels or fewer.
-When the viewport width is greater than 800 pixels, the browser will use the larger scale `1000px`.
+### optional settings
```
stevepiercy:
```suggestion
### Optional settings
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+
+The `sourceset` property is an array and can have more than one entry.
+If we have the following two entries, the `image_srcset` outputfilter will generate one source-tag for each entry and an additional img-tag from the last entry.
```
stevepiercy:
```suggestion
If we have the following two entries, the `image_srcset` outputfilter will generate one `source` tag for each entry and an additional `img` tag from the last entry.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+}
+```
 
-### Pixel density
+(classic-ui-images-responsive-image-support-filtering-scales)=
+#### Filtering scales
 
-Another metric is the pixel density of the user's screen.
-This metric denotes the pixel density, or resolution, of an output device in {term}`dots per inch` (DPI).
+By default for every `srcset` all available scales will be included in the `srcset`.
```
stevepiercy:
```suggestion
By default, for every `srcset`, all available scales will be included in the `srcset`.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+    },
 }
 ```
 
-With this definition the browser will use the `huge` scale of 1600 pixels when the screen has a density of 192 DPI, also knows as `2x`.
-We use here two different media queries to also support older Safari versions.
-Mobile devices with Safari-like iPhones only support the old non-standard media query.
-If you do not care about IE support, you can use `min-resolution: 2dppx`, which is closer to `2x`.
-The most common variants are:
+to restrict the list of used scales inside of a `srcset`, one can set the `additionalScales` parameter with an array of allowed scales.
```
stevepiercy:
```suggestion
To restrict the list of used scales inside of a `srcset`, you can set the `additionalScales` parameter with an array of allowed scales.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+Without this parameter all scales which are not globally excluded scales will be used.
```
stevepiercy:
```suggestion
Without this parameter, all scales which are not globally excluded scales will be used.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+        ],
+    },
+```
 
-- `1x`: 96 DPI
-- `1.25x`: 120 DPI
-- `1.5x`: 144 DPI
-- `2x`: 192 DPI
-- `3x`: 288 DPI
+This means the generated `srcset` will contain the scales from `preview` up to `huge`, but not `mini` for example.
```
stevepiercy:
```suggestion
This means the generated `srcset` will contain the scales from `preview` up to `huge`, but not `mini`, for example.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+This is useful when you want to define a picture variant to be used in templates only.
+
+```json
+    "leadimage": {
+        "title": "Lead image",
+        "sourceset": [
+            {
+              "scale": "preview",
+              "additionalScales": ["large", "larger"],
+              "hideInEditor": True,
```
stevepiercy: I think this edit will fix the syntax so that it is highlighted as `json`. See the preview:

https://deploy-preview-1252--6-dev-docs-plone-org.netlify.app/classic-ui/images.html#hiding-a-picture-variant-in-editors

```suggestion
              "hideInEditor": true,
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+            },
+        ],
+    },
+```
+
+
+(classic-ui-images-responsive-image-support-art-direction)=
+### Art direction
+
+With `image size optimization` the browser is able to choose the optimal image for each situation.
```
stevepiercy: See previous comment about using backticks for this.

```suggestion
With image size optimization, the browser is able to choose the optimal image for each display size.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+But we have no control over which scale the browser will actually use.
+For example to force the Browser to use a zoomed version of an image for smaller screens we can use media queries.
```
stevepiercy:
```suggestion
For example to force the browser to use a zoomed version of an image for smaller screens, we can use media queries.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+        "title": "Small",
+        "sourceset": [
+            {"scale": "preview"},
+        ],
+    },
 }
+
+```
+
+which will result in a srcset like this, for a medium image:
```
stevepiercy:
```suggestion
This will result in a `srcset` as in the following example for a medium image:
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+       class="image-richtext image-size-medium"
+       loading="lazy"
+       src="resolveuid/45fed06defa54d15b37c5b1dc882710c/@@images/image/teaser"
+       width="600"
+       height="400">
+</picture>
 ```
+
+```{note}
+Please notate that this example has the `resolve_uid_and_caption` filter disabled, to see the scale names better.
```
stevepiercy:
```suggestion
Please note that this example has the `resolve_uid_and_caption` filter disabled to see the scale names better.
```

docs/classic-ui/images.md
```diff
@@ -254,81 +254,199 @@ The `ImageScaling` view explicitly checks the permissions of the current user.
...
+The real src-urls look more like `http://localhost:8080/Plone50/dsc04791.jpg/@@images/778f9c06-36b0-485d-ab80-12c623dc4bc3.jpeg`
```
stevepiercy:
```suggestion
The real `src` URLs look more like `http://localhost:8080/Plone50/dsc04791.jpg/@@images/778f9c06-36b0-485d-ab80-12c623dc4bc3.jpeg`.
```

stevepiercy (changes requested):
Excellent work. Thank you!

Mostly just English grammar and syntax, with a couple of questions.

docs/classic-ui/images.md
MrTango: no this is classic ui, there is no slate.

stevepiercy (approved):

PR #1256, opened 2022-05-20
Extend the extrahead block with Matomo analytics tag manager

stevepiercy:
This PR adds the Matomo analytics tag manager by extending the `extrahead` template block and using `super()` to inherit this block from its parent theme.


stevepiercy:
@fredvd please let me know which tag is the correct one to use.

You can also use the Netlify preview link to see if it works as intended:
https://deploy-preview-1256--6-dev-docs-plone-org.netlify.app/

ericof (approved):

PR #1259, opened 2022-06-09
Install improvements

ericof:


stevepiercy (approved):
I don't understand why the three configurations were broken, and how these changes fixes them. Someone who is on top of that should review.

I approve removing the extra column syntax from `backend.md`.

ericof:
@stevepiercy I will merge this, as I was the one adding the faulty configurations in the first place :-)

PR #1265, opened 2022-06-26
add a guide on how to contribute translations

erral:


stevepiercy:
Please search for any existing internationalization documentation with which this PR would conflict.

https://6-dev-docs.plone.org/search.html?q=internationalization

Someone more familiar with i18n in Plone and its packages should verify that part first for conflicts and accuracy.

Once that is complete, then I will review this PR for English and MyST syntax and grammar.

jensens:
This is better than nothing, but very thin. Can't we take some of the old material from old docs and transfer and overhaul it here?

stevepiercy:
@jensens, @erral, and other interested parties:

Assuming the previous i18n content has value, and that someone would be willing to to take on the responsibility to review the content and its organization for accuracy, then I would be happy to take the first step of migrating the previous content from https://docs.plone.org/develop/plone/i18n/index.html by converting it from reStructuredText to MyST. I am not qualified to review the content itself, or its organization and structure. Please let me know.

erral:
I started this because I thought there was nothing else. I will review the existing documentation and restructure it. So if you can convert it to MyST, I will be happy to rework this PR on top of that.

stevepiercy:
I grabbed the files from the `5.2` branch, converted them to MyST, and started cleaning up their English and MyST syntax.

Question for y'all, where the heck should I put them? It's currently 5 files in a directory at the root level of the documentation navigation. i18n is one of those things that touches Classic UI, Volto, and several core packages. I have no idea where it should go. Many of the concepts apply across all components, so it makes sense to keep them grouped together. What do y'all think?

jensens:
> Many of the concepts apply across all components, so it makes sense to keep them grouped together. What do y'all think? 

True. I think this needs either an own section or - if there are more than this topic which covers everything - we need an section for those kind of docs.

erral:
but i18n is different in Volto and Classic, although concepts are the same, the details are important here. I would add a separate section one in Classic and another in Volto.

stevepiercy:
Various packages have their own i18n or l10n implementation details. Yet we need to have one place to present i18n and l10n core concepts. Thus let's present core concepts at the root level, with references to implementation details to each package. Roughly an outline like this:

# i18n and l10n

## Common concepts

- gettext
- translation strings
- locales
- another concept

## Implementation details

- Volto https://6-dev-docs.plone.org/volto/recipes/i18n.html
- plone.restapi https://6-dev-docs.plone.org/plone.restapi/docs/source/i18n.html
- other packages

That will save a lot of duplication of effort, and make it easier for package maintainers.

If other core concepts pop up, we can group them in navigation. Or not. Maybe it is better to keep navigation flat.

ksuess:
Replaced by "Create skeleton framework for internationalization and localization" https://github.com/plone/documentation/pull/1268

PR #1288, opened 2022-09-04
Change port of "make livehtml" to 8050

### LINKED ISSUE #1289: Change port of "make livehtml" to 8050 to avoid conflicts with other apps running

- [ ] 5.x
- [X] 6.x

### Description

was port 8000

ksuess:


stevepiercy (approved):
This is fine. What else uses port 8000, the default for sphinx-autobuild, that causes a conflict?

PR #1296, opened 2022-09-06
Add i18n and l10n

stevepiercy:


stevepiercy:
There we go! This PR should be _much_ easier to review. Only 12 files changed in this round. Sorry for the noise in the previous PR.

Link to preview build:

https://deploy-preview-1296--6-dev-docs-plone-org.netlify.app/i18n-l10n/index.html

docs/i18n-l10n/translating-text-strings.md
```diff
@@ -0,0 +1,447 @@
...
+If you do not add that environment variable, you will need to provide the MO files in your package.
+To make this easier, and if you use [zest.releaser](https://pypi.org/project/zest.releaser/) to publish your packages, you can use [zest.pocompile](https://pypi.org/project/zest.pocompile/).
+This script hooks into the release process and builds the MO files for you.
+
+
+(translating-text-strings-installing-i18ndude-label)=
+
+### Installing i18ndude
+
+The recommended method is to have {term}`i18ndude` installed via your [buildout](https://www.buildout.org/en/latest/).
```
mauritsvanrees, line 276: I am still a fan of buildout, but here I would say: use `bin/pip install i18ndude`. I don't mind though.

mauritsvanrees (dismissed):
LGTM.
I think a few todo markers are left.
I left one minor comment: install i18ndude with pip.

docs/i18n-l10n/translating-text-strings.md
stevepiercy: Is the future direction away from buildout to pip?

These docs refer to buildout in several places, mostly because that was what was used when they were first written many years ago. Should they also be changed to use pip?

Sorry, I am a novice with buildout, and have used it only a few times.

stevepiercy:
The TODOs are an unfortunate thing, but we need to keep them here to let readers know that these instructions might not work 100% and we want pull requests to fix any bugs. I have not actually run through them myself. I was hoping someone could be a test subject. :grin:

erral (dismissed):
I need to check to points regarding the language controlpanel settings that ate already marked as TODO, but it will not possible for me to check them in the coming days.

It is OK for me to merge theis and I will open a new PR to clarify those pending todos.

docs/i18n-l10n/translating-text-strings.md
mauritsvanrees: It's fine to keep the buildout reference then. Those parts can be rewritten later if needed.

stevepiercy:
Thank you both @mauritsvanrees and @erral!

I will merge this for now, and appreciate any time you can spend to fine tune it.

PR #1310, opened 2022-09-12
Plone backend version substitution

stevepiercy:
This PR adds support for replacements within `code-block` directives that are not provided by `rst_epilog` or other substititutions.

The first key allows us to enter a value for `PLONE_BACKEND_VERSION` in `conf.py`, and it will automatically render its value in 25 places.

See https://github.com/plone/plone-backend/pull/66#issuecomment-1243065360 and https://stackoverflow.com/a/56328457/2214933.

h/t @davisagli


stevepiercy:
Ignore this. Bad merge.

PR #1311, opened 2022-09-12
Allow configuration of source replacements.

stevepiercy:
This PR adds support for replacements within `code-block` directives that are not provided by `rst_epilog` or other substititutions.

The first key allows us to enter a value for `PLONE_BACKEND_VERSION` in `conf.py`, and it will automatically render its value in 25 places.

See https://github.com/plone/plone-backend/pull/66#issuecomment-1243065360 and https://stackoverflow.com/a/56328457/2214933.

h/t @davisagli

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -279,3 +279,21 @@ latex_logo = "_static/logo_2x.png"
 html_context = {
     "edit_page_url_template": "https://6.dev-docs.plone.org/contributing/index.html?{{ file_name }}#making-contributions-on-github",
 }
+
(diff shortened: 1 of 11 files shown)


stevepiercy:
ooo, shiny!

https://deploy-preview-1311--6-dev-docs-plone-org.netlify.app/install/containers/images/backend.html

davisagli (approved):
Nice. I suppose this may come at the expense of readability if someone happens to be looking at the source files on github instead of the published docs, but I say we try it.

PR #1319, opened 2022-09-16
Remove mentioning of mxmake.

ksuess:


stevepiercy:
What is the reason for removal of `mxmake`? Or perhaps I should ask, why was it added in the first place?

ksuess:
> What is the reason for removal of `mxmake`? Or perhaps I should ask, why was it added in the first place?

`mxmake` https://github.com/mxstack/mxmake is wrapping `mxdev` and provides for example a Makefile. As there will be a all-In-one installer (I expect https://github.com/collective/cookiecutter-plone-starter), which will have one Makefile for all, backend, frontend, etc.,  there is no need to use `mxmake`. 
In particular https://github.com/plone/documentation/pull/1319/files#diff-2f3ade6d5cd522ccd0e04da61d5b8d3d8d37b5507f3b474ab2a7f0da1c2c58aaL221 is superfluous.

stevepiercy (approved):
Per chat, we don't use `mxmake` at all.

PR #1356, opened 2022-11-08
Update landing page

### LINKED ISSUE #1355: Add a nice one-liner to landing page of Plone 6 documentation

- [ ] 5.x
- [X] 6.x

### Description

https://6.dev-docs.plone.org/

ksuess:
Why "HTML" documentation?
I took the intro from docs.plone.org which seems to fit quite well.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -22,7 +22,7 @@ This documentation is for:
 - Developers who customize a site’s capabilities, create add-ons, and contribute to Plone itself
 
 ```{seealso}
-The documentation for Plone 5 can be found here: https://docs.plone.org/.
+Read the [documentation for the previous version, Plone 5](https://docs.plone.org/).
 ```
 
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -15,7 +15,7 @@ This is the community-maintained documentation for the Plone content management 
 
 This documentation is for:
 
-- Content editors who write, update, and order content on a site
+- Editors who write and update content on a site
 - Site administrators who install Plone and add-ons, and set up a site
 - Designers who create site themes
 - Deployers who configure servers for site hosting
--- a/docs/index.md
+++ b/docs/index.md
@@ -11,7 +11,20 @@ myst:
 
 # Plone 6 Documentation
 
-This website is the HTML documentation of Plone, an open source, enterprise-level content management system.
+This is the community-maintained documentation for the Plone content management system.
+
+This documentation is for:
+
+- Content editors who write, update, and order content on a site
+- Site administrators who install Plone and add-ons, and set up a site
+- Designers who create site themes
+- Deployers who configure servers for site hosting
+- Developers who customize a site’s capabilities, create add-ons, and contribute to Plone itself
+
+```{seealso}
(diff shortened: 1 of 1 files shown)


davisagli (dismissed):
LGTM

docs/index.md
```diff
@@ -11,7 +11,20 @@ myst:
 
 # Plone 6 Documentation
 
-This website is the HTML documentation of Plone, an open source, enterprise-level content management system.
+This is the community-maintained documentation for the Plone content management system.
+
+This documentation is for:
+
+- Content editors who write, update, and order content on a site
```
stevepiercy: "order" is a strange verb here. Perhaps "design" or "lay out" instead?

```suggestion
- Content editors who write, update, and order content on a site
```

docs/index.md
```diff
@@ -11,7 +11,20 @@ myst:
...
+- Site administrators who install Plone and add-ons, and set up a site
+- Designers who create site themes
+- Deployers who configure servers for site hosting
+- Developers who customize a site’s capabilities, create add-ons, and contribute to Plone itself
+
+```{seealso}
+The documentation for Plone 5 can be found here: https://docs.plone.org/.
```
stevepiercy: Usability improvement for visually impaired and screen readers.

```suggestion
Read the [documentation for the previous version, Plone 5](https://docs.plone.org/).
```

stevepiercy (changes requested):
A question and an accessibility improvement.

docs/index.md
ksuess: You are good in finding appropriate wording. What's your first choice?

davisagli: Maybe "write, update, and review"

stevepiercy: @ksuess can you give me more context for your idea of using "order"? What words would you use in your native language? In English "order" can mean "to give a command" or "put into an arrangement".

@davisagli brings up the concept of workflows, such as "review" and "publish", but is that always the job of an Editor? For example, I can edit content on plone.org, but I don't have permission to "publish" or "approve" it unless given explicit permission to do so. Maybe that's getting into the weeds and nitpicky. `¯\_(ツ)_/¯`

ksuess: Oh my, this is getting a big thing.

"order":  I suggest to skip this. I took the text from docs.plone.org. I think the arrangement of content is not a primary task. And this intro should be as short as possible. It's even now longish.

Why not take Davids "write, update, and review"? +1 from me.
Yes, an editor should not be reviewer. But often he is, on smaller sites.

stevepiercy: Heh, the curse of copy-pasta! We know that Plone 5 docs seldom went through quality control over the years, and until now, I never noticed that weird line.

Let's drop "order", remove repetition of "content", and move on. This is good enough for me.

`Editors who write and update content on a site`

davisagli: ~~It would also be fine to just say "...who write and edit content on a site." The end-user docs can go into more detail. :)~~ I wrote that at the same time as Steve, and I like his suggestion.

I don't feel strongly about the wording, any of the above options would work and overall this intro is a nice improvement.

Issue #1347, opened 2022-10-17
Consider using Vale for spellcheck and style guide


stevepiercy:
### Plone Documentation versions affected

- [ ] 5.x
- [X] 6.x

### Description

Our current spellchecker is not well supported. For example, it will mark "add-on" as an incorrectly spelled word because [it does not support hyphenation](https://github.com/sphinx-contrib/spelling/pull/147) among other shortcomings. It also cannot be installed on M1 (and maybe M2) based macs.

@polyester suggested [Vale](https://vale.sh/docs/vale-cli/structure/) as a replacement. I read its docs, and I am thoroughly impressed by its defaults, extensibility, and integration with editors and GitHub Actions as a Docker image.

If others agree, then I would implement it as a guide only to provide feedback to correct common misspellings and style errors, and suggest corrections. I would suggest that authors optionally run Vale after they have completed their first draft, allowing them to both "stay in the zone" and improve their writing quality.

ksuess:
I like your suggestion to use Vale.
For testing I ran Vale against one chapter of the training.
```
 213:29   suggestion  In general, use active voice    Google.Passive      
                      instead of passive voice ('are                      
                      listed').                                           
 226:1    warning     Try to avoid using              Google.We           
                      first-person plural like 'We'.                 
```  
Vale can help me writing better style like the both examples above illustrate.
The CLI is good for a final check. The VSCode extension is helpful while writing.

Apart from a tool for obeying style guides, I have a **spell checker extension enabled in VSCode:** "Code Spell Checker" from "Street Side Software" which is awesome.
"Quick fix": `cmd-.`

![grafik](https://user-images.githubusercontent.com/1144737/196346914-70596253-1d41-4c26-9bcf-4a621d96e4dc.png)

stevepiercy:
Thanks for the feedback.

I also want to try it out before we decide which styles to adopt. Is the passive voice or first-person plural OK to use in some contexts? Trainings has a different context than Documentation, so perhaps "We" is OK in Trainings because we are working together and in person.

For VSCode, I think we should mention that extension in our contributing docs. PyCharm has a built-in spellchecker. Vale integrates with both, so we can add specific words to include or exclude in our spellcheck dictionaries.

Let's keep experimenting and get more feedback from authors. I think it is OK to suggest tools now so that authors can try them and so we can collect feedback from their use.

stevepiercy:
@ksuess @polyester do you have a preference for Vale configuration? I have no preference.

[...]

PR #1366, opened 2022-12-02
Replace sphinxcontrib-spellcheck with Vale

stevepiercy:
Make it clear that the spellchecker is optional, to reduce a barrier to contributors.


fredvd (commented):
Did you already test locally with Vale if that is easy to integrate for the docs? The issue is still that the enchant library is not optional, when I test the instructions by checking out the documentation repo in a clean dir and run `make html` I get the following error:

```
Documentation of plone.api initialized.
cd ./docs/ && /Users/fred/test/documentation/bin/sphinx-build -b html -d ../_build//doctrees   . ../_build//html
Running Sphinx v4.5.0

Extension error:
Could not import extension sphinxcontrib.spelling (exception: The 'enchant' C library was not found and maybe needs to be installed.
See  https://pyenchant.github.io/pyenchant/install.html
for details
)
make: *** [html] Error 2
```

stevepiercy:
> Did you already test locally with Vale if that is easy to integrate for the docs?

See https://github.com/plone/documentation/issues/1347 which is a link in this PR. ;)

> The issue is still that the enchant library is not optional, when I test the instructions by checking out the documentation repo in a clean dir and run `make html` I get the following error:

Ah, that's right. I already have it installed, so I missed that. I should remove that step from the `Makefile` for now, too, in this PR. There are actually a bunch of other places where it needs to be removed, including volto. I will clean this up tomorrow. Thank you!

I think we are agreed, however, that the current spellcheck is broken and lacks a maintainer, and despite my request for guidance to fix it, I have received no response from the pyenchant maintainer. I have no hope of it getting fixed at this point.

stevepiercy:
Instead of just ripping out `sphinxcontrib.spelling`, enchant, and pyenchant, I'm going to do a comprehensive replacement with Vale, either this or the next week.

stevepiercy:
@ksuess @fredvd @polyester this is ready for review.

I kept the file `docs/spelling_wordlist.txt` because it has some good words that we should consider moving into our custom Vale dictionaries eventually, but not in this PR. We ought to leave something for newbies to do!

stevepiercy:
I am not sure yet whether to install vale or use a [GitHub Action](https://github.com/errata-ai/vale-action). For the latter, I would have to change some settings.

For now I removed it from the `test` target in Makefile.

ksuess:
> I am not sure yet whether to install vale or use a [GitHub Action](https://github.com/errata-ai/vale-action). For the latter, I would have to change some settings.

The GitHub Action seems to be maintained. So +1 from me because this means less to care for for us/you.

> 
> For now I removed it from the `test` target in Makefile.

ksuess:
> I kept the file `docs/spelling_wordlist.txt` because it has some good words that we should consider moving into our custom Vale dictionaries eventually, but not in this PR. We ought to leave something for newbies to do!

yes, this can go reviewed in vocab/**/accept.txt

ksuess (approved):

stevepiercy:
To use the GitHub Action, I would need to translate the command from Makefile to get the list of files to lint, and set an environment variable:

`VALEFILES       := $(shell find -L $(DOCS_DIR) -type d \( -path $(DOCS_DIR)/plone.restapi/lib/* -o  -path $(DOCS_DIR)"/plone.restapi/performance/*" \) -prune -false -o -type f -name "*.md" -print)`

How would I translate that to GitHub Actions? I have no idea.

It might be easier to just install Vale where all the other tests are run in CI and use `make vale`.

PR #1372, opened 2022-12-11
Behavior chapter from scratch

jensens:


docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
 
 # Behaviors
 
+## What are Behaviors in Plone
+
+In Plone, behaviors are a way to add additional, reusable functionality to content objects without modifying the objects themselves.
```
stevepiercy:
```suggestion
In Plone, behaviors are a way to add reusable functionality to content objects without modifying the objects themselves.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+- enable a particular event handler,
+- enable one or more views, viewlets, or other UI components,
+- do anything else which may be expressed in code via an adapter and/or marker interface.
+
+Behaviors can be added to content types on an as-needed basis, allowing for a high degree of flexibility and customization.
+
+Plone already provides lots of behaviors for the built-in content types.
+
+Other behaviors are implemented as add-on products, which can be installed and configured through the Plone control panel.
+Once a behavior has been installed, it can be applied to any content type by selecting it in the Dexterity configuration section of the Plone control panel.
```
stevepiercy: I think the name Dexterity has been dropped in Plone 6, leaving it as Content Types.

```suggestion
Once a behavior has been installed, it can be applied to any content type by selecting it in the {guilabel}`Content Types` control panel.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+| plone.eventrecurrence  | Event Recurrence | Recurrence extension for Events. |
+| plone.excludefromnavigation  | Exclude From navigation | Allow items to be excluded from navigation |
+| plone.constraintypes| Folder Addable Constrains | Restrict the content types that can be added to folderish   content |
+| plone. textindexer | Full-Text Indexing | Enables the enhanced full-text indexing for a content type. If a field in the schema is marked with the `searchable` directive, its content gets added to the    `SearchableText` index in the catalog |
+| plone.leadimage  | Lead Image | Adds image and image caption fields |
+| plone.locking  | Locking | Locking support for dexterity |
+| plone.translatable  | Multilingual Support | Make this content type multilingual aware. Multilingual support must be installed. |
+| plone.namefromfilename | Name from file name | Automatically generate short URL name for content based on its    primary field file name
+| plone.namefromtitle | Name from title | Automatically generate short URL name for content based on its initial    title
+| plone.navigationroot  | Navigation root | Make all items of this type a navigation root |
```
stevepiercy:
```suggestion
| plone.navigationroot  | Navigation root | Make all items of this type a navigation root |
| plone.__________  | Navigation title | Navigation title used in sections, menus and doormats |
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+| plone.nextpreviousenabled  | Next previous navigation | Enable next previous navigation for all items of this type |
+| plone.nextprevioustoggle  | Next previous navigation toggle | Allow items to have next previous navigation enabled |
+| plone.ownership  | Ownership | Adds creator, contributor, and rights fields. |
```
stevepiercy:
```suggestion
| plone.ownership  | Ownership | Adds creator, contributor, and rights fields. |
| plone.____  | Preview Image | Preview image for listings |
| plone.____  | Preview Image Link | Preview image for listings based on links |
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+| plone.publication  | Date range | Adds effective date and expiration date fields. |
+| plone.dublincore | Dublin Core metadata | Adds standard metadata fields (equals Basic metadata +    Categorization + Effective range + Ownership  |
+| plone.eventattendees  | Event Attendees | Attendees extension for Events. |
+| plone.eventbasic  | Event Basic | Basic Event schema. |
+| plone.eventcontact  | Event Contact | Contact extension for Events. |
+| plone.eventlocation  | Event Location | Location extension for Events. |
+| plone.eventrecurrence  | Event Recurrence | Recurrence extension for Events. |
+| plone.excludefromnavigation  | Exclude From navigation | Allow items to be excluded from navigation |
+| plone.constraintypes| Folder Addable Constrains | Restrict the content types that can be added to folderish   content |
+| plone. textindexer | Full-Text Indexing | Enables the enhanced full-text indexing for a content type. If a field in the schema is marked with the `searchable` directive, its content gets added to the    `SearchableText` index in the catalog |
```
stevepiercy:
```suggestion
| plone. textindexer | Full-Text Indexing | Enables the enhanced full-text indexing for a content type. If a field in the schema is marked with the `searchable` directive, its content gets added to the    `SearchableText` index in the catalog |
| plone.____  | Head title field | Adds Head title field |
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+This allows the object to gain the additional functionality provided by the behavior.
+
+Overall, behaviors are an important part of the Plone content management system and allow for powerful customization and extensibility of content objects.
+
+## Built-in behaviors
+
+| short name  | Title            | Desription                    |
+|-------------|------------------|-------------------------------|
+| plone.allowdiscussion  | Allow discussion | Allow discussion on this item |
+| plone.basic  | Basic metadata | Adds title and description fields. |
```
stevepiercy:
```suggestion
| plone.basic  | Basic metadata | Adds title and description fields. |
| plone.___?  | Blocks | Enables Volto Blocks support |
| plone.___?  | Blocks (Editable Layout) | Enables Volto Blocks (editable layout) support |
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+## Adding or removing a behavior from a type
+
+There are two ways to add or remove a behavior on a type:
+
+- through the web using the control panel,
+- using a custom add-on GenericSetup profile.
+
+### Through the Web
+
+1. Go to the {guilabel}`Site Setup` and chose the {guilable}`Content Types` control panel
```
stevepiercy:
```suggestion
1. Go to the {guilabel}`Site Setup` and chose the {guilabel}`Content Types` control panel.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+| plone.nextprevioustoggle  | Next previous navigation toggle | Allow items to have next previous navigation enabled |
+| plone.ownership  | Ownership | Adds creator, contributor, and rights fields. |
+| plone.relateditems  | Related items | Adds the ability to assign related items |
+| plone.richtext  | RichText | Adds richtext behavior |
+| plone.shortname  | Short name | Gives the ability to rename an item from its edit form. |
+| plone.tableofcontents  | Table of contents | Adds a table of contents |
+| plone.thumb_icon | Thumbs and icon handling | Options to suppress thumbs and/or icons and to override thumb   size in listings, tables etc.
+| plone.versioning  | Versioning | Versioning support with CMFEditions |
+
+## Adding or removing a behavior from a type
```
stevepiercy:
```suggestion
## Adding or removing a behavior from a content type
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+There are two ways to add or remove a behavior on a type:
```
stevepiercy:
```suggestion
There are two ways to add or remove a behavior on a content type:
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+- through the web using the control panel,
```
stevepiercy:
```suggestion
- Through the web using the {guilabel}`Content Types` control panel.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+- using a custom add-on GenericSetup profile.
```
stevepiercy:
```suggestion
- Using a custom add-on GenericSetup profile.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+### Through the Web
+
+1. Go to the {guilabel}`Site Setup` and chose the {guilable}`Content Types` control panel
+2. Select the type where you want to add or remove a behavior
```
stevepiercy:
```suggestion
2. Select the content type to which you want to add or remove a behavior.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+3. Then click on the {guilabel}`Behaviors` tab of the settings of the type.
```
stevepiercy:
```suggestion
3. Then click on the {guilabel}`Behaviors` tab of the settings of the content type.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+4. A list of all available behaviors appears.
+   Select or unselect the checkbox of the behavior you want to add to or remove from the type.
+5. Save the form by clicking on the {guilabel}`Save` button at the bottom of the page.
+
+### Using a GenericSetup profile
+
+Given you already have a custom add-on with a `profiles/default` directory.
+Given you created a custom behavior named `mybehavior.subtitle`.
```
stevepiercy:
```suggestion
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+1. Go to the {guilabel}`Site Setup` and chose the {guilable}`Content Types` control panel
+2. Select the type where you want to add or remove a behavior
+3. Then click on the {guilabel}`Behaviors` tab of the settings of the type.
+4. A list of all available behaviors appears.
+   Select or unselect the checkbox of the behavior you want to add to or remove from the type.
+5. Save the form by clicking on the {guilabel}`Save` button at the bottom of the page.
+
+### Using a GenericSetup profile
+
+Given you already have a custom add-on with a `profiles/default` directory.
```
stevepiercy:
```suggestion
Given you already have a custom add-on with a `profiles/default` directory, and you created a custom behavior named `mybehavior.subtitle`.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+Given you created a custom behavior named `mybehavior.subtitle`.
+
+If you want to enable a behavior on an existing content type, create a new directory `types` under `profiles/default`.
+In there create a file named the same as the content type you want to change.
+In the example here, you want to add a behavior to the built-in Event type.
```
stevepiercy:
```suggestion
In the example here, you want to add a behavior to the built-in Event content type.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+The file to create is named `Event.xml` and it is a Factory Type Information definition.
+You need to change the behaviors configuration only so that all other parts can be omitted.
```
stevepiercy:
```suggestion
You need to change only the behaviors configuration.
All other parts can be ignored.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+This looks like so:
+
+```XML
```
stevepiercy:
````suggestion
```xml
````

File: docs/backend/behaviors.md [hunk 19]
```diff
@@ -11,12 +11,305 @@ myst:
...
+
```
stevepiercy, line 101:
```suggestion
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+    xmlns:i18n="http://xml.zope.org/namespaces/i18n">
+  <property name="behaviors" purge="false">
+    <element value="myproject.subtitle" />
+  </property>
+</object>
+```
+
+After you apply the profile (or uninstall and install the custom add-on), the behavior is available on the Event content type.
+
+## How behaviors are working
```
stevepiercy:
```suggestion
## How behaviors work
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+At the most basic level, a behavior is like a conditional adapter.
+For a Dexterity content type, the default condition is, "is this behavior listed in the behaviors property in the FTI?"
```
stevepiercy:
```suggestion
For a content type, the default condition is, "is this behavior listed in the behaviors property in the FTI?"
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+### Using a GenericSetup profile
+
+Given you already have a custom add-on with a `profiles/default` directory.
+Given you created a custom behavior named `mybehavior.subtitle`.
+
+If you want to enable a behavior on an existing content type, create a new directory `types` under `profiles/default`.
+In there create a file named the same as the content type you want to change.
+In the example here, you want to add a behavior to the built-in Event type.
+The file to create is named `Event.xml` and it is a Factory Type Information definition.
```
stevepiercy: Please add a term and definition to the Glossary in `glossary.md`. Here's markup for this bit.

```suggestion
The file to create is named `Event.xml` and it is a {term}`Factory Type Information` (FTI) definition.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+    <element value="myproject.subtitle" />
+  </property>
+</object>
+```
+
+After you apply the profile (or uninstall and install the custom add-on), the behavior is available on the Event content type.
+
+## How behaviors are working
+
+At the most basic level, a behavior is like a conditional adapter.
```
stevepiercy: Adapter should be added to `glossary.md`. Do we have a definition somewhere? I could not find one in https://5.docs.plone.org/appendices/glossary.html

```suggestion
At the most basic level, a behavior is like a conditional {term}`adapter`.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+For a Dexterity content type, the default condition is, "is this behavior listed in the behaviors property in the FTI?"
+But the condition itself is an adapter; in rare cases, this can be overruled.
+When a behavior is enabled for a particular object, it will be possible to adapt that object to the behavior's interface.
+If the behavior is disabled, adaptation will fail.
+
+A behavior consists at the very least of an interface and some metadata, namely a name, title, and description.
+This is also called a schema-only interface
```
stevepiercy:
```suggestion
This is also called a schema-only interface.
```

docs/backend/behaviors.md
stevepiercy: I started reading this section, and my eyes glazed over.

Who is the audience for this section?

What is the source of this section?

I have very serious concerns about this section of the page. It smells of copy-pasta. It is "word salad" to me. The terms are technical, have no definitions, and lack a base context to help the reader understand them.

I think it may be better to remove it, keeping on the reference to:

https://github.com/plone/plone.behavior/blob/master/README.rst

@polyester what would you suggest?

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+In this case, the behavior interface and the marker interface must be the same.
+
+If a factory is given a marker interface different from the behavior interface must be declared!
+
+Behaviors are registered globally, using the `<plone.behavior />` ZCML directive.
+This results in, among other things, a named utility providing `plone.behavior.interfaces.IBehavior` being registered.
+This utility contains various information about the behavior, such as its name, title, interface, and (optional) factory and marker interface.
+
+The utility name is the name attribute of the behavior directive.
+Historically the full dotted name to the behavior interface is registered as name too, but usage is discouraged.
```
stevepiercy:
```suggestion
Historically the full dotted name to the behavior interface is registered as its name, too, but this usage is discouraged.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+
+Behaviors are named adapters.
+They adapt an interface, for which they are registered and they provide another interface with new functionality such as attributes and methods.
+
+Dexterity types are transparently looking up behaviors registered for a type.
```
stevepiercy: This is a circular reference. Should it be deleted?

```suggestion
Content types transparently look up the behaviors registered for a content type.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+If a behavior provides new attributes it is provided as an attribute of the adapted object.
```
stevepiercy:
```suggestion
If a behavior provides new attributes, it is provided as an attribute of the adapted object.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
 ```
 
-```{todo}
-Contribute to this documentation!
-See issue [Backend > Behaviors needs content](https://github.com/plone/documentation/issues/1302).
-```
\ No newline at end of file
+## Custom behaviors
+
+From the last section you can take away, that there are two main types of behaviors:
```
stevepiercy:
```suggestion
There are two types of behaviors:
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+- Schema-only behaviors: These behaviors have only a schema with fields.
```
stevepiercy:
```suggestion
Schema-only behaviors
: These behaviors have only a schema with fields.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+- Full behaviors: A python class containing the logic of the behavior, an interface or schema defining the contract of the behavior, and a marker interface applied to the content type.
```
stevepiercy:
```suggestion
Full behaviors
: A python class containing the logic of the behavior, an interface or schema defining the contract of the behavior, and a marker interface applied to the content type.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+### Creating a schema-only behavior
+
+Given you want to add a field `subtitle` to some existing content types of your custom add-on.
+
+You need to create a file `subtitle.py` in your addon like so:
```
stevepiercy:
```suggestion
You need to create a file `subtitle.py` in your add-on:
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+    subtitle = schema.Text(
+        title="Subtitle",
+        description="A title to be displayed below the title",
+        default="",
+        required=False,
+    )
+```
+
+You need to add a ZCML snippet to the `configure.zcml` next to `subtitle.py` like so:
```
stevepiercy:
```suggestion
You need to add a ZCML snippet to the `configure.zcml` next to `subtitle.py`:
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+```XML
```
stevepiercy:
````suggestion
```xml
````

File: docs/backend/behaviors.md [hunk 34]
```diff
@@ -11,12 +11,305 @@ myst:
...
+<plone:behavior
+      name="myproject.subtitle"
+      provides=".subtitle.ISubtitleBehavior"
+      title="Subtitle"
+  />
+```
+
+After a restart of Plone, the behavior can be added to the type in the Content Types control panel.
```
stevepiercy:
```suggestion
After a restart of Plone, the behavior can be added to the content type in the {guilabel}`Content Types` control panel.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+The add and edit forms are containing a new field `Subtitle`.
```
stevepiercy:
```suggestion
The add and edit forms contain a new field `Subtitle`.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+This field is not displayed in most views.
+To display the data entered in this field you need to modify the page template and access the field as `context.subtitle` there.
```
stevepiercy:
```suggestion
To display the data entered in this field, you need to modify the page template by adding the field `context.subtitle`.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+### Creating a behavior with an adapter and factory
+
+Given you want to display a price with different content types.
+The price is stored as the net value on the type as a floating point number.
+For display, you need at several places the VAT and the gross value.
```
stevepiercy:
```suggestion
For display, you need at several places the value added tax (VAT) and the gross value.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+You create a schema with the net value for the form and attributes for the calculated values.
+You create an adapter to calculate the VAT and gross values.
+You need a marker interface to distinguish between context and adapter.
+
+The Python code in a file `price.py` looks like so:
```
stevepiercy:
```suggestion
Add Python code in the file `price.py`:
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+      factory=".price.PriceAdapter"
+      for=".price.IPriceMarker"
+      marker=".price.IPriceMarker"
+      name="myproject.price"
+      provides=".price.IPriceBehavior"
+      title="Price with net, VAT and gross"
+  />
+  ```
+
+After a restart of Plone, the behavior can be added to the type in the Content Types control panel.
```
stevepiercy:
```suggestion
After a restart of Plone, the behavior can be added to the content type in the {guilabel}`Content Types` control panel.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+The add and edit forms are containing a new field `Price (net)`.
```
stevepiercy:
```suggestion
The add and edit forms contain a new field `Price (net)`.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+This field is not displayed in most views.
+To display the data entered in this field you need to modify the page template and access the `price_net` field as `context.price_net` there.
```
stevepiercy:
```suggestion
To display the data entered in this field, you need to modify the page template by adding the `price_net` field as `context.price_net`.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+To access the `price_vat` and `price_gross` fields you need to get the adapter in your view class like so:
```
stevepiercy:
```suggestion
To access the `price_vat` and `price_gross` fields, you need to get the adapter in your view class:
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+    def gross(self):
+        adapter = IPriceBehavior(context)
+        return adapter.price_gross
+```
+
+### Creating a behavior with PloneCLI
+
+To add a behavior to your add-on, you can use PloneCLI as follows:
+
+```bash
```
stevepiercy: We use the generic lexer `shell`.

````suggestion
```shell
````

File: docs/backend/behaviors.md [hunk 44]
```diff
@@ -11,12 +11,305 @@ myst:
...
+plonecli add behavior
+```
+
+This will create the behavior Python file in the `behaviors` folder where you can define your behaviors schema fields and registers the behavior in the `configure.zcml`.
```
stevepiercy:
```suggestion
This will create the behavior Python file in the `behaviors` folder, where you can define your behaviors schema fields, and registers the behavior in the `configure.zcml`.
```

stevepiercy (changes requested):
Thank you! This is good stuff. Here is my first pass review.

- Dexterity name is no longer used.
- Added some behaviors per https://6-classic.demo.plone.org/dexterity-types/Document/@@behaviors. This should be double-checked because the Demo might have additional behaviors that are not considered "built-in". Also I do not know the new behaviors' short names.
- Needs some Glossary terms
- Serious concern about section "How behaviors work"

The rest of my suggestions are MyST and English syntax and grammar fixes.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
 
 # Behaviors
 
+## What are Behaviors in Plone
+
+In Plone, behaviors are a way to add additional, reusable functionality to content objects without modifying the objects themselves.
+Behaviors are essentially small chunks of code that can be plugged onto content objects to provide new features or capabilities.
```
davisagli: Should the above say "content types" everywhere it currently says "content objects"?

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+A Plone behavior could be used to
+
+- add a set of form fields (on standard add and edit forms),
+- add logic as part of the adapter,
```
davisagli, line 20: Instead of this, maybe "add logic for how field values are read or written"

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+- enable one or more views, viewlets, or other UI components,
+- do anything else which may be expressed in code via an adapter and/or marker interface.
+
+Behaviors can be added to content types on an as-needed basis, allowing for a high degree of flexibility and customization.
+
+Plone already provides lots of behaviors for the built-in content types.
+
+Other behaviors are implemented as add-on products, which can be installed and configured through the Plone control panel.
+Once a behavior has been installed, it can be applied to any content type by selecting it in the Dexterity configuration section of the Plone control panel.
+This allows the object to gain the additional functionality provided by the behavior.
```
davisagli:
```suggestion
This allows items of this content type to gain the additional functionality provided by the behavior.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+
+Overall, behaviors are an important part of the Plone content management system and allow for powerful customization and extensibility of content objects.
```
davisagli, line 34: A key feature of behaviors is that they allow _encapsulating_ functionality so that it can be reused for multiple content types without needing to reimplement it. It might be worth highlighting this more.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+| plone.navigationroot  | Navigation root | Make all items of this type a navigation root |
+| plone.nextpreviousenabled  | Next previous navigation | Enable next previous navigation for all items of this type |
+| plone.nextprevioustoggle  | Next previous navigation toggle | Allow items to have next previous navigation enabled |
+| plone.ownership  | Ownership | Adds creator, contributor, and rights fields. |
+| plone.relateditems  | Related items | Adds the ability to assign related items |
+| plone.richtext  | RichText | Adds richtext behavior |
+| plone.shortname  | Short name | Gives the ability to rename an item from its edit form. |
+| plone.tableofcontents  | Table of contents | Adds a table of contents |
+| plone.thumb_icon | Thumbs and icon handling | Options to suppress thumbs and/or icons and to override thumb   size in listings, tables etc.
+| plone.versioning  | Versioning | Versioning support with CMFEditions |
```
davisagli: We could probably write at least a paragraph about each of these explaining what it does in more detail. But this is fine for a start. And maybe the detailed explanation of standard/core behaviors belongs elsewhere than the overview of how to use behaviors.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+- through the web using the control panel,
+- using a custom add-on GenericSetup profile.
+
+### Through the Web
+
+1. Go to the {guilabel}`Site Setup` and chose the {guilable}`Content Types` control panel
+2. Select the type where you want to add or remove a behavior
+3. Then click on the {guilabel}`Behaviors` tab of the settings of the type.
+4. A list of all available behaviors appears.
+   Select or unselect the checkbox of the behavior you want to add to or remove from the type.
```
davisagli:
```suggestion
   Select or deselect the checkbox of the behavior you want to add to or remove from the type.
```

https://english.stackexchange.com/questions/18465/unselect-or-deselect

docs/backend/behaviors.md
```diff
@@ -11,12 +11,305 @@ myst:
...
+5. Save the form by clicking on the {guilabel}`Save` button at the bottom of the page.
+
+### Using a GenericSetup profile
+
+Given you already have a custom add-on with a `profiles/default` directory.
+Given you created a custom behavior named `mybehavior.subtitle`.
+
+If you want to enable a behavior on an existing content type, create a new directory `types` under `profiles/default`.
+In there create a file named the same as the content type you want to change.
```
davisagli:
```suggestion
In the `types` directory, create a file named the same as the content type you want to change.
```

docs/backend/behaviors.md
jensens: in fact I had the possibility of instance behaviors in mind, but its confusing, so better: 
```suggestion
Behaviors are essentially small chunks of code that can be plugged onto content types to provide new features or capabilities.
```

docs/backend/behaviors.md
jensens: It misses adding new logic, what about:
```suggestion
- add logic for how field values are read/ written,
- add methods for calculating values or fetch data from other sources,
```
seconds could be written better I guess, any native speaker here to help ;) ?

docs/backend/behaviors.md
jensens, line 34: I like the sentence as is:
```suggestion
A key feature of behaviors is that they allow encapsulating functionality so that it can be reused for multiple content types without needing to reimplement it.
Overall, behaviors are an important part of the Plone content management system and allow for powerful customization and extensibility of content objects.
```

docs/backend/behaviors.md
jensens: I had to have a look at the Volto demo site while writing it. Unfortunately  its named `volto.blocks` and `volto.blocks.editable.layout` (who named this!)

docs/backend/behaviors.md
jensens:
```suggestion
| plone.versioning  | Versioning | Versioning support with CMFEditions |
| volto.blocks  | Blocks | Enables Volto Blocks support |
| volto.blocks.editable.layout  | Blocks (Editable Layout) | Enables Volto Blocks (editable layout) support |
```

docs/backend/behaviors.md
jensens: this is `volto.navtitle`

docs/backend/behaviors.md
jensens: This are `volto.previewimage` and `volto.preview_image_link`.

docs/backend/behaviors.md
jensens: Yes, but +1 to let get this in PR here in first and then start to enhance.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,308 @@ myst:
...
+| plone.nextprevioustoggle  | Next previous navigation toggle | Allow items to have next previous navigation enabled |
+| plone.ownership  | Ownership | Adds creator, contributor, and rights fields. |
+| plone.relateditems  | Related items | Adds the ability to assign related items |
+| plone.richtext  | RichText | Adds richtext behavior |
+| plone.shortname  | Short name | Gives the ability to rename an item from its edit form. |
+| plone.tableofcontents  | Table of contents | Adds a table of contents |
+| plone.thumb_icon | Thumbs and icon handling | Options to suppress thumbs and/or icons and to override thumb   size in listings, tables etc.
+| plone.versioning  | Versioning | Versioning support with CMFEditions |
+| volto.blocks  | Blocks | Enables Volto Blocks support |
+| volto.blocks.editable.layout  | Blocks (Editable Layout) | Enables Volto Blocks (editable layout) support |
```
jensens:
```suggestion
| volto.blocks  | Blocks | Enables Volto Blocks support |
| volto.blocks.editable.layout  | Blocks (Editable Layout) | Enables Volto Blocks (editable layout) support |
| volot.navtitle  | Navigation title | Navigation title used in sections, menus and doormats |
| volto.preview_image  | Preview Image | Preview image for listings |
| volto.preview_image_link | Preview Image Link | Preview image for listings based on links |
```

docs/backend/behaviors.md
jensens: this is `volto.head_title`

docs/backend/behaviors.md
```diff
@@ -11,12 +11,311 @@ myst:
...
+| volot.navtitle  | Navigation title | Navigation title used in sections, menus and doormats |
```
jensens:
```suggestion
| volto.head_title  | Head title field | Adds Head title field |
| volto.navtitle  | Navigation title | Navigation title used in sections, menus and doormats |
```

docs/backend/behaviors.md
stevepiercy: An emerging trend in how to write documentation in English is to use simple verb tense.

"Writing documentation" becomes "How to write documentation"

This is due in large part to how non-English speaking people search for documentation.

"how to do thing"

```suggestion
- add logic for how to read or write field values,
```

docs/backend/behaviors.md
stevepiercy: > And maybe the detailed explanation of standard/core behaviors belongs elsewhere than the overview of how to use behaviors.

Indeed it would. https://github.com/plone/documentation/issues/1380

docs/backend/behaviors.md
stevepiercy: Agreed, and with the insertion of @davisagli's suggestion, it all ties together nicely.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,312 @@ myst:
...
+| plone.publication  | Date range | Adds effective date and expiration date fields. |
+| plone.dublincore | Dublin Core metadata | Adds standard metadata fields (equals Basic metadata +    Categorization + Effective range + Ownership  |
+| plone.eventattendees  | Event Attendees | Attendees extension for Events. |
+| plone.eventbasic  | Event Basic | Basic Event schema. |
+| plone.eventcontact  | Event Contact | Contact extension for Events. |
+| plone.eventlocation  | Event Location | Location extension for Events. |
+| plone.eventrecurrence  | Event Recurrence | Recurrence extension for Events. |
+| plone.excludefromnavigation  | Exclude From navigation | Allow items to be excluded from navigation |
+| plone.constraintypes| Folder Addable Constrains | Restrict the content types that can be added to folderish   content |
+| plone. textindexer | Full-Text Indexing | Enables the enhanced full-text indexing for a content type. If a field in the schema is marked with the `searchable` directive, its content gets added to the    `SearchableText` index in the catalog |
```
stevepiercy:
```suggestion
| plone.textindexer | Full-Text Indexing | Enables the enhanced full-text indexing for a content type. If a field in the schema is marked with the `searchable` directive, its content gets added to the `SearchableText` index in the catalog |
```

docs/backend/behaviors.md
stevepiercy: Note that the table is no longer sorted. It was originally sorted by its Title, because that is what is displayed to the user. I would prefer that sort order be restored, unless there is a good reason not to do that.

Also please remove double-spaces in the table, as that shall appear to be intentional to the next author and copy-pasta may creep in.

stevepiercy (changes requested):
Thanks for the taking the effort to get all the details. I have a couple of minor corrections, and then I think it is good to go.

jensens:
> * Serious concern about section "How behaviors work"

@stevepiercy OK, I rewrote the whole section. It was indeed confusing, but explaining this short and to the point is really difficult.

docs/backend/behaviors.md
jensens: naturally I would expect it to be sorted by behavior name, at least as developer I would be confused if it is sorted by title.

docs/backend/behaviors.md
jensens: so, combined both are:
```suggestion
- add logic for how to read or write field values,
- add logic for how to calculate values or fetch data from other sources,
```

jensens:
and I moved the `how behaviors work` to the end.

docs/backend/behaviors.md
stevepiercy: We should define who is the audience. Is it an end user, a developer, or another role?

Here is what the end user sees, sorted by `Title`, with no `short name` displayed:

https://6-classic.demo.plone.org/dexterity-types/Document/@@behaviors

If a developer, where would a developer find the list sorted by `short name`?

I'm fine with whatever makes the most sense for the intended audience.

docs/backend/behaviors.md
stevepiercy: Yup!

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+
+In Plone, behaviors can be globally enabled on content types at runtime.
+With add-ons, behaviors can even be enabled even on a single content object or for a whole subtree in the content hierarchy.
+
+### Interfaces and adapters
+
+When a behavior is enabled for a particular object, it will be possible to adapt that object to the behavior's interface.
+Otherwise, when the behavior is disabled, adaptation will fail or falls back to a more generic adapter, if any is registered.
+
+A behavior is at least a combination of an interface (also as a form field provider), metadata such as a name, title, and description, and sometimes an adapter factory with a marker interface.
```
stevepiercy: English punctuation supports nested listings.

```suggestion
A behavior is at least a combination of an interface (also as a form field provider); metadata such as a name, title, and description; and sometimes an adapter factory with a marker interface.
```

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+When a behavior is enabled, an interface is added to the content object to indicate its presence - it now provides the interface.
```
stevepiercy:
```suggestion
When a behavior is enabled, an interface is added to the content object to indicate its presence.
In other words, the content object now provides the interface.
```

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+
+Behaviors without an adapter factory can be used either as a simple marker or to provide additional form fields.
+In this case, adapting a content object with this interface returns the content object itself, because adapting an object that already provides the exact same interface returns the very same object.
+Based on the now-provided interface, i.e specific views can be registered with the content type, or event handlers can be registered to respond to specific actions.
```
stevepiercy:
```suggestion
Based on the now-provided interface, specific views can be registered with the content type, or event handlers can be registered to respond to specific actions.
```

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+
+In other cases, there is also an adapter factory (usually a class), which will be invoked (initialized) to get an appropriate adapter when requested.
```
stevepiercy:
```suggestion
In other cases, there is also an adapter factory (usually a Python class), which will be invoked (initialized) to get an appropriate adapter when requested.
```

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+If an adapter factory is used an explicit marker interface is required.
```
stevepiercy:
```suggestion
If an adapter factory is used, an explicit marker interface is required.
```

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+
+With an adapter factory in place, custom getters and setters for form fields can be implemented, or even new methods i.e. for calculations or to combine data can be added.
```
stevepiercy:
```suggestion
With an adapter factory in place, custom getters and setters for form fields can be implemented, or even new methods.
For example, calculations or the combination of data can be added.
```

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+
+### Registration
+
+Behaviors are registered globally using the <plone.behavior /> {term}`ZCML` directive.
```
stevepiercy:
```suggestion
Behaviors are registered globally using the `<plone.behavior />` {term}`ZCML` directive.
```

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+Internally, this directive registers a named utility that provides `plone.behavior.interfaces.IBehavior`, which contains combined information about the behavior such as its name, interface, factory/marker interface, and metadata.
```
stevepiercy:
```suggestion
Internally, this directive registers a named utility that provides `plone.behavior.interfaces.IBehavior`.
This utility contains combined information about the behavior, such as its name, interface, factory or marker interface, and metadata.
```

docs/backend/behaviors.md
```diff
@@ -315,8 +278,50 @@ plonecli add behavior
...
+
+```{seealso}
+The [README file of `plone.behavior`](https://github.com/plone/plone.behavior/blob/master/README.rst) explains the concepts and different ways to register a behavior in detail.
+```
+
+### Lookup and provide
+
+Plone content objects have logic to look up the behaviors names registered from their types configuration, the Factory Type Information (FTI).
```
stevepiercy:
```suggestion
Plone content objects have logic to look up the behaviors' names registered from their types' configuration, the Factory Type Information (FTI).
```

stevepiercy (changes requested):
I'd like someone more familiar with the jargon to review the rewritten "How behaviors work" section. This version is an improvement.

Other than the suggestions in this review, the only remaining thing is a response to https://github.com/plone/documentation/pull/1372/files#r1050266634 and its resolution.

docs/backend/behaviors.md
jensens: An end user would see the same without the names in the UI. So I would say this here is is more for Devs as reference.

stevepiercy: That's OK. We need to tell the developer where to find the list. My concern as a docs author is not being able to keep this list up to date because I have no clue where these short names are maintained. The only place I know where to look at this moment is for the end user, not developer.

jensens: If we would have a central reference  section it would be easier to find. In the current structure reference information is spread around. But IMO this is a different task/topic. I would keep it as is for now, and improve later when more reference code is coming in.

jensens:
> I'd like someone more familiar with the jargon to review the rewritten "How behaviors work" section. This version is an improvement.

@davisagli ?

docs/backend/behaviors.md
stevepiercy: That still does not answer my question. How do you know the short names? Where do they come from? Is it in source code only? At the very least, give a link as reference for future PRs. Else I would say get rid of `short names` from the table altogether and refer only to what anyone can find here: https://6-classic.demo.plone.org/dexterity-types/Document/@@behaviors.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,320 @@ myst:
...
+    meta_type="Dexterity FTI"
+    name="Event"
+    xmlns:i18n="http://xml.zope.org/namespaces/i18n">
+  <property name="behaviors" purge="false">
+    <element value="myproject.subtitle" />
+  </property>
+</object>
+```
+
+After you apply the profile (or uninstall and install the custom add-on), the behavior is available on the Event content type.
```
ksuess:
```suggestion
After you apply the profile (or uninstall and install the custom add-on), the behavior is effective on the Event content type.
```

It is not only available but definitely applied.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,320 @@ myst:
...
+
+## Custom behaviors
+
+There are two types of behaviors:
+
+Schema-only behaviors
+: These behaviors have only a schema with fields.
+
+Full behaviors
+: A python class containing the logic of the behavior, an interface or schema defining the contract of the behavior, and a marker interface applied to the content type.
```
ksuess:
```suggestion
: A python class containing the logic of the behavior, an interface or schema defining the contract of the behavior, and a marker interface applicable to a content type.
```

I would say the behavior **can** be applied to a content type of the developers choice.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,320 @@ myst:
...
+      title="Price with net, VAT and gross"
+  />
+  ```
+
+After a restart of Plone, the behavior can be added to the content type in the {guilabel}`Content Types` control panel.
+The add and edit forms contain a new field `Price (net)`.
+
+This field is not displayed in most views.
+To display the data entered in this field, you need to modify the page template by adding the `price_net` field as `context.price_net`.
+To access the `price_vat` and `price_gross` fields, you need to get the adapter in your view class:
```
ksuess:
```suggestion
To access the `price_vat` and `price_gross` fields, you need to get the adapted object in your view class:
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,320 @@ myst:
...
+class SomeViewClass:
+
+    def vat(self):
+        adapter = IPriceBehavior(context)
+        return adapter.price_vat
+
+    def gross(self):
+        adapter = IPriceBehavior(context)
+        return adapter.price_gross
+```
```
ksuess:
```suggestion
```Python
from .price import IPriceBehavior

class SomeViewClass:

    def vat(self):
        adapted_context = IPriceBehavior(context)
        return adapted_context.price_vat

    def gross(self):
        adapted_context = IPriceBehavior(context)
        return adapted_context.price_gross
```
```

**Review by ksuess — CHANGES_REQUESTED** [review]

File: docs/backend/behaviors.md (hunk 49 above)
Comment (ksuess) [reply docs/backend/behaviors.md]: A list of behaviors with behavior names and title is very important and nowhere else to find. It is super helpful for a developer to find the implementation or just add it to a type of an add-on.

Apart from this documentation chapter, I think https://6-classic.demo.plone.org/dexterity-types/Document/@@behaviors should show or offer hover tip with behavior names.
Will do a PR.

File: docs/backend/behaviors.md (hunk 66 above)
Comment (jensens) [reply docs/backend/behaviors.md]: in fact its the adapter you need, the adapted object is the `context` of the adapter.

File: docs/backend/behaviors.md (hunk 67 above)
Comment (jensens) [reply docs/backend/behaviors.md]: OK, with this code i see what you want to express. but its confusing, at least to me.

File: docs/backend/behaviors.md (hunk 66 above)
Comment (jensens) [reply docs/backend/behaviors.md]: What about:
```suggestion
To access the `price_vat` and `price_gross` fields from a browser view, you need to get the adapter to the context of the view:
```
OT: this complexity is the reason why I once wrote a PLIP to add a dexterity/behavior API to `plone.dexterity`, but I never implemented it. Plan was to have an API like `context.behavior['myproject.price'].vat`.

ksuess: The adapter adapts the to be adapted object and returns the adapted object.

jensens: > The adapter adapts the to be adapted object and returns the adapted object.

brain-explosion

docs/backend/behaviors.md
stevepiercy: I agree! What I think you all are missing is that I do not know where these short names can be found outside of this table. Please share the secret locations of your knowledge with everyone!

docs/backend/behaviors.md
stevepiercy: Let's adapt the suggestion for English grammar.

```suggestion
To access the `price_vat` and `price_gross` fields from a browser view, you need to get the adapter from the context of the view:
```

docs/backend/behaviors.md
jensens: any other opinions?

stevepiercy: Please lowercase `Python` for the lexer.

I don't have an opinion about the code itself. I don't understand what you find confusing about @ksuess's suggestion.

In my brain (which is a scary place for anyone to dwell), using `adapter` implies that the thing can do adaptations, whereas `adapted_context` implies the context has been adapted in the thing, leaving it indeterminate whether it can do adaptations.

To be honest, this is all confusing, so maybe it is less/as/more confusing with the suggestion? I don't know.

jensens: OK, what about just `adapted`? In my brain, not much different from yours, the idea pops up I get a context, which is not the case.

docs/backend/behaviors.md
```diff
@@ -11,12 +11,320 @@ myst:
...
+  ```
+
+After a restart of Plone, the behavior can be added to the content type in the {guilabel}`Content Types` control panel.
+The add and edit forms contain a new field `Price (net)`.
+
+This field is not displayed in most views.
+To display the data entered in this field, you need to modify the page template by adding the `price_net` field as `context.price_net`.
+To access the `price_vat` and `price_gross` fields from a browser view, you need to get the adapter from the context of the view:
+
+```Python
```
jensens:
```suggestion
```python
```

docs/backend/behaviors.md
jensens: > Please lowercase Python for the lexer.

👍🏼

docs/backend/behaviors.md
jensens: They are in registered in ZCML and used in the FTIs of the types. To find the all names I used the browser inspect feature and looked at the values of the checkboxes of the types control panel. @ksuess improvement to show them on hover or in some other way is very valuable.

docs/backend/behaviors.md
stevepiercy: > OK, what about just `adapted`? In my brain, not much different from yours, the idea pops up I get a context, which is not the case.

Hm, I see your point. `adapted` might be part of the name of the thing. Maybe the answer to this question will yield the correct answer?

"When you pass a context to a behavior interface, what thing do you get in return?"

docs/backend/behaviors.md
stevepiercy: Oh, now I understand. Let's do the following.

1. Create a new issue in the correct repo with the following content. Sorry, I don't know the correct repo, else I would do that myself.

> When browsing a specific content type from the Content Type control panel, a list of behaviors is shown in the user interface with their Titles and Description. For each behavior, one may view the source code of the checkbox (its `name` attribute) to view their Short Names (is there a better name than "Short Names"?). For example, see the content type page's behaviors: https://6-classic.demo.plone.org/dexterity-types/Document/@@behaviors
>
> content-type-control-panel-page-behaviors
>
> See https://github.com/plone/documentation/pull/1372#discussion_r1050794700

2. Add a `todo` below the table in this PR.

````
```{todo}
For each behavior in the table above, one may view the source code of the checkbox (its `name` attribute) to view its Short Name.
An issue has been created to better expose these in the user interface. [Insert link to new issue]
```
````

3. Sort the table items by their `Title` to align with their order in the control panel.
4. Upon release of the new feature, delete the `todo`.

Comment (jensens) [reply docs/backend/behaviors.md]: If in doubt, then file the issue under Products.CMFPlone. And since the control panel is defined there (IIRC) it is the perfect place.

File: docs/backend/behaviors.md (hunk 67 above)
Comment (jensens) [reply docs/backend/behaviors.md]: so what happens there?

1. An abstract factory is called. (IPriceBehavior) Yes its an interface, but here in its shortcut as a factory usage, to do an getAdapter(context, IPriceBehavior). We could call latter with the exact same result. 
2. The factored adapter specific for the given content item is assigned to the variable. 

So its an adapter, like when you travel the world and you have a German Schuko plug on your notebooks power cable. I  carry a bag with adapters for all the different power outlets with me. Now, on my stop in Italy I find a Type L outlet. I say getAdapter by looking at the outlet (ITypeL) , what my plug needs (ISchuko) and then choose the right one. Then I plug the adapter in the wall. This is what the abstract factory does and return. Now I use the new interface. It provides (ISchuko) and plug in my Schuko plug to get power.

**Comment by ksuess** [comment]
Feel free to ignore for now my comment about naming and merge this precious new chapter.

I still do not aggree on naming the output of an adapter (or adapter factory) also adapter. Even if this is widely spread practice, in my opinion it is correct 
- to name the output of an adapter factory "adapted object" and 
- to name the input of an adapter factory "to be adapted object" or something similar.

Will do a change request later.

File: docs/backend/behaviors.md (hunk 49 above)
Comment (stevepiercy) [reply docs/backend/behaviors.md]: 1. Issue created https://github.com/plone/Products.CMFPlone/issues/3706

**Comment by stevepiercy** [comment]
I agree with @ksuess, provided items 2 and 3 in my [comment](https://github.com/plone/documentation/pull/1372#discussion_r1051658721) are fulfilled.

I'm also adapting to @jensens power adapter analogy to be a new introduction to the section `### Interfaces and adapters`. Suggestion coming in a moment.

File: docs/backend/behaviors.md [hunk 69]
```diff
@@ -11,12 +11,320 @@ myst:
...
+  />
+  ```
+
+After a restart of Plone, the behavior can be added to the content type in the {guilabel}`Content Types` control panel.
+The add and edit forms contain a new field `Price (net)`.
+
+This field is not displayed in most views.
+To display the data entered in this field, you need to modify the page template by adding the `price_net` field as `context.price_net`.
+To access the `price_vat` and `price_gross` fields from a browser view, you need to get the adapter from the context of the view:
+
```
stevepiercy, line 265:
```suggestion

(behavior-code-example)=

```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,320 @@ myst:
...
+```{note}
+Skip this section if you do not want to dive deeper into the internals of behaviors.
+You do not *need* to know this, but it may help if you run into problems.
+```
+
+In Plone, behaviors can be globally enabled on content types at runtime.
+With add-ons, behaviors can even be enabled even on a single content object or for a whole subtree in the content hierarchy.
+
+### Interfaces and adapters
+
```
stevepiercy, line 312:
```suggestion
To explain interfaces and adapters, let's begin with an analogy using electrical systems.

An electrical outlet provides an interface through which electricity passes.
When you travel to another country, you may need an outlet adapter for the outlet (the interface).
For example, assume you have a device that has plug for Schuko outlets, and in Italy there are Type L outlets.
If we were to represent the behavior of choosing the correct power adapter in Plone, you would do the following.

-   Call `getAdapter(context, ITypeL)` to determine the type of adapter needed for the electric outlet's interface.
-   Call `getAdapter(context, ISchuko) to determine the type of adapter that will accept your device's plug.
-   You choose the correct electrical adapter to use, and plug that into the wall outlet.
    This is similar to calling an abstract factory which returns the correct adapter to use.
    This interface provides an ISchuko adapter.
-   Finally to use the new interface, you plug in your Schuko plug into the adapter to get power to your device.

Similarly, using the {ref}`behavior code example <behavior-code-example>` above:

-   You would call an abstract factory with `getAdapter(context, IPriceBehavior)` to get an adapter.
    Although it is an interface, it is more of a shortcut to factory usage.
-   The adapter that is specific to the given content type is assigned to the variable `adapter`.
    Now you can use `adapter` for whatever you like.

```

stevepiercy (changes requested):
Add adapted @jensens adapter analogy to introduce adapters and interfaces, with a cross reference.

docs/backend/behaviors.md
stevepiercy: It is possible that I might have made some mistakes in terms, so please correct them as needed.

jensens: It is not correct in details, but great anyway, I will refine this. but at the moment I have a x-mas overload here. Please stay tuned.

jensens: I commit this and propose a new change, this is probably easier and more transparent than editing yours

docs/backend/behaviors.md
```diff
@@ -11,12 +11,340 @@ myst:
...
+```
+
+In Plone, behaviors can be globally enabled on content types at runtime.
+With add-ons, behaviors can even be enabled even on a single content object or for a whole subtree in the content hierarchy.
+
+### Interfaces and adapters
+To explain interfaces and adapters, let's begin with an analogy using electrical systems.
+
+An electrical outlet provides an interface through which electricity passes.
+When you travel to another country, you may need an outlet adapter for the outlet (the interface).
```
jensens, line 301:
```suggestion
When you travel to another country, you may need an outlet adapter for the specific outlet (the provided interface).
```

stevepiercy, line 316: Yes!

docs/backend/behaviors.md
```diff
@@ -295,6 +295,26 @@ In Plone, behaviors can be globally enabled on content types at runtime.
 With add-ons, behaviors can even be enabled even on a single content object or for a whole subtree in the content hierarchy.
 
 ### Interfaces and adapters
+To explain interfaces and adapters, let's begin with an analogy using electrical systems.
+
+An electrical outlet provides an interface through which electricity passes.
+When you travel to another country, you may need an outlet adapter for the outlet (the interface).
+For example, assume you have a device that has plug for Schuko outlets, and in Italy there are Type L outlets.
```
stevepiercy:
```suggestion
For example, assume you have a device that has a plug for Schuko outlets, and in Italy there are Type L outlets.
```

docs/backend/behaviors.md
```diff
@@ -11,12 +11,340 @@ myst:
...
+If we were to represent the behavior of choosing the correct power adapter in Plone, you would do the following.
+
+-   Call `getAdapter(context, ITypeL)` to determine the type of adapter needed for the electric outlet's interface.
+-   Call `getAdapter(context, ISchuko) to determine the type of adapter that will accept your device's plug.
+-   You choose the correct electrical adapter to use, and plug that into the wall outlet.
+    This is similar to calling an abstract factory which returns the correct adapter to use.
+    This interface provides an ISchuko adapter.
+-   Finally to use the new interface, you plug in your Schuko plug into the adapter to get power to your device.
```
jensens:
```suggestion
-  You need an electrical adapter for your Schuko plug.
  1. you look at the outlet and see it is TypeL.
  2. you look in your box containing different adapters and choose one the correct electrical adapter to use, and
  3. plug that into the wall outlet.
  4. Finally you can use your Schuko providing device on an Italian TypeL outlet.
-  In Python you call `getAdapter(context, ISchuko)` (context is here the outlet) which then 
   1. determine the type of interface provided by the `context` and as a result it finds `ITypeL`.
   2. looks in the component registry if there is a class that adapts to `ITypeL` and at the same time provides the requested `ISchuko`.
   3. initializes the adapter class with the context and return it as the result
   4. Finally the `ISchuko` providing adapter can be used on a `ITypeL` providing context.

This process of choosing the right adapter based on the information of the context and the requested interface implements the design pattern of an abstract factory.
```

docs/backend/behaviors.md
```diff
@@ -295,6 +295,26 @@ In Plone, behaviors can be globally enabled on content types at runtime.
...
 ### Interfaces and adapters
+To explain interfaces and adapters, let's begin with an analogy using electrical systems.
+
+An electrical outlet provides an interface through which electricity passes.
+When you travel to another country, you may need an outlet adapter for the outlet (the interface).
+For example, assume you have a device that has plug for Schuko outlets, and in Italy there are Type L outlets.
+If we were to represent the behavior of choosing the correct power adapter in Plone, you would do the following.
+
+-   Call `getAdapter(context, ITypeL)` to determine the type of adapter needed for the electric outlet's interface.
+-   Call `getAdapter(context, ISchuko) to determine the type of adapter that will accept your device's plug.
```
stevepiercy:
```suggestion
-   Call `getAdapter(context, ISchuko)` to determine the type of adapter that will accept your device's plug.
```

docs/backend/behaviors.md
```diff
@@ -295,6 +295,26 @@ In Plone, behaviors can be globally enabled on content types at runtime.
...
+-   You choose the correct electrical adapter to use, and plug that into the wall outlet.
+    This is similar to calling an abstract factory which returns the correct adapter to use.
+    This interface provides an ISchuko adapter.
```
stevepiercy:
```suggestion
    This interface provides an `ISchuko` adapter.
```

stevepiercy (changes requested):
Just a couple of minor MyST and English syntax and grammar corrections. I like your proposed change, too.

docs/backend/behaviors.md
stevepiercy: Careful of indentation of nested lists. Make sure they align with the first letter of the parent list item. This is one reason I always indent list items 4 spaces.

Also you might need to add a blank line between a nested list. In reST this was strictly enforced, but in the last year or so and with MyST, I am not sure. I just do it by habit now. Check the Netlify build to make sure it looks OK.

```
- Parent List Item

    #.  Don't use actual numbers because they often need renumbering in the future.
    #.  Instead use `#.`.
    #.  Plus when you hit 10 or more enumerated items, you still indent the right number of spaces.

- Parent List Item
```

stevepiercy: Thank you for the revision. This is a lot better.

docs/backend/behaviors.md
jensens: Question here is, what does the call return? `IPriceBehavior(context)` which is similar to  `getAdapter(context, IPriceBehavior)`. It returns an adapter bound to the context. So if we stick to Zope component architectures naming (which IMO is a good idea) this is the actual instance of the "adapter". It does not return something adapted, which implies we still have this something there, but somehow enhanced, like wrapped. But we do not have a wrapper here (also a design pattern) but an "object adapter pattern". So we get something new, nothing polymorphic.
`adapted_context` sounds to my brain like its context, but provides something _addtional_ - which is not the case. It only provides the provided interface, nothing else.
We could call it:
- as is `adapter` - because that's what its is in ZCA
- more specific `behavior_adapter` - which indicates its the behavior specific adapter we get
- `price_for_context` the semantic way to tell what we get (I like this one and would use it in my code) or 
- more general `price_adapter`.

stevepiercy: > `price_for_context` the semantic way to tell what we get (I like this one and would use it in my code) or

We have a winner! 🔔

stevepiercy: Done in 00e56f3db0ae85ebe1c4433510787c5545c1a272

stevepiercy (approved):
I completed my review and I'm satisfied with the many revisions. Thank you to everyone for your contributions and review, especially @jensens. I'm going to merge this PR.

If you notice any mistakes, please do open a new issue. Thank you!

PR #1384, opened 2022-12-16
Switch PLONE_BACKEND_VERSION to 6.0.0 (was 6.0.0b3)

ksuess:


ksuess:
@ericof, this affects the chapters "installation with docker". Would you please be so kind and check if this is now OK and up to date. I think so.

stevepiercy:
It effects these files: https://github.com/plone/documentation/search?q=PLONE_BACKEND_VERSION

stevepiercy (approved):

docs/conf.py
```diff
@@ -300,7 +300,7 @@ def source_replace(app, docname, source):
 
 # Dict of replacements.
 source_replacements = {
-    "{PLONE_BACKEND_VERSION}": "6.0.0b3",
+    "{PLONE_BACKEND_VERSION}": "6.0.0",
```
ericof, line 303: From now on, we could always recommend `6.0` for the Backend version

ksuess: @stevepiercy: @ericof 's comment is about that he managed to set up Plone image in a way that the use can be done without mentioning the PATCH version.
As you merged this already, I'll open a new PR.

PR #1389, opened 2022-12-19
`docker run` Plone 6 without mentioning the PATCH version

ksuess:


ksuess:
I expect Plone 6.0.1 later this day. So hold on for now.

docs/install/containers/images/backend.md
```diff
@@ -39,13 +39,13 @@ We encourage users of the `Plone` images to familiarize themselves with the opti
...
-| `VERBOSE_SECURITY` | `verbose-security` | `false` |
-| `DEFAULT_ZPUBLISHER_ENCODING` | `default-zpublisher-encoding` | `utf-8` |
-| `LISTEN_PORT` | (no equivalent) | `8080` |
+| Environment variable             | Zope option                      | Default value |
+| -------------------------------- | -------------------------------- | ------------- |
+| `DEBUG_MODE`                     | `debug-mode`                     | `off`         |
+| `SECURITY_POLICY_IMPLEMENTATION` | `security-policy-implementation` | `C`           |
+| `VERBOSE_SECURITY`               | `verbose-security`               | `false`       |
+| `DEFAULT_ZPUBLISHER_ENCODING`    | `default-zpublisher-encoding`    | `utf-8`       |
+| `LISTEN_PORT`                    | (no equivalent)                  | `8080`        |
```
stevepiercy: Please configure your editor to not reformat tables.

docs/conf.py
```diff
@@ -300,7 +300,8 @@ def source_replace(app, docname, source):
 
 # Dict of replacements.
 source_replacements = {
-    "{PLONE_BACKEND_VERSION}": "6.0.0",
+    "{PLONE_BACKEND_MINOR_VERSION}": "6.0",
+    "{PLONE_BACKEND_VERSION}": "6.0.1",
```
stevepiercy: I don't see the point in renaming the variable. Let's keep it short.

```suggestion
    "{PLONE_BACKEND_VERSION}": "6.0",
```

docs/conf.py
```diff
@@ -300,7 +300,8 @@ def source_replace(app, docname, source):
 
 # Dict of replacements.
 source_replacements = {
-    "{PLONE_BACKEND_VERSION}": "6.0.0",
+    "{PLONE_BACKEND_MINOR_VERSION}": "6.0",
```
stevepiercy, line 298: Why do we need two different versions?

```suggestion
```

stevepiercy (changes requested):

docs/conf.py
ksuess: Buildout wants to know the PATCH version. 
Docker can handle MINOR version.

See `Semantic Versioning`https://semver.org/lang/de/

docs/conf.py
ksuess: Buildout needs the PATCH version. See comment above.

docs/install/containers/images/backend.md
ksuess: Here is a before and after. 
I personally prefer the (editor extension) reformatted version.


Item-0_and_AppleTimeMachineExtra_and_Item-0_and_Item-0
 
Item-0_and_AppleTimeMachineExtra_and_Item-0_and_Item-0

stevepiercy: Please not in this PR. It just adds noise, and we need to discuss whether to reformat all tables in the Docs Team. I find it easier to read diffs when they have shorter lines, and the reformat makes it exceedingly difficult to do that.

docs/conf.py
stevepiercy: OK, in that case, let's be explicit with the name.

```suggestion
    "{PLONE_BACKEND_PATCH_VERSION}": "6.0.1",
```

stevepiercy: However, where do we document the use of buildout? I thought [we made the decision not to do that in Plone 6](https://6.docs.plone.org/install/install-from-packages.html#install-plone-6)?

ksuess: No buildout. I should not have mentioned it.
It's the installation with pip that uses the constraints of the selected Plone version. The url is https://dist.plone.org/release/{PLONE_BACKEND_VERSION}/constraints.txt with the complete PATCH version

stevepiercy: I think this is the only place that needs that distinction:

https://github.com/plone/documentation/blame/6-dev/docs/install/manage-add-ons-packages.md#L64

davisagli: https://dist.plone.org/release/6.0-latest/constraints.txt is available. But, in the context of a project that's going to be maintained over time the best practice is to pin to a specific version, so that your CI doesn't break until you've submitted a PR to explicitly update to the new patch (so that if it breaks, you know why)

ksuess: > https://dist.plone.org/release/6.0-latest/constraints.txt is available. But, in the context of a project that's going to be maintained over time the best practice is to pin to a specific version, so that your CI doesn't break until you've submitted a PR to explicitly update to the new patch (so that if it breaks, you know why)

I'm not sure if I understand. The documentation recommends to pin the version.

davisagli: @ksuess Sorry, I also ended up recommending to pin the version, but I was not very clear.

stevepiercy (approved):
Looks good here and in preview. Thank you!

PR #1391, opened 2022-12-19
Attempt to run Vale in GitHub Actions

stevepiercy:


.github/workflows/test.yml
```diff
@@ -25,4 +25,6 @@ jobs:
       - name: Run tests with make test
         run: |
           echo "VALEFILES=$(find -L ./docs/ -type d \( -path ./docs/plone.restapi/lib/* -o -path "./docs/plone.restapi/performance/*" \) -prune -false -o -type f -name "*.md" -print)" >> $GITHUB_ENV
+          echo "${{ env.VALEFILES }}"
+          sudo vale sync
           make vale ${{ env.VALEFILES }}
```
davisagli: @stevepiercy maybe:
```
run: |
  VALEFILES=$(find -L ./docs/ -type d \( -path ./docs/plone.restapi/lib/* -o -path "./docs/plone.restapi/performance/*" \) -prune -false -o -type f -name "*.md" -print) make vale
```

(Setting GITHUB_ENV is only needed if you need it to be persisted for subsequent GHA steps. And I suspect using `env` in an expression like that only includes the environment variables that were set when the step started, but I'm not certain.)

stevepiercy: Fixed in 2202388bae30fdd77846e92a07c2c13809faba35. Thank you!

stevepiercy:
All right! I got it working finally. Vale and Microsoft styles are properly installed now, and can run.

The final step is clean up and configuration options for the test run. I think that we should show the full output, but allow a failure, similar to what we used to do for spellcheck. Thoughts @ksuess @polyester @spereverde ?

ksuess (dismissed):
I trust that you know what you are doing.
Just wondering, if `make vale` (which is in fact a `vale sync`) is already doing what `git clone  https://github.com/errata-ai/Microsoft.git; cp -r ./Microsoft/Microsoft ./styles`does?

stevepiercy:
> I trust that you know what you are doing.

I don't. I'm making it up as I go along. :grin:

> Just wondering, if `make vale` (which is in fact a `vale sync`) is already doing what `git clone https://github.com/errata-ai/Microsoft.git; cp -r ./Microsoft/Microsoft ./styles`does?

Here's a couple of tests where I removed the manual installation of Microsoft styles, and attempted to install them through `vale sync`:

https://github.com/plone/documentation/actions/runs/3747488211/jobs/6363764596#step:5:898

https://github.com/plone/documentation/actions/runs/3747543946/jobs/6363881090#step:5:895

I was not successful to get `vale sync` to work in GitHub Actions. Maybe I'm doing it wrong, but nothing obvious jumped out at me. I checked the vale issue tracker for this error message, and it means that vale cannot find its `.vale.ini`, but if that was true, then vale would not run at all. `¯\_(ツ)_/¯`

ksuess:
> > I trust that you know what you are doing.
> 
> I don't. I'm making it up as a go along. 😁
> 
> > Just wondering, if `make vale` (which is in fact a `vale sync`) is already doing what `git clone https://github.com/errata-ai/Microsoft.git; cp -r ./Microsoft/Microsoft ./styles`does?
> 
> Here's a couple of tests where I removed the manual installation of Microsoft styles, and attempted to install them through `vale sync`:
> 
> https://github.com/plone/documentation/actions/runs/3747488211/jobs/6363764596#step:5:898
> 
> https://github.com/plone/documentation/actions/runs/3747543946/jobs/6363881090#step:5:895
> 
> I was not successful to get `vale sync` to work in GitHub Actions. Maybe I'm doing it wrong, but nothing obvious jumped out at me. I checked the vale issue tracker for this error message, and it means that vale cannot find its `.vale.ini`, but if that was true, then vale would not run at all. ¯_(ツ)_/¯

Sorry for interfering. It's fine if it's doing what you expect. Our time is precious…

ksuess:
Ready or still draft?
I would approve.

stevepiercy:
As far as allowing this run to pass, we want to see the output of vale, but not have it fail. I just pushed another commit that handles that. It passes now.

https://github.com/plone/documentation/actions/runs/3747802118/jobs/6364422404

However I had to temporarily remove the `make test` step because that currently fails, due to the redirects that were introduced in Volto's changelog. I'm working on that next.

stevepiercy:
Hmm.. linkcheckbroken should cause this to fail, but it does not. Oh, I see. https://github.com/plone/documentation/issues/1396. I will fix that in a separate PR. This PR is now final.

davisagli (approved):

ksuess (approved):
LGTM

Issue #1395, opened 2022-12-21
Deprecation notices from lack of `pyproject.toml` and `wheel`


stevepiercy:
### Plone Documentation versions affected

- [ ] 5.x
- [X] 6.x

### Description

We have a couple of deprecation notices that appear on CI.

> DEPRECATION: livereload is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559
> 
> DEPRECATION: sphinx-sitemap is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559

We could add pins for `setuptools` and `wheel` to our `requirements.txt`, and add a `pyproject.toml` file. I don't know what we would put in it, though.

Anyone have any opinions? @plone/documentation-team

mauritsvanrees:
That sounds like these two projects have no `pyproject.toml`, which is something we cannot do anything about, except perhaps file an issue.

The `wheel` package is not installed. I usually install it. Plone has it in its [`requirements.txt`](https://dist.plone.org/release/6.0.0/requirements.txt).

But even if we install `wheel`, it will find no wheels for these two packages: their latest version only has a source distribution. So in the end it would call `python setup.py install` anyway.

 I guess the problem then is that pip would very much like to avoid calling `setup.py install`. I would need to reread the linked issue and related documentation a few more times to understand the big picture, and also see where the `pep517` and `build` packages come into this.

Summary seems to be: only the authors of those two packages can do anything about this, mostly be uploading wheels to PyPI.

stevepiercy:
Refs:
- https://github.com/lepture/python-livereload/issues/240
- https://github.com/jdillard/sphinx-sitemap/pull/39

stevepiercy:
A maintainer of livereload uploaded a wheel. I'm going to add `wheel` to our requirements.

That leaves sphinx-sitemap as the remaining todo.

stevepiercy:
And the maintainer of sphinx-sitemap released a wheel to PyPI.

stevepiercy:
[...]

PR #1398, opened 2022-12-21
Add wheel to requirements

stevepiercy:
See #1395

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -18,7 +18,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
+          pip install -q -r requirements-initial.txt
           pip install -q -r requirements.txt
           pip freeze
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
+          pip install -q -r requirements-initial.txt
           pip install -q -r requirements.txt
 
       - name: Run tests with make test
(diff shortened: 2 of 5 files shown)


mauritsvanrees (changes requested):
Can you put `wheel` at the top of the list? Otherwise I fear it will not be available to pip until it is too late. I must say I am not 100 percent sure putting it at the top actually helps, or that it only helps for a completely new invocation of pip.

stevepiercy:
From https://pip.pypa.io/en/stable/cli/pip_install/#installation-order

> As of v6.1.0, pip installs dependencies before their dependents, i.e. in “topological order.” This is the only commitment pip currently makes related to order. While it may be coincidentally true that pip will install things in the order of the install arguments or in the order of the items in a requirements file, this is not a promise.

Thus if wheel is a dependency of any of the other requirements (it most likely is not) then wheel will get installed first.

Nonetheless putting it first is not a guarantee that it will be installed first. One way to guarantee order is to iterate over the requirements file, and call `pip install` on each line. But that feels like overkill to me.

Is there some other way to guarantee that wheel gets installed before the packages that might not have a wheel?

mauritsvanrees:
Have an `initial-requirements.txt` file (or similar) with these (plus pins) to get the "packaging build system" up to full strength:

```
pip
setuptools
wheel
```

stevepiercy:
I assume that I should also use two `pip install`s, one for each requirements files, to ensure that the requirements are installed in order of each file. If not, please let me know.

stevepiercy:
> Have an `initial-requirements.txt` file (or similar) with these (plus pins) to get the "packaging build system" up to full strength:
> 
> ```
> pip
> setuptools
> wheel
> ```

Done in https://github.com/plone/documentation/pull/1398/commits/65e969963174e3af56a61bc2bad074f47613babb

stevepiercy:
Could I get an approval from someone?

davisagli (approved):

Issue #1415, opened 2023-01-02
Glossary needs cleanup


jensens:
### Plone Documentation versions affected

- [ ] 5.x
- [X] 6.x

### Description

In the glossary there are some terms like.

```
Hydration (SSR)
    After loading an HTML page generated with SSR in the browser, React can "populate" the existing DOM elements, recreate and attach their coresponding components.

```

Someone looking by alphabet for SSR won't not find it.
It would be better to define it as two terms:

```
SSR
Hydration
    After loading an HTML page generated with SSR in the browser, React can "populate" the existing DOM elements, recreate and attach their coresponding components.
```

stevepiercy:
SSR != Hydration. They should be two separate Glossary entries, and wherever:

```
{term}`Hydration (SSR)`
```

appears, it should be changed to:

```
{term}`Hydration`
```

Also this term:

```
Server-Side Rendering (SSR)
```

Should be:

```
SSR
Server-Side Rendering
```

You can "stack" terms and define them.

stevepiercy:
Similarly:

```
Single Page Application (SPA)
```

Should be:

```
SPA
Single Page Application
```

and

```
Hot Module Replacement (HMR)
```

should be:

```
HMR
Hot Module Replacement
```


and

```
middleware (Redux)
```

s/b

```
middleware
```


and

```
hooks (React)
```

s/b

```
hooks
```


and

```
hoisting (Yarn)
```

s/b

```
hoisting
```

We also need to check whether any `term` includes the parenthetical elsewhere in the docs, specifically in Volto.

stevepiercy:
PRs are ready:

- https://github.com/plone/documentation/pull/1416
- https://github.com/plone/volto/pull/4211

stevepiercy:
Completed and updated submodule.

PR #1416, opened 2023-01-02
Glossary cleanup

stevepiercy:
Refs: https://github.com/plone/documentation/issues/1415 and https://github.com/plone/volto/pull/4211

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -154,11 +154,17 @@ record
     A Rapido app is able to store data as records.
     Records are basic dictionaries.
 
-Project (Volto)
-    The product of running the package `@plone/generator-volto`, resulting in a customizable instance of Volto.
+Add-on
+    An add-on in Plone extends its functionality.
+    It is code that is released as a package to make it easier to install.
 
-Add-on (Volto)
-    A JavaScript package that integrates with Volto's configuration registry and is able to enhance, extend, and customize it.
+    In Volto, an add-on is a JavaScript package.
+
+    In Plone core, an add-on is a Python package.
+
+    -   [Plone core add-ons](https://github.com/collective/awesome-plone#readme)
+    -   [Volto add-ons](https://github.com/collective/awesome-volto#readme)
+    -   [Add-ons tagged with the trove classifier `Framework :: Plone` on PyPI](https://pypi.org/search/?c=Framework+%3A%3A+Plone)
 
 Add-on configuration loader (Volto)
     A function with signature `config => config`.
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -168,15 +168,19 @@ Add-on
 
 Volto configuration loader
     A function with signature `config => config`.
-    It gets the Volto Configuration registry, and it must return it back after mutating it.
-    It is similar to `GenericSetup` profiles in Plone Backend.
+    It gets the Volto configuration registry, and it must return it back after mutating it.
+    It is similar to `GenericSetup` profiles in the Plone backend.
     An add-on must provide a default configuration loader that is always loaded when Volto runs.
     An add-on can have multiple configuration loaders, and they can be loaded optionally from the Volto configuration.
 
-Configuration registry (Volto)
-    A singleton object modeled using JavaScript modules.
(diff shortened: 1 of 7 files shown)


docs/glossary.md
```diff
@@ -194,14 +212,21 @@ Express
...
 
-Server-Side Rendering (SSR)
-    When first loading any Plone page, users will get HTML markup that closely matches the final DOM structure of the React components used to render that page.
+SSR
+server-side rendering
+    When first loading any Plone page, SSR will send HTML markup that closely matches the final DOM structure of the React components used to render that page.
+
+    More generally, when a client sends a request to an SSR page, the HTML for the page is created on the server on each request.
+    SSR enables a developer to customize a website per request and per user.
+    In addition, SSR can improve performance and search engine optimization (SEO) for a website.
```
jensens: I would write it like so:
```suggestion
    When an HTTP client (web browser) sends a request, the HTML markup for the page is created on the server.
    
    When loading a page in  Volto the first time, SSR will send HTML markup that closely matches the final DOM structure of the React components used to render that page. 
    After initial loading the {term}`hydration` mechanism is taking over and then rendering happens browser side.
    
    When loading a page in ClassicUI the full page is loaded on every request (rarely HTML-snippets are loaded, i.e. for overlays).

    SSR enables a developer to customize a website per request and per user.
    In addition, SSR can improve performance and search engine optimization (SEO) for a website.
```

docs/glossary.md
```diff
@@ -194,14 +212,21 @@ Express
...
 
-Single Page Application (SPA)
+SPA
+single page application
     A type of JavaScript application that aims to provide a better user experience by avoiding unnecessary reloading of the browser page, instead using AJAX to load backend information.
```
jensens: Should we define `AJAX` too? https://en.wikipedia.org/wiki/Ajax_(programming)

docs/glossary.md
```diff
@@ -235,16 +260,17 @@ JSX
...
-    They allow customizing the behavior of the data flow inside the redux store.
+Redux
+Redux middleware
+    Custom wrappers for the [Redux](https://redux.js.org/) store dispatch methods.
+    They allow customizing the behavior of the data flow inside the Redux store.
 
-hooks (React)
-    Hooks are a React API that allow function components to use React features such as lifecycle methods, states, and so on.
+hooks
+    [Hooks](https://reactjs.org/docs/hooks-overview.html) are a React API that allow function components to use React features, such as lifecycle methods, states, and so on.
```
jensens: There are hooks in Python too. Should be like 
- What are hooks for in prgramming
- Hooks in React
- Hooks in Python

jensens (changes requested):

docs/glossary.md
jensens: Also, DOM should be defined as term then.

stevepiercy: Rewritten in [5ace25b](https://github.com/plone/documentation/pull/1416/commits/5ace25b34ae9dd83453632a092a971d38098c701). Please let me know if this hits the mark.

docs/glossary.md
stevepiercy: Done in [f466123](https://github.com/plone/documentation/pull/1416/commits/f4661232f14259df434967d0fc224cec98f2104a)

docs/glossary.md
stevepiercy: Rewritten in [3a96dbb](https://github.com/plone/documentation/pull/1416/commits/3a96dbbc9a2ac4d7d0f7498f76dfa3c7cab20a5d)

stevepiercy: I forgot to mention, there are no specific Python hooks discussed in the docs, just general programming hooks and React hooks.

docs/glossary.md
stevepiercy: > Also, DOM should be defined as term then.

Done in [c538e9c](https://github.com/plone/documentation/pull/1416/commits/c538e9cde2e2b7001b5f433ed3d63671120a7ff2)

jensens (approved):

PR #1417, opened 2023-01-05
'a' tag should have text contained

jensens:
Improve example.


stevepiercy (approved):
Thanks for the quick fix!

PR #1405, opened 2022-12-24
Add a shortcut that brings the cursor to the search input field

MAX-786:
### Plone Documentation version affected
- [x] 6.x

## Feature
When the `ctrl+k` shortcut is pressed, the cursor is automatically placed in the search input field.

## Additional Changes
I am considering adding the `^k` to the right side of the search input field OR changing the placeholder to `use "ctrl+k" to set focus`. Before I implement this feature, I would like to get feedback from others. Please let me know your thoughts.

Issue #1184, opened 2022-02-18
Search and find Plone. Find Plone docs and more: packages documentation, videos, podcasts included.


ksuess:
Plone Documentation versions affected
6.x

Include more than plone/documentation:

- packages with elaborated documentation
   - plone.api (signature and narration)
   - maybe also plone.restapi (already a full featured documentation site at https://plonerestapi.readthedocs.io)
   - more
- Videos of Plone foundation channel on Vimeo (oops, yt)
- Podcasts
- Blog posts (some of us are still blogging tips and tricks!)
- Tweats with tips and tricks
- more tips and tricks

https://nuclia.com/developers/
OSS AI search with Nuclia, powered by Guillotina.
https://www.youtube.com/watch?v=WDYcchE1vqc

I am not affiliated and do not get a cent. Just curious.

stevepiercy:
This is compelling. I think this is bigger than Documentation only, and should be proposed to the plone.org update team. I have done so in https://discord.com/channels/786421998426521600/903929948228681758/945070834798690396

tkimnguyen:
This is very interesting, but a big project of its own, and I can tell you the admin team is stretched pretty thin as it is that we won't be jumping at the chance to run more infrastructure.

In Discord, @stevepiercy said "eat your own dog food" so I initially thought he meant we should index all those things so that Plone's built-in search would do this.

tkimnguyen:
Could this be a GSoC project?

MAX-786:
> Could this be a GSoC project?

@tkimnguyen, I didn't see this on last year's project lists, so if this is available and still considered I would love to look into it and have more description about it. :) Thanks!

stevepiercy:
plone.org uses Nuclia search now.

React implementation:
https://github.com/plone/plone.org/search?q=nuclia

Demo:
https://plone.org/search?SearchableText=piercy

We might want to jump into Discord https://discord.com/channels/786421998426521600/903929948228681758 and chat about it.

PR #1421, opened 2023-01-09
Addition to "Add a shortcut that brings the cursor to the search input field #1405 

ksuess:
Addition to #1405

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -177,23 +177,21 @@ div.admonition.admonition-todo>.admonition-title::before {
 
 /* Search */
 
-#search-shortcut-page {
-  position: absolute;
-  right: 156px;
-  top: 0.375rem;
-  border-radius: 0.25em;
-  background: #d9eef9;
-  padding: 0.1787rem 0.2813rem;
-  color: #4a4a4a;
-  font-weight: 600;
-  line-height: 1.25rem;
-}
-
 /* Show search form. It is hidden by default. */
 #search-documentation,
 #search-documentation~form,
 #search-documentation~p {
   display: block;
+}
+
+#shortcut.input-group-text {
+  padding-top: 0;
+  padding-bottom: 0;
+}
+
+.form-control:focus {
+  box-shadow: none;
+  border-width: 2px;
 }
 
 ul.search {
@@ -244,6 +242,18 @@ ul.search li {
 
 ul.search li p.context {
   margin-left: 0;
+}
+
+/* Search form sidebar */
+
+.bd-search {
+  font-size: .8rem;
+}
+
+.bd-search input,
+.bd-search .input-group-text {
+  font-size: .8rem;
+  padding-left: .5em;
 }
 
 .pathseparator {
@@ -317,6 +327,7 @@ div.sidebar:not(.margin) .admonition {
   margin-right: .5em;
   background-color: #ffffff;
 }
+
 #search-shortcut-sidebar {
   position: absolute;
   right: 1.25rem;
--- a/docs/_static/search_shortcut.js
+++ b/docs/_static/search_shortcut.js
@@ -3,36 +3,33 @@
  */
 
 $(document).ready(() => {
-    $(document).keydown(function(event) {
-        if ((event.ctrlKey || event.metaKey) && event.key == "k") {
-            event.preventDefault();
-            $('#search-input').focus();
-        }
-    });
-    if (navigator.platform.indexOf('Mac') === -1) {
-        $('#search-shortcut-sidebar').html("^K"); // if OS isn't Mac change Visual indication of search field
-    }
-    
-    // Show & hide seach input when focusin & focusout
-    $('#search-input').focus(() => {
-        $('#search-shortcut-sidebar').hide();
-    });
-    $('#search-input').focusout(() => {
-        $('#search-shortcut-sidebar').show();
-    });
-
-    // For Search page
-            
     if (window.location.pathname === '/search.html') {
         $('form.bd-search').css('visibility','hidden'); // Hide Sidebar Search field
+
+        $(document).keydown(function(event) {
+            if ((event.ctrlKey || event.metaKey) && event.key == "k") {
+                event.preventDefault();
+                $('#q').focus();
+            }
+        });
+    } else {
+        $(document).keydown(function(event) {
+            if ((event.ctrlKey || event.metaKey) && event.key == "k") {
+                event.preventDefault();
+                $('#search-input').focus();
+            }
+        });
     }
+
     if (navigator.platform.indexOf('Mac') === -1) {
-        $('#search-shortcut-page').html("^K"); // if OS isn't Mac change Visual indication of search field
+        $('#search-shortcut').html("^"); // if OS isn't Mac change visual indication of search field
     }
-    $(document).keydown(function(event) {
-        if ((event.ctrlKey || event.metaKey) && event.key == "k") {
-          event.preventDefault();
-          $('#q').focus();
-        }
-    });
+
+    // // Toggle search shortcut.
+    // $('#search-input').focus(() => {
+    //     $('#search-shortcut-sidebar').hide();
+    // });
+    // $('#search-input').focusout(() => {
+    //     $('#search-shortcut-sidebar').show();
+    // });
 });
--- a/docs/_templates/search-field.html
+++ b/docs/_templates/search-field.html
@@ -1,5 +1,18 @@
 <form class="bd-search d-flex align-items-center" action="{{ pathto('search') }}" method="get">
-    <i class="icon fas fa-search"></i>
-    <input type="search" class="form-control" name="q" id="search-input" placeholder="{{ _(theme_search_bar_text) }}" aria-label="{{ theme_search_bar_text }}" autocomplete="off" >
-    <span id="search-shortcut-sidebar">⌘K</span>
-  </form>
+  <div class="input-group">
+    <label for="q" class="sr-only">Search string
+    </label>
+    <div class="input-group-prepend">
+      <span class="input-group-text" id="shortcut">
+        <kbd>⌘</kbd>&nbsp;<kbd>K</kbd>
+      </span>
+    </div>
+    <input
+      type="search" name="q" id="search-input"
+      class="form-control"
+      placeholder="{{ _(theme_search_bar_text) }}"
+      aria-labelledby="{{ _(theme_search_bar_text) }}"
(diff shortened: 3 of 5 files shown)


MAX-786:
@ksuess, this design gives the Search field a 'UNIQUE to Plone' design, so I can '+' vote for this design. I really appreciate your expertise and experience, and I have also been thinking about this design problem, and I just want to share my opinion and see if it might be worth considering as we continue to discuss the design:

1. Windows users who never used macOS sometimes take `⌘` for `Windows key` and IMHO I prefer changing indication to `^` from `⌘` according to the user's OS rather than hardcoding it (adding id='search-shortcut' to `<kbd>` will do ).
Screenshot_20230110_002733

2.  I think it would be a good UX if we have this 'shortcut hint' on the right since it would be more easily noticeable.
Screenshot_20230110_015645

Feedback welcome. Thanks 😊.

ksuess:
>     1. Windows users who never used macOS sometimes take `⌘` for `Windows key` and IMHO I prefer changing indication to `^` from `⌘` according to the user's OS rather than hardcoding it (adding id='search-shortcut' to `<kbd>` will do ).
>        Screenshot_20230110_002733
> 
Yes, of course. You already prepared this distinction for OS.
I tried a short while to find a JINJA only solution for search.html and search_field.html. But it's not a low hanging fruit.

>     2. I think it would be a good UX if we have this 'shortcut hint' on the right since it would be more easily noticeable.
>        Screenshot_20230110_015645
> 
Yes, I am fine with the short cut on the right side. 
Reason for moving it to the left was the search page, where there is already a 'search' and 'reset' on the right side.
And then having the short cut on the same side for both: sidebar and search page.

--

A nice to have would be the elimination of the 'reset' by a little cross appearing for a filled search field. ;-)

Another nice to have would be to find a way that on ^k the search field is not glued to the top but with a small margin to the top.

ksuess:
BTW 
grafik
is not unique to Plone but the kbd style throughout the documentation, either from Sphinx-book-theme or pydata-theme or Sphinx itself. @stevepiercy knows best.

stevepiercy:
Before getting into the details of this implementation, I would like to discuss whether we should follow the search implementation of the grandparent theme, [`pydata-sphinx-theme`](https://pydata-sphinx-theme.readthedocs.io/en/stable/). I am more concerned about functionality—the features we implement—than their appearance at this point.

Advantages compared to this implementation:
- Modal is brought up with keyboard shortcut or clicking Search icon.
- No sidebar search input means no blank space in the sidebar on the Search results page.
- Clicking the search icon in the top navigation on the Search results page merely focuses the existing search input.

Disadvantages compared to this implementation:
- It is not clear that the Search input can be displayed with a keyboard shortcut until it is actually displayed.
- Even though the input is of `type="search"`, there is no delete icon to reset the input. I was unable to determine why this is so.
- The keyboard shortcut persists even when entering text. I would like it to go away so that the user does not attempt to use the keyboard shortcut to submit the search query.
- No submit button. It is not clear that the user can hit the `Enter` key to search, but maybe people just know that now?

ksuess:
> Before getting into the details of this implementation, I would like to discuss whether we should follow the search implementation of the grandparent theme, [`pydata-sphinx-theme`](https://pydata-sphinx-theme.readthedocs.io/en/stable/). I am more concerned about functionality—the features we implement—than their appearance at this point.
> 
> Advantages compared to this implementation:
> 
>     * Modal is brought up with keyboard shortcut or clicking Search icon.
> 
>     * No sidebar search input means no blank space in the sidebar on the Search results page.
> 
>     * Clicking the search icon in the top navigation on the Search results page merely focuses the existing search input.
> 
> 
> Disadvantages compared to this implementation:
> 
>     * It is not clear that the Search input can be displayed with a keyboard shortcut until it is actually displayed.
> 
>     * Even though the input is of `type="search"`, there is no delete icon to reset the input. I was unable to determine why this is so.
> 
>     * The keyboard shortcut persists even when entering text. I would like it to go away so that the user does not attempt to use the keyboard shortcut to submit the search query.
> 
>     * No submit button. It is not clear that the user can hit the `Enter` key to search, but maybe people just know that now?

A modal would be in my opinion the next step. It would be the next step if the modal would include all features of the search form, especially the results. IMHO a search modal that leads to a classic search form instead of showing results (while typing) is not a plus to the current way of searching in Plone 6 documentation. Pydata theme does forward the user to a classic search page.
My current favorite for documentation search is the one of SvelteKit: https://kit.svelte.dev/
But the SveltKit search, even if someone would do the effort to implement it, is not 1:1 enough for Plone documentation, because the facet (filter by 'frontend', 'backend', etc) is in my eyes important. It is important as we will have a huge amount of pages / search results while the docu is growing to a size of the former Plone 5 documentation.

IMHO the search as it is now is OK and with the enhancement by the shortcut ^k it is even better.

What could be done for cleaning the situation of having two search fields while the search page is open:
Replace the search field in the left sidebar by a link to the search page.

MAX-786:
> A nice to have would be the elimination of the 'reset' by a little cross appearing for a filled search field. ;-)

Yep, I was thinking the same :)

> Another nice to have would be to find a way that on ^k the search field is not glued to the top but with a small margin to the top.

I didn't quite understand ... looks fine to me.

> is not unique to Plone but the kbd style throughout the documentation, either from Sphinx-book-theme or pydata-theme or Sphinx itself. @stevepiercy knows best.

Oh got it,

MAX-786:
> But the SveltKit search, even if someone would do the effort to implement it, is not 1:1 enough for Plone documentation, because the facet (filter by 'frontend', 'backend', etc) is in my eyes important. It is important as we will have a huge amount of pages / search results while the docu is growing to a size of the former Plone 5 documentation.
> 
> IMHO the search as it is now is OK and with the enhancement by the shortcut ^k it is even better.

@ksuess, yes IG for now we can go with this enhancement only, but in the future, if we are willing to implement Search Modal  we can go with the svelte 1:1 and I also have a solution to implement the facet, it may require a lot of work and attention 😖 but this is what I propose:

Instead of radio options to filter we can have a select option on the Search modal inside the search bar on the right side, and results can be shown just like svelte does.  

Screenshot_20230110_143241

As I said this requires a lot of work and may not be compatible with sphinx but it will modernize the Plone 6 docs if this can be done :).
@stevepiercy, love to hear your thoughts on this and this may overcome the disadvantages you stated. Thanks!

stevepiercy:
Svelte's is nice, but I much prefer a page with full results instead of just 4 results. I like to open links in new tabs from the search results with adequate descriptions to give more context.

I mistakenly omitted from the Advantages that ours allows narrowing searches by topic/category.

Although I would like to discuss other search interfaces and results, for the sake of improving what we have now, I will defer that discussion to when we upgrade our theme to the latest from sphinx-book-theme (0.4.0rc1 is out now for testing). That will introduce a dark theme, which has its own design issues, including `kbd` styles that will affect search. Also we, and now the Plone Foundation, have been talking about using Nuclia for search on plone.org and docs: https://github.com/plone/documentation/issues/1184.

Back to my review of this PR, here are my suggestions.

Screen Shot 2023-01-10 at 2 08 39 AM

## `kbd` style

In a page where there is narrative text, the `kbd` style is fine as is, but in a search input where the font is much smaller than the body text, its `font-size` needs to be increased. The `⌘` is not legible. The vertical padding is too much. Here are suggested revised styles and a preview. Keep in mind that I do not have a Retina monitor for super crisp images, and I have a slight visual impairment (yay for aging!).

```css
.input-group-text kbd {
  padding: 0rem 0.4rem;
  font-size: 135%;
```

Screen Shot 2023-01-10 at 2 13 40 AM

## Search icon

I think the search icon (magnifying glass) is helpful and universally understand as "search". Google, DuckDuckGo, and many other search engines have this search icon in their search inputs. I would like to see it restored. In fact, I would prefer the icon instead of placeholder text to keep the design clean.

## Placeholder

I thought about this more, and I think it is better to have either just "Search", dropping "the docs", or nothing at all as the placeholder text. We should spell out words for the sake of both visually impaired users and translations, and "docs" should be avoided. Thus it becomes only "Search", but with the search icon reinstated, any placeholder text becomes unnecessary. Additionally it is clear that the search context is documentation, so we don't need to explicitly say "Search the docs", especially since it says "Plone Documentation v6.0" just above the sidebar search input.

## Search page

- Make the input a `type="search"`. This provides a clear or remove icon `×` that displays when the user starts typing in most browsers.
- Remove the `reset` button as redundant to the previous item.
- Replace the `search` text in the button with the search icon.
- Remove the placeholder text as redundant to the search icon.

Let's try to get consensus on this before moving forward. Thank you!

ksuess:
From me a +100.

---
Placing the label on the right side as @MAX-786 suggested, is fine for me. I have no preferences anymore.

MAX-786:
> Let's try to get consensus on this before moving forward. Thank you!

A "+" from me too :) (newbie question time: should I commit these suggestions or ksuess will do? and is it safe to close this PR #1405 now?)

> Also we, and now the Plone Foundation, have been talking about using Nuclia for search on plone.org and docs: https://github.com/plone/documentation/issues/1184.

After going through the PR, just wondering that does it have the potential to be a gsoc project? whether or not I would love to contribute to it, I may ask there directly for more info on this. :)

stevepiercy:
> > Let's try to get consensus on this before moving forward. Thank you!
> 
> A "+" from me too :) (newbie question time: should I commit these suggestions or ksuess will do? and is it safe to close this PR #1405 now?)

I would say the easiest way forward would be to grab the bits and pieces that we like and merge and implement them into your PR for review.

> > Also we, and now the Plone Foundation, have been talking about using Nuclia for search on plone.org and docs: #1184.
> 
> After going through the PR, just wondering that does it have the potential to be a gsoc project? whether or not I would love to contribute to it, I may ask there directly for more info on this. :)

It might. I have no idea of the full scope. @ebrehault is one of the leaders of GSoC for Plone, and @polyester would have a good idea of what is necessary for search integration. I think he did something with Algolia search for previous versions of the docs. He ran into some trouble updating it, and we might just dump it for Nuclia.

polyester:
Working on the Nuclia stuff, and have dropped the idea of making this into a GSOC project. Give it a few days.

MAX-786:
> I would say the easiest way forward would be to grab the bits and pieces that we like and merge and implement them into your PR for review.

Sure! consider it done..... will be done by tomorrow (its 3am here so 😁😴)

> Working on the Nuclia stuff, and have dropped the idea of making this into a GSOC project. Give it a few days.

Oh looking forward to it 😁

stevepiercy:
@ksuess can this PR be closed in favor of #1405?

ksuess:
Suggestions are included in original PR https://github.com/plone/documentation/pull/1405

PR #1423, opened 2023-01-14
Add more details about plone.volto

FarooqAlaulddin:
I found this script helpful plone.volto.browser.migrate_richtext.migrate_richtext_to_slate with richtext migration. It was not mentioned on this page or other pages related to blocks migration (from what I saw). I had to spend more time looking for something like this. I think this note will help other people.

### DIFF
--- a/docs/backend/upgrading/version-specific-migration/migrate-to-volto.md
+++ b/docs/backend/upgrading/version-specific-migration/migrate-to-volto.md
@@ -69,7 +69,8 @@ The required steps are:
     ```shell
     docker run -it -p 5000:5000 plone/blocks-conversion-tool:latest
     ```
-
+    plone.volto pacakge provide a script `migrate_richtext_to_slate` that can help you migrate your richtext to slate or draftjs (see YOUR-SITE/@@migrate_richtext).
+    
     For more options read https://github.com/plone/blocks-conversion-tool.
 
 1.  **Pages, News Items, and Events are made folderish**


stevepiercy:
In the introduction, there is one sentence:

> Plone provides a form `/@@migrate_to_volto` that allows you to run all the required changes to your existing site to make it compatible with Volto.

But then in your PR, you say there is another form at `/@@migrate_richtext`.

@FarooqAlaulddin were you aware of `/@@migrate_to_volto`?

If so, what did not work for you?

What did `/@@migrate_richtext` do for you that `/@@migrate_to_volto` did not do?

@pbauer why do we have two forms for migrating rich text fields from Classic UI to Volto?

It looks like `/@@migrate_richtext` is for legacy conversions of rich text fields to DraftJS blocks for Volto 15.

It also looks like `/@@migrate_to_volto` is intended to convert rich text fields to Slate blocks for Volto 16.

For reference:

- https://6-classic.demo.plone.org/@@migrate_to_volto
- https://6-classic.demo.plone.org/@@migrate_richtext
- https://github.com/plone/plone.volto/tree/main/src/plone/volto/browser

FarooqAlaulddin:
@stevepiercy I was not aware of `/@@migrate_to_volto` or noticed this note when I was reading the migration notes from Classic UI to Volto. All I saw regrading the richText to slate blocks is the mentioned of the conversion tool.

If so, what did not work for you?
What did /@@migrate_richtext do for you that /@@migrate_to_volto did not do?
- What worked for me is @@migrate_richtext. This allows me to select what I want to migrate as not all of my content-types had a richText fieldname = 'text'. Some of my schema has a different name for richText. `/@@migrate_richtext` allowed me to change the fieldname to handle other fieldnames (editing the script).

stevepiercy:
> @stevepiercy I was not aware of `/@@migrate_to_volto` or noticed this note when I was reading the migration notes from Classic UI to Volto. All I saw regrading the richText to slate blocks is the mentioned of the conversion tool.

This can be improved with a rewrite. I can take that on in a separate pull request.

We want to direct the vast majority of users to use `/@@migrate_to_volto` because it handles most use cases.

> If so, what did not work for you? What did `/@@migrate_richtext` do for you that `/@@migrate_to_volto` did not do?
> 
> * What worked for me is `@@migrate_richtext`. This allows me to select what I want to migrate as not all of my content-types had a richText fieldname = 'text'. Some of my schema has a different name for richText. `/@@migrate_richtext` allowed me to change the fieldname to handle other fieldnames (editing the script).

This might be a misunderstanding between languages. If I read the following code correctly, the attribute `fieldname`'s name does not matter, but its type does. `migrate_richtext_to_blocks()` will convert only portal objects whose `fieldname` attribute is an instance of `RichTextValue`.

https://github.com/plone/plone.volto/blob/7e16d711ec92185d1315826e113b36a9995511b5/src/plone/volto/browser/migrate_richtext.py#L83-L87

Thus I think that if you had used form `/@@migrate_to_volto`, you would not need to customize `/@@migrate_richtext`. I might not understand what we were trying to do, but it appears to me that you "reinvented the wheel".

FYI, the code called by the form at `/@@migrate_to_volto` imports and calls the same methods called by the form at `/@@migrate_richtext`.

https://github.com/plone/plone.volto/blob/7e16d711ec92185d1315826e113b36a9995511b5/src/plone/volto/browser/migrate_to_volto.py#L11-L12

In any case, let's see if @pbauer can enlighten us about the intent of these two forms.

fredvd:
I think I can give some more guidance here as well:

"Migrate to richtext" is only concerned with converting (a) richtext field(s) on selected or all content items in your site to the new blocks_layout field in the plone server where all contnt is stored.  That's why you can select content types to convert.

But to successfluly migrate your 'ClassicUI' site contents to the *default setup* that the Plone 6 ReactUI based (Volto) frontend requires, more needs to be done.   Such as: 

* Activate the plone.volto profile in your site
* Pages and Folders are merged to Folderish pages
* default views are converted by pulling the contents from the default page to the folderish page above
* Convert the Collection content type to a Folderish page with the collection/listing tile in the block layout
* AND convert the richtext fields to blocks layout

As you can see  "migrate to volto" does it all, migrate richtext is a part of if.  For custom setups, migrate to richtext allows you low level access to this tool . 

Why document the 'migate richtext' view?  Because if you are an integrator, developer or expert on migration, and don't want/need the default setup with the context type  it gives us flexibility in when to process the individual steps. 

@FarooqAlaulddin Is this an explanation that gives you more understanding, are there other questions that should be answered specifically to this topic?

@stevepiercy Do you want to update the paragraphs/section involved, or should I suggest an improvement after @FarooqAlaulddin gives his feedback?

stevepiercy:
@fredvd this is outside my domain of understanding, so I would gladly turn it over to you.

PR #1432, opened 2023-01-26
Improve spellcheck

lord2anil:


stevepiercy:
This file is deprecated and no longer used, except as a record of what needs to be migrated eventually into the new spell checker, Vale. See https://6.docs.plone.org/contributing/setup-build.html#vale for the proper implementation of new words.

PR #1439, opened 2023-02-07
Update deprecated command of `set-output`

stevepiercy:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
         working-directory: submodules/volto
-        run: echo "{name}={value}" >> $GITHUB_OUTPUT
+        run: echo "{dir}={$(yarn cache dir)}" >> $GITHUB_OUTPUT
       - uses: actions/cache@v1
         id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_deploy.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
         working-directory: submodules/volto
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+        run: echo "{name}={value}" >> $GITHUB_OUTPUT
```
fredvd: @stevepiercy I think this should be changed to:

run: echo "{dir}={$(yarn cache dir)}" >> $GITHUB_OUTPUT

@sneridagh @ericof    That's quite a few workflows we maybe have to check

fredvd (changes requested):

.github/workflows/build_deploy.yml
stevepiercy: D'oh! That's what happens when we go on auto-pilot and copy-pasta. Fixed in [8e0614a](https://github.com/plone/documentation/pull/1439/commits/8e0614a6773a15858e6adc54c279bc0495df7f6e). Please check again.

fredvd (approved):
This should work now I guess.  And probably more repo's where we need to update the workflow,

fredvd:
@stevepiercy That documenation on the set-output deprecation is very vague: those accolades shouldn't be there as well. I'm now creating a branch to upgrade the workflows with updated actions  so the node 12 warning goes away. 

For the action-rsyncer update to 2.0 I found this commit that shows how to update the ::set-output:
https://github.com/Pendect/action-rsyncer/commit/c917719dd01a4b68c4f2e1527c8cca6835217f0f

But updating setup-node to v3 means we can drop the set-output alltogether

stevepiercy:
I confirmed that this resolved the deprecation warning for `set-output`. Thanks for the PR on the other CI issues.

PR #1441, opened 2023-02-07
add docs for static js and css resource registry

toalba:
As a result of the Alpine City Sprint 2023, I  updated the documentation for static resource registry in the classic ui.


stevepiercy (approved):
Thank you!

PR #1442, opened 2023-02-08
Fix layout of search

ksuess:
Fix order of _static directories. Last wins.
volto/_static or other submodules _static should not overwrite main _static.


stevepiercy:
Thank you!

PR #1444, opened 2023-02-12
Update build-deploy workflow for outdated actions

fredvd:
* Remove matrix variables because there is (no longer?) a matrix configuration
* Update setup-python to remove node 12 warning, use built-in cache 
* update setup-node, user built-in cache support
* set yarn lock file location
* Update action-rsyncer  to 2.0.0


fredvd:
The updated workflow seems to run and deploy fine, but the advertised now builtin caching support in the latest setup-node and setup-python versions seem to interact.  The python built is cached, but the node one isn't, and if you check the output of the post actions where the cache is saved you can see why::

https://github.com/plone/documentation/actions/runs/4158053094/jobs/7192998699

build-deploy-post

The setup-node cache is 0 mb.  So I'll either need to look up hopefully existing parameters to point to the right directory, or go back to manual cache actions

fredvd:
Hmm.  I'll look into the cache action tomorrow.  Apparently the out of the box cache integration in setup-node for now only works if the project is in the root,  other locations is 'work in progress' with weird workarounds: 

https://github.com/actions/setup-node/blob/main/docs/adrs/0000-caching-dependencies.md#goals--anti-goals

https://github.com/actions/setup-node/issues/531

fredvd:
@stevepiercy @ericof updating the build-deploy on this branch/pr

fredvd:
Sorry, I better ping @sneridagh here.   

Now that I’ve seen this workflow, and want to fix cache support, why are storybook build/deploy and the docs deploy in one workflow? 

Isn’t it better to separate them?  Do we keep different storybooks per released volto or docunentation releases?

.github/workflows/build_deploy.yml
```diff
@@ -4,18 +4,18 @@ on:
   push:
     branches:
       - "6-dev"
+      - "update_workflows"
```
stevepiercy, line 7: Should we use this branch for updating workflows in the future, or is it just a one-time thing and should be removed before merging?

stevepiercy (dismissed):
LGTM, pending resolution of merge conflict, and review from AI team.

stevepiercy:
> Now that I’ve seen this workflow, and want to fix cache support, why are storybook build/deploy and the docs deploy in one workflow?

I'm sorry, I don't follow. Can you point out where you see this, so I understand what you mean?

> Isn’t it better to separate them? Do we keep different storybooks per released volto or docunentation releases?

We do not maintain separate versions of storybook. Perhaps we should? That's more of a Volto internal question.

Originally the idea was to move the storybook from `voltocms.com` to `docs.plone.org`. It was piggy-backed onto deploying docs to give it a home under `plone.org`. These days, I think it makes sense to pull the storybook out of `6.docs.plone.org` and into `#.storybook.plone.org`, where it can keep its versioning aligned with Volto. It would definitely speed up docs deployment.

.github/workflows/build_deploy.yml
fredvd: this is just temporarily because otherwise the worklow doesn't run and the first real test will be when it gets merged to main/master.  Like what happened with our updated set variable line.

sneridagh:
> Sorry, I better ping @sneridagh here.
> 
> Now that I’ve seen this workflow, and want to fix cache support, why are storybook build/deploy and the docs deploy in one workflow?
> 
> Isn’t it better to separate them? Do we keep different storybooks per released volto or docunentation releases?

@fredvd for now it's ok to keep them together. In the future if the storybooks for different versions really differ, we can consider doing it.

sneridagh:
@fredvd @stevepiercy reading again, just to be more clear. I meant together, as not maintain several versions of the storybook build. Separate them in different build jobs is fine for me!

fredvd:
Created issuse #1450 to continue/store the discussion on splitting storybook. 

Closing this PR/branch and creating a new one with only the workflow updates.  I did a rebase on 6-dev but I don't trust myself, why is my branch suddenly 14 commits behind 6-dev....

fredvd:
New PR: #1451

Issue #1453, opened 2023-02-21
graphviz not rendering


stevepiercy:
### Plone Documentation versions affected

- [ ] 5.x
- [X] 6.x

### Description

The graphviz directive stopped rendering on the deployed site.

https://6.docs.plone.org/classic-ui/views.html

I'm not sure why. Recently merged PRs' Netlify previews look fine.

stevepiercy:
Fixed in https://github.com/plone/documentation/commit/e83a788bf49083213456b28a222d85d1c9a9df53

@MrTango @erico FYI https://6.docs.plone.org/classic-ui/views.html

PR #1454, opened 2023-02-22
Attempt to fix graphviz rendering

stevepiercy:
Remove redundant configuration option

Refs: #1453


stevepiercy:
Netlify preview looks OK: https://deploy-preview-1454--6-docs-plone-org.netlify.app/classic-ui/views.html

I'm going ahead and merging. Maybe there is some bad cache that needs to get cleared.

PR #1465, opened 2023-03-13
Add Override Example to Layers

### LINKED ISSUE #1426: Classic UI > Layers > Unconditional Overrides needs content

- [ ] 5.x
- [X] 6.x

### Description

Classic UI > Layers > Unconditional Overrides needs content

https://6.docs.plone.org/classic-ui/layers.html#unconditional-overrides

### Source material

?

toalba:
Since i needed it today and didn't found it in the docs i just quickly added it. I will beautify it later as well.


docs/classic-ui/layers.md
```diff
@@ -58,13 +58,54 @@ Then views and viewlets from your product can be enabled on the site instance us
 ### Unconditional overrides
 
 If you want to override a view or a viewlet unconditionally for all sites without the add-on product installer support, you need to use `overrides.zcml`.
+You can override classes and templates in this file.
```
MrTango, line 61: This is not completely true. You can override a view also by registering it in a configure.zcml if you use a more specific Browser layer. If the original does not have a browser layer defined, it is usually enough to just use the default browser layer of your addon package. If the original has a layer already, you often want to subclass your browser layer interface from the original layer interface. https://6.docs.plone.org/classic-ui/layers.html#add-on-layer-for-changing-existing-behavior

The fact that it works without the addon being installed/enabled in a Plone site, is true for all zcml-config. Only if you combine the registration with a browser layer, which will be enabled on install/enabling the addon inside a Plone side, the config will not work in other Plone sides where the addon is not enabled.
Also note, that not all registrations allow to set a browser layer. Here by default everything is active in all Plone sides, which is often not what you want. You can avoid this by checking for the Browser Layer in code, like explained here: https://mrtango.planetcrazy.de/plone-restrict-subscriber-to-plone-sides-where-the-addon-is-enabled.html and here https://6.docs.plone.org/classic-ui/layers.html#layers-are-activated-on-the-current-request-object

MrTango:
I think it does not make sense to replicate all zcml and Python code for overrides, as they are the same as for configure.zcml. The only difference is, that when you have the same registration in configure.zcml, you'll get a configuration conflict. If you put the same registration in overrides.zcml your config will replace the original one. But i would always try to got with a more specific Browser Layer where possible.
and it is also documented here: https://6.docs.plone.org/classic-ui/views.html#overriding-a-view-class

MrTango (dismissed):
i simplified the changes to whats needed

stevepiercy:
Please reopen this PR against the `6.0` branch. `6-dev` is no longer under active development.

MrTango:
@stevepiercy i don't know why there are HTTPS errors, the certs look fine of those URL's

stevepiercy:
@MrTango webarchive is often problematic. Is there another resource that can be linked? If not, can we remove the link?

I checked the cert for https://eev.ee/blog/2016/07/31/python-faq-how-do-i-port-to-python-3/ and it has not expired and it was renewed about a month ago, despite what linkcheckbroken says. I restarted the job to see if it was just temporary.

MrTango:
the links are not from me and as you said, they are not really the issue, the check somehow does not work properly.

stevepiercy:
@MrTango it seems to be working fine now. Someone fixed the Internet! 🤣

docs/classic-ui/layers.md
```diff
@@ -58,12 +58,7 @@ Then views and viewlets from your product can be enabled on the site instance us
 ### Unconditional overrides
 
 If you want to override a view or a viewlet unconditionally for all sites without the add-on product installer support, you need to use `overrides.zcml`.
-
-```{todo}
-Explain how to use an `overrides.zcml`.
-
-See https://github.com/plone/documentation/issues/1426
-```
+You can override classes and templates in this file. To do this, you put the ZCML-registration in a files called overrides.zcml in the package root (next to the top most configure.zcml).
```
stevepiercy:
```suggestion
You can override classes and templates in this file.
To do this, you put the ZCML-registration in a file called `overrides.zcml` in the package root, next to the top-most `configure.zcml`.
```

docs/classic-ui/views.md
```diff
@@ -173,7 +173,7 @@ You need to declare the `browser` namespace in your `configure.zcml` to use `bro
 ```
 
 The view in question is registered against a {ref}`layer <classic-ui-layers-label>`.
-It will be available after restart, after running the {ref}`GenericSetup profile <backend-configuration-registry-generic-setup-label>`.
+It will be available after restart and running the {ref}`GenericSetup profile <backend-configuration-registry-generic-setup-label>` or enabling the addon.
```
stevepiercy:
```suggestion
It will be available after restart and running the {ref}`GenericSetup profile <backend-configuration-registry-generic-setup-label>`, or enabling the add-on.
```

stevepiercy (changes requested):
A couple easy grammar, style, and MyST fixes, and this is good to go. Thank you!

docs/classic-ui/layers.md
```diff
@@ -58,12 +58,8 @@ Then views and viewlets from your product can be enabled on the site instance us
...
 
 If you want to override a view or a viewlet unconditionally for all sites without the add-on product installer support, you need to use `overrides.zcml`.
-
-```{todo}
-Explain how to use an `overrides.zcml`.
-
-See https://github.com/plone/documentation/issues/1426
-```
+You can override classes and templates in this file.
+To do this, you put the ZCML-registration in a file called `overrides.zcml` in the package root, next to the top-most `configure.zcml`.
```
stevepiercy:
```suggestion
To do this, you put the ZCML registration in a file called `overrides.zcml` in the package root, next to the top-most `configure.zcml`.
```

stevepiercy (changes requested):
Minor grammar, and then it is good to go.

@MrTango was your comment from last week addressed?

stevepiercy (approved):
Thank you! Merging.

docs/classic-ui/layers.md
stevepiercy: @MrTango resolved this in https://github.com/plone/documentation/pull/1465/commits/5b359dcbe960adacf124a82aa6e8781d6aef732e

PR #1471, opened 2023-03-20
TinyMCE upgrade to Plone 6, Pt. 1.

thet:
Pt. 1 of an upgrade guide for TinyMCE while working on it.

Missing: how to upgrade custom plugins.


MrTango:
I don't think there is anything different for custom plugins. And even if so, people should consult the TinyMCE docs.

MrTango (approved):
I updated the docs a bit. I think we an merge it like this.

stevepiercy:
Closing this PR because it is against a dead branch. Please see https://github.com/plone/documentation/pull/1471

PR #1472, opened 2023-03-20
Fix video warnings link errors

stevepiercy:
https://github.com/plone/volto/pull/4578 needs to be merged before this PR.


MAX-786 (dismissed):

docs/contributing/myst-reference.md
```diff
@@ -270,7 +270,7 @@ Don't use file-relative paths.
 The above MyST markup renders as shown below.
 
 ```{video} /_static/user-manual/blocks/block-copy-cut.mp4
-    :width: 100%
+    :width: 100
```
jensens: I am not sure if this is what you want to show (100px width).

screenshot_2023-03-20_16:06:56_selection

MAX-786: Yep! That's why i set it to 100%, but IG it was taking the full page width which was not intended. 
Need to check it again...

jensens: 100% is probably not that wrong. It does not look terrible large here.

If you want to reduce it to a maximum width, like 600px, but let it scale down, as far as I know the solution is in CSS and can not be passed as attribute. 

I would moreover say a width 100% is fine (at least on mobile, tablet to a certain break-point) as long as the CSS for the documentation theme sets additional a max-width on larger devices. 

```
video {
  max-width: 600px;
  height: auto;
}
```

As long as the theme is not enhanced with such, I would keep the 100%. At least 100 (px) is just not what we want.

stevepiercy:
TIL.

- Yesterday I noticed [warnings to be emitted in recent GitHub Actions](https://github.com/plone/documentation/actions/runs/4467654371/jobs/7847388030#step:5:832).
- [`<video>` should not use percentages, only integers in pixels](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#width).
- A change in the Sphinx extension [sphinx-contrib/video](https://github.com/sphinx-contrib/video/blame/ac205a04a867ac1a237e4964e16fbdbbfca27f4f/sphinxcontrib/video.py#L105-L110) was merged 27 days ago, and released 2 days ago. This change now enforces that requirement, stripping the `width` attribute when it is not an integer, and emits a warning.

The fix is to remove the `width` attribute altogether and add a style to the `<video>` object:

```css
video {
  width: 100%;
}
```

I'll revise this PR.

docs/contributing/myst-reference.md
stevepiercy: Fixed in https://github.com/plone/documentation/pull/1472/commits/33402a130e7662e9e91d900e74ff6c77115eff6d

jensens (approved):

PR #1476, opened 2023-04-01
Ignore python version save file when using pyenv

fredvd:
for python version management as a developer/contribuor to docs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 pyvenv.cfg
 /_build
 /styles/Microsoft
+.python-version
 
 # symlinked from submodule
 docs/volto


.gitignore
```diff
@@ -8,6 +8,7 @@
 pyvenv.cfg
 /_build
 /styles/Microsoft
+.python-version
```
stevepiercy, line 11: This should go in the user's gitignore, with whatever implementation they choose.

This `.gitignore` should include only those things that are generated by the project, not things that the user may or may not use.

stevepiercy (changes requested):

.gitignore
fremwilplone: https://github.com/plone/buildout.coredev/blob/6.0/.gitignore

fredvd: https://github.com/plone/buildout.coredev/blob/6.0/.gitignore

fredvd: https://github.com/plone/volto/blob/master/.gitignore

stevepiercy: That's fine for those repos, but not this one.

stevepiercy:
I'm closing this as "won't fix".

PR #1492, opened 2023-04-24
Update backend patch version to 6.0.4.

mauritsvanrees:


stevepiercy (approved):
Thank you!

stevepiercy:
The failing job is fine. I notified the site owner that their cert expired, and asked that they let me know. If it is still a problem in the next run, we can decide what to do, if anything.

PR #1497, opened 2023-05-15
Create github-administration.md

gforcada:
Explain the tags that we plan to use to organize repositories in Plone GitHub organization.

See https://github.com/plone/meta/issues/73


stevepiercy (approved):
@gforcada I pushed a commit, instead of making a review, to save effort and explanation of documentation guidelines. Please feel free to revise, discuss, or approve and merge. I approve these modifications. Thank you!

gforcada:
@stevepiercy thanks! 💯 🙇🏾 

The changes look perfect!

I will wait for others to double check if they are fine, specially this week that we have the sprint 😄

mauritsvanrees:
Possible problem is that `core` will be very large at the moment, but may drop in size in Plone 7.0, maybe 6.1 already, so packages would need to shift to a different tag then.

Would it make sense to look towards the future already, only tagging repos as core when it is highly likely that they will remain so in Plone 7?

Because right now, theoretically you could say that the `classic ui` tag would only contain `mockup`: all other packages that might belong here are dependencies of `Products.CMFPlone`, so they would be in core.

Note that it would be perfectly fine to re-tag repos every now and then.

In general: yes, good to have this documented here, and we can always revise.

stevepiercy:
Sorry for the long delay in merging this new page. Will merge when green.

Issue #1496, opened 2023-05-13
Create redirects of `docs.plone.org` to `6.docs.plone.org`


stevepiercy:
Search engines have indexed `docs.plone.org` heavily. We need to set up 301 redirects on the server.

I don't have access to do this. @polyester @fredvd have access. We should also discuss how to discover 301 redirects. Matomo should be able to do it, according to [this blog article](https://matomo.org/blog/2022/01/seo-web-analytics/), but I see no such report. It might be for the paid version only. We might need to parse server logs.

## List of redirects (add more as they are discovered)

| Source | Target |
|---|---|
| https://docs.plone.org/develop/plone.api/docs/* | https://6.docs.plone.org/plone.api/* |

Reference: post by @1letter on Community Forum:

https://community.plone.org/t/plone-6-documentation-update-2023-05-12-plone-6-documentation-released/17451/2

fredvd:
I've chewed a while on this today. @stevepiercy The 'crawling errors report' in that article is promotion for the paid SearchEngineKeywordPerformance plugin, and it only works for Bing! and Yahoo search, not for Google. 

For some searches on Plone terms I already get 6.docs.plone.org results on Google.  Within maybe a few weeks most links will already be reindexed and old suggestions dropped. 

I thought about adding a custom 404 page where we track 404's using instructions here  https://matomo.org/faq/how-to/faq_60.  But that doesn't work that way because we use a tag container. There are instructions for that situation here where you can check for the page title: https://matomo.org/blog/2019/07/how-to-analyse-404-pages/

But that's not happening here, we redirect any docs.plone/(.*) to 6.docs.plone.org, so on the destination site the page view is for the homepage, which is not a 404.  So to detect 'old' redirects on the target site we need to change the current redirect so that it actually generates a 404 into the new docs site that we can track. Then a 404 report would help. 

We could add/install Google Search Console access on the site to get more information from there. But I checked search console for another site I manage, and I don't think it will help. We would need to first register docs.plone.org in google search console, but that url is redirecting to ... 6.docs.plone.org.  If we somehow fix that (disable redirect, install token for an accoun to please search console, enable it again), then you will probably just get a big report of all pages from the old docs site that where once there that google can't find anymore. That doesn't give us any extra information.

[...]

PR #1504, opened 2023-05-26
Change wording and include reference to previous versions of Plone do…

stevepiercy:
…cumentation

Refs: #1496

### DIFF
--- a/docs/_templates/404.html
+++ b/docs/_templates/404.html
@@ -15,6 +15,11 @@
         <li><a href="https://4.docs.plone.org/">Plone 4</a></li>
         <li><a href="https://3.docs.plone.org/">Plone 3</a></li>
     </ul>
+    <h2>Why is this happening?</h2>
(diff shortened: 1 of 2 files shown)


polyester (approved):

fredvd:
Would it be helpful if we mention in an extra paragraph that the cause of the 404 is because of our restructuring of the documentation and that we unfortunately could not create redirects all old to new pages, but will add redirects in the coming months for the most requested paths. 

That would create some understanding from visitors that don't realise 'why' this is happening, but that it is a trade-off we had to make.

mamico (approved):

stevepiercy:
> Would it be helpful if we mention in an extra paragraph

Good idea. PR updated.

stevepiercy:
Oopsie, I forgot to ping @fredvd for a review on the update.

fredvd (approved):

PR #1508, opened 2023-06-08
Set value for `sitemap_url_scheme` to correctly add URL entries to si…

stevepiercy:
…temap.xml.

Currently we have values that follow the default settings value, such as the following. 

`<loc>https://6.docs.plone.org6.0/backend/annotations.html</loc>`

This change fixes the issue.


stevepiercy:
Verified that https://deploy-preview-1508--6-docs-plone-org.netlify.app/sitemap.xml looks good. Merging.

PR #1512, opened 2023-06-26
Rename the heading for accuracy

stevepiercy:
There is confusion between marketing and reality. We market Volto as the "default theme" for Volto 6, but this heading says the default theme is Barceloneta LTS. We should avoid marketing either theme as "default" in documentation to avoid alienating users of either theme. Also LTS for a theme is not a thing.


stevepiercy:
I submitted an [issue for the failing linkcheck on sass-lang.com](https://github.com/sass/sass-site/issues/749) in the GitHub Action failure. It is safe to merge this PR.

davisagli (approved):

yurj (commented):
I agree

PR #1518, opened 2023-07-11
Exclude video markup from text output

stevepiercy:
For the Plone Nuclia project, we are experimenting with various outputs for easier indexing. This PR excludes videos from only the text output from the `make text` command. It affects only the text builder.


stevepiercy:
@justdaksh check this out after CI passes.

stevepiercy:
The failing test is for anchors on GitHub source files, which can be ignored. I will fix that in another PR. Merging this one so that I do not hold up our Plone Nuclia search GSoC project.

Issue #1190, opened 2022-02-21
Improve spellcheck


stevepiercy:
It would be good to improve spellcheck and consistency of spellings. For example:

```
JavaScript, instead of javascript / Javascript / js
Node.js, instead of nodejs / NodeJS / node
```

Use caution when doing global find and replaces to not overwrite commands or code blocks that use those strings.

To improve spellcheck, add words to:

- `styles/config/vocabularies/Plone/accept.txt`
- `styles/config/vocabularies/Plone/reject.txt`

Note that when you add a correct spelling in `accept.txt`, then you do not need to add incorrect casings of the same term to `reject.txt`. See https://vale.sh/docs/keys/vocab#case-sensitivity.

## Vale usage

- https://6.docs.plone.org/contributing/documentation/setup-build.html#vale

## Contributing

- https://6.docs.plone.org/contributing/
- https://6.docs.plone.org/contributing/first-time.html
- https://6.docs.plone.org/contributing/documentation/index.html

samprasdsouza:
Hello @stevepiercy , Can I take up this issue ??

chirayu-humar:
i observed that this issue is mentioned in a pull request named https://github.com/plone/plone.restapi-client/pull/5, which is already merged so i want to conferm whether this issue is resolved now or i can start working on this ?

stevepiercy:
That issue is for a different repository and project. Spell checking can always be improved in this repository. Please feel free to work on improving the quality of spelling.

chirayu-humar:
I am working on this issue

chirayu-humar:
i have generated a draft pull request but i am not confident enough that it will work properly because i have not tested it on my local system because i did knew where is this spell check is used further in codebase, to figure it out it will take some time.

Akshat2Jain:
@stevepiercy, plz close this issue as this is already solved

stevepiercy:
@Akshat2Jain no, there is a LOT more work to do. There are still [hundreds of errors, warnings, and suggestions](https://github.com/plone/documentation/pull/1523#issue-1850843452).

IsaacRoy:
Ok sir can I start resolving this issue ?

IsaacRoy:
I have solved the issue and 3 test case passed . Can you just preview it

YashasviniSingh:
Hi, I’d like to work on this issue. Please let me know if I can take it.

stevepiercy:
[...]

PR #1523, opened 2023-08-15
Update Vale custom vocabularies

stevepiercy:
## Before

✖ 2066 errors, 1146 warnings and 4327 suggestions in 335 files.

## After

✖ 1787 errors, 1146 warnings and 4323 suggestions in 335 files.

Nice! A 13.5% drop in errors.

See https://github.com/plone/documentation/issues/1190, https://github.com/plone/volto/pull/5095, and https://github.com/plone/volto/pull/5056


styles/Vocab/Plone/reject.txt
```diff
@@ -1,5 +1,2 @@
-node
-nodejs
-javscript
-js
-Javascript
\ No newline at end of file
+[^.]js
```
davisagli, line 1: I don't understand this one. Reject any 3 letter word ending in js unless the first character is a period?

stevepiercy: Close. It matches words that end with "not-dot" followed by "js". Then reject the matches.

Thus it will match "Nodejs" and reject it, as the correct spelling is "Node.js".

side rant: Let it be known that punctuation in your branding and identity is a stupid idea. Wal-Mart is now Walmart, after they learned their lesson from bad SEO.

davisagli (approved):

PR #1524, opened 2023-08-23
Use GHA environment to deploy docs.plone.org

fredvd:
Use environment vars for plain variables, only the deploy key is secret.
Environment vars have been configured
Remove branch 6-dev from 'on:' filter rule

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -3,12 +3,14 @@ name: Build and deploy Plone 6 documentation to 6.docs.plone.org
 on:
   push:
     branches:
-      - "6-dev"
       - "6.0"
 
 jobs:
   build_deploy:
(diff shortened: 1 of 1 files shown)


fredvd:
To be able to add staging we first need to update the deploy for the live docs deploy to use the main environment
@stevepiercy  I think the 6-dev branch filtering rule is no longer needed, I also see no commits there for the last months. 

@ericof I've created environments and moved the variables and secrets there. I will remove the repo secrets as soon as this branch has been checked and merged. 

Then we should do a docs release to test it actually works to still deploy live. 

After that we can modify the build_deploy.yml on the nuclia branch to deploy using the staging environment.

@justdaksh: ^^ FYI

ericof (approved):

stevepiercy (commented):
Just got around to this. LGTM. Thank you!

PR #1461, opened 2023-03-02
add classic-ui: forms, fields, schemas, content types

MrTango:
bring the following chapters alive:

- content types
- schemas
- fields
- forms

PR #1531, opened 2023-08-30
Move dexterity reference from training to documentation

ksuess:
This is the slightly changed dexterity reference from training.
Feedback is very welcome! Feel free to enhance!


stevepiercy:
@ksuess apologies, I should have updated the issue with this update: https://github.com/plone/plone.app.dexterity/issues/360. If you agree with that direction, then let's close this issue and focus on the long-term plan.

MrTango:
Thx @ksuess for getting started with this. I think we need to improve the structure more before we merge this.
We should not mix to much UI in the fields chapter, as this in only backend. Widgets (classic UI) should go into the existing widgets chapter. Volto UI Related Widgets probably into the Volto / Frontend chapter.

I know it is not so easy to split fields from widgets, also because as far as i understand widgets and there settings might also have effects on Volto-Frontend. That is why we put the widgets chapter into backend and not classicui.
I'll try to be available during PloneConf Sprint and work on docs remotely.

MrTango (changes requested):
We should not mix to much UI in the fields chapter, as this in only backend. Widgets (classic UI) should go into the existing widgets chapter. Volto UI Related Widgets probably into the Volto / Frontend chapter.

stevepiercy:
Closing in favor of https://github.com/plone/documentation/pull/1461 per  https://github.com/plone/plone.app.dexterity/issues/360.

PR #1540, opened 2023-10-03
Add hint to set the version of Yarn

stevepiercy:
First reported on https://community.plone.org/t/error-installing-yarn-3-according-to-plone-install-docs/17737

### DIFF
--- a/docs/install/install-from-packages.md
+++ b/docs/install/install-from-packages.md
@@ -179,6 +179,12 @@ Install the latest Yarn 3 version (not the Classic 1.x one) using `npm`.
     ```console
     3.2.3
     ```
+    
+    If you do not see a version of Yarn 3, then try the following to set the active version.
+    
+    ```shell
+    yarn set version 3.x
+    ```
 
 
 (install-prerequisites-make-label)=


stevepiercy:
The failing check is a temporary issue with reaching https://expressjs.com/ and can be merged.

davisagli (commented):
Does this actually work? I think the `yarn set` command is only available if you already have yarn 3.

I think we should probably follow the official yarn docs and recommend installing with corepack. https://yarnpkg.com/getting-started/install

stevepiercy:
It works after installing Yarn 3 with the previous command of `npm install yarn@3`.

I assume `corepack enable` works from a clean install, but I have not verified.

It's easy enough to switch from `npm install yarn@3` to `corepack enable`, then set and update to latest version:

```shell
yarn set version stable  # or 3.x
yarn install
```

What do y'all think? I prefer to use authoritative documentation sources. Do we have good reason to stick with what we've done for years? Sorry, I don't know the history of installing Yarn for Volto. Maybe @sneridagh or @tiberiuichim or other Voltoistas can weigh in?

tkimnguyen:
The first time I ran into the problem I had to use corepack

stevepiercy:
Does it matter that corepack is experimental? https://nodejs.org/dist/latest/docs/api/corepack.html

tiberiuichim (approved):

stevepiercy:
I'm going ahead and merging this PR as is because it addresses a common issue.

Switching to corepack for installation is still up for discussion, but it should be discussed at a Volto Team meeting before becoming the recommended installation path going forward. I'll create an issue in the Volto repo to track it.

stevepiercy:
See https://github.com/plone/volto/issues/5276

PR #1542, opened 2023-10-05
Clean up MyST syntax

stevepiercy:


stevepiercy:
Failing linkcheckbroken is due to web.archive.org dying again. I pushed a commit to ignore it sitewide, as it is still useful but reallllllly slow sometimes yielding false negatives.

MrTango (approved):

PR #1551, opened 2023-10-20
Add instructions for how to add a site to Netlify for preview builds …

stevepiercy:
…of documentation and storybooks

### DIFF
--- a/docs/contributing/documentation/admins.md
+++ b/docs/contributing/documentation/admins.md
@@ -86,9 +86,6 @@ Commit and push your changes to a remote, and submit a pull request against [`pl
 
 ## Add a project to Netlify
 
-```{todo}
-Some details need to be added to this section, including who to ping about the request with the repository name.
-```
 To add a new site to Netlify to preview built documentation or storybooks, you need to add a new site to Netlify.
 
 1.  Visit [Team Overview](https://app.netlify.com/teams/plone/overview).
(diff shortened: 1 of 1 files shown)


stevepiercy:
We can safely ignore the failed link for https://52.demo.plone.org/. See https://community.plone.org/t/5-2-demo-is-down/18059

tkimnguyen:
Seems good. Do you still need the todo at the top of the new material about whom to ping? You pretty much covered that in the new material.

stevepiercy:
There is still a question about what the Admin actually does, for the benefit of their future self. I was hoping to capture that as well. Thus the requestor would know when they can complete their subsequent configuration steps.

tkimnguyen:
Weren't you the one who told me that we didn't need to document everything here, especially if about some other software or system? 🤓 Netlify may change what they have us do

stevepiercy:
Always being dogmatic serves no one well. There's always exceptions to guidance. In this case, we found that not documenting it wastes our time more than not, as this process and configuration is not obvious nor well-documented by Netlify, and it is performed so infrequently that we always forget what we did.

tkimnguyen:
I wasn't the one who responded to the last Netlify add request so I'm not familiar with the steps required. Who responded in that case? Or would you like to submit a new one so I can go through it and document it?

stevepiercy:
@fredvd or @polyester responded. If they want to add the missing piece, then that's up to them. If they don't respond in a week or so, then we can merge this as is, and their future selves will need to figure it out if we ever do this again. 😉 

I don't want to submit a new Netlify request.

stevepiercy:
Will merge this when green.

stevepiercy:
OK to merge. Temporary outage on linkcheck.

PR #1558, opened 2023-10-26
Add warning of requirements to build Plone

stevepiercy:
Fixes [How much RAM is required to build a Volto front end?](https://community.plone.org/t/how-much-ram-is-required-to-build-a-volto-front-end/17949) and [Dealing with heap exhaustion while building Volto 17 on limited-RAM host](https://community.plone.org/t/dealing-with-heap-exhaustion-while-building-volto-17-on-limited-ram-host/18078).

### DIFF
--- a/docs/install/install-from-packages.md
+++ b/docs/install/install-from-packages.md
@@ -28,8 +28,6 @@ Plone 6 has both hardware requirements and software pre-requisites.
 
 The hardware requirements below give a rough estimate of the minimum hardware setup needed for a Plone server.
 
-{term}`Add-on` products and caching solutions may increase RAM requirements.
-
 A single Plone installation is able to run many Plone sites.
 You may host multiple Plone sites on the same server.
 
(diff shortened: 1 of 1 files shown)


tkimnguyen (approved):
You might also consider explaining that the ram build requirements are greater for Volto front ends as compared to classic UI

stevepiercy:
@kenmanheimer

stevepiercy:
Will merge when green.

docs/install/install-from-packages.md
```diff
@@ -38,6 +36,17 @@ You may host multiple Plone sites on the same server.
 -   Minimum 512 MB hard disk space is required.
     40 GB or more hard disk space is recommended.
 
+````{warning}
+When you _build_ Plone, it will approximately double RAM requirements to _install_ Plone.
```
davisagli, line 40: Huh? Building and installing are the same, no? The distinction we really want to make is that building Plone takes more RAM than running a basic Plone site, right?

stevepiercy: In this context of a cookiecutter install, I thought we install already built Docker images. Is that not correct?

davisagli: @stevepiercy No, not really.

Running the cookiecutter creates a project structure including a Makefile with various commands. It does not immediately run any of those commands.

Some of the commands install and run Plone without using containers: `make install`, `make start-backend`, `make start-frontend`.

Some of the commands build Docker images for the project (based on the released base images, plus project-specific code) and run Plone using them: `make build-images`, the various commands starting with `stack-`, and the various commands related to acceptance tests.

All of these require the RAM to build Plone (particularly I think it's the webpack compilation that is RAM-intensive)

stevepiercy: OK, then maybe this?

```
When you _install_ or _build_ Plone, it will approximately double RAM requirements to _run_ Plone.
```

Also is the preceding bullet point even true anymore?

```
-   Minimum 256 MB RAM and 512 MB of swap space per Plone site is required.
    2 GB or more RAM per Plone site is recommended.
```

davisagli: How about:

Installing and running the Plone backend requires at least 256 MB of RAM and 512 MB of swap space per Plone site.
2 GB or more RAM per Plone site is recommended.
Building the Volto frontend requires 2 GB of RAM.

stevepiercy: Wasn't there something about lxml when installing Plone backend that bumped up RAM requirements? Ah, maybe swap? https://community.plone.org/t/struggling-to-install-plone-on-ubuntu-droplet/3301/6

davisagli: @stevepiercy hmm, ok, I didn't know about that. Although these days in most cases on Linux pip will be able to fetch a prebuilt binary wheel for lxml rather than building it in situ.

How about something along these lines:

Running Plone requires at least 512 MB of RAM per Plone site. 2 GB or more RAM per Plone site is recommended.
2 GB or more RAM may also be needed temporarily while installing Plone and building the frontend assets with webpack. On a machine without much RAM, one way to get through the installation is to add swap.

stevepiercy: This is a messy section. See new PR: https://github.com/plone/documentation/pull/1561

PR #1561, opened 2023-10-28
Revise system hardware requirements

stevepiercy:
See https://github.com/plone/documentation/pull/1558#pullrequestreview-1702829041

### DIFF
--- a/docs/install/install-from-packages.md
+++ b/docs/install/install-from-packages.md
@@ -31,16 +31,18 @@ The hardware requirements below give a rough estimate of the minimum hardware se
 A single Plone installation is able to run many Plone sites.
 You may host multiple Plone sites on the same server.
 
--   Minimum 256 MB RAM and 512 MB of swap space per Plone site is required.
+-   Installation of the Plone backend and Classic UI frontend requires a minimum of 256 MB of RAM and 2GB of disk swap space.
+-   Installation of the Volto frontend requires a minimum of 2GB of RAM.
+-   After installation, running Plone requires a minimum of 256 MB RAM and 512 MB of swap space per Plone site.
     2 GB or more RAM per Plone site is recommended.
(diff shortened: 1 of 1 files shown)


docs/install/install-from-packages.md
```diff
@@ -31,16 +31,18 @@ The hardware requirements below give a rough estimate of the minimum hardware se
 A single Plone installation is able to run many Plone sites.
 You may host multiple Plone sites on the same server.
 
--   Minimum 256 MB RAM and 512 MB of swap space per Plone site is required.
+-   Installation of the Plone backend and Classic UI frontend requires a minimum of 256 MB of RAM and 2GB of disk swap space.
+-   Installation of the Volto frontend requires a minimum of 2GB of RAM.
+-   After installation, running Plone requires a minimum of 256 MB RAM and 512 MB of swap space per Plone site.
     2 GB or more RAM per Plone site is recommended.
```
davisagli, line 37: I think we could cover most of this in a simpler way:

2 GB of RAM is recommended for installing and running Plone.
It may be possible to run Plone with as little as 256 MB of RAM, but in that case you will probably still need at least 2 GB of disk swap space to complete installation.

stevepiercy: Simplification makes it less informative for those who want to minimize their resources or not install Volto.

davisagli: I think it still gives the important info for that scenario (i.e. use swap if you run into problems) but I don't feel strongly.

davisagli (approved):

docs/install/install-from-packages.md
stevepiercy: Yeah I agree, but I see too many recent posts in the Community Forum about this, so I want to be explicit and definitive. If it creates more confusion then prevents frustration, then we can revise it again.

PR #1563, opened 2023-10-30
Remove regex from sphinx-copybutton config, now that linenos are exclu…

stevepiercy:
…ded by default

See https://sphinx-copybutton.readthedocs.io/en/latest/use.html#automatic-exclusion-of-prompts-from-the-copies

Fixes #5345.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -204,11 +204,6 @@ ogp_custom_meta_tags = [
 ]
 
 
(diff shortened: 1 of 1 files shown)


stevepiercy:
@arujjval when Netlify completes its preview build of this PR, would you please test it out to verify whether this fixes the bug your reported? Thank you!

PR #1567, opened 2023-11-04
Reduce the severity level of `Microsoft.Contractions` and `Microsoft.…

stevepiercy:
…Units from `error` to `suggestion`.

These two rules' default severity level is `error`, which is robotic, annoying, and unnecessary. For example:

```console
 155:69  error       Use 'can't' instead of 'cannot'.                              Microsoft.Contractions    
 171:94   error       Don't spell out the number in 'in pixels'.                 Microsoft.Units  # "in" is not a number!
```

This will also move us closer to requiring Vale to pass without errors in our CI.

## Before

✖ 3678 errors, 1590 warnings and 5066 suggestions in 335 files.

## After

✖ 2836 errors, 1590 warnings and 5908 suggestions in 335 files.

See https://vale.sh/docs/topics/config/#basedonstyles


tkimnguyen (approved):
Hallelujah. I thought those errors were ridonkulous. I ain't havin' any of it

PR #1580, opened 2023-11-23
Update nginx-plone.md

hipertracker:
Add an Nginx mod rewrite rule for the created Plone site instance so it can be opened at http://plone.localhost instead of http://plone.localhost/Plone.


ericof (approved):
Thanks for the fix

PR #1584, opened 2023-12-01
Update licenses used in Plone and paths to news

stevepiercy:
We omitted licenses used for Documentation, Training, and Volto.

Volto moved to a monorepo, and its news path changed. Also, technically speaking, only packages, not projects, have releases with change log entries.

This PR corrects all these issues.


sneridagh (approved):

davisagli (commented):
@stevepiercy Thanks!

PR #1585, opened 2023-12-01
Fix storybook build. Update tip to volto.

sneridagh:

PR #1586, opened 2023-12-02
Fix storybook build without submodule

stevepiercy:
Replaces #1585


stevepiercy:
The same changes to this file were approved yesterday. Merging.

PR #1591, opened 2023-12-07
Move Vale from a manual installation to pip

stevepiercy:
- Move Vale from a manual installation to pip
- Move the install and configuration documentation of Vale to `authors.md`.

Thanks to https://github.com/daniperez/vale-python-package, and a [quickly merged pull request](https://github.com/daniperez/vale-python-package/pull/29), we now have this option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,8 +196,8 @@ linkcheckbroken: deps  ## Run linkcheck and show only broken links
 
 .PHONY: vale
 vale: deps  ## Run Vale style, grammar, and spell checks
-	vale sync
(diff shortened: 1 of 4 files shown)


polyester (approved):
Checked on an Ubuntu/PopOS 22.04 machine, with also an instance of vale installed via OS; it does not interfere with that setup, and works.

stevepiercy:
Thanks for the verification. Mine was macOS M2 (arm64) which necessitated my PR with the other repo. I think we're good to go.

PR #1594, opened 2023-12-24
Use includes for installation of prerequisites

stevepiercy:
This PR must be merged immediately after https://github.com/plone/volto/pull/5564 is merged and the git submodule tip for volto is updated.

The purpose of this PR is to avoid copy-pasta and drifting installation instructions. Further centralization may be needed.

See also https://github.com/plone/volto/issues/3289.


stevepiercy:
I notified the owners of opensource.org that their secure certificate is not valid. Merging.

PR #1600, opened 2024-01-14
Try upgrading sphinx and sphinx-book-theme

davisagli:
@stevepiercy Tried this locally. It installs but I'm getting errors while building the docs, so I'll push it here to make it easier to take a look.


davisagli:
@stevepiercy It looks like there were some updates to the theme templates that went along with upgrading the theme in https://github.com/plone/training/pull/757. I haven't had time to look at the details yet.

stevepiercy:
@davisagli I'm putting this off until I can release https://github.com/plone/plone-sphinx-theme. There are too many things that need to be copied over from Training to make this work for all the docs.

davisagli:
@stevepiercy oh, phooey, I didn't know about that, and I've been working on copying things from training for the past hour or two

davisagli:
@stevepiercy I pushed what I got to... probably some of this should go in plone-sphinx-theme instead. Things seem to be generally working, but the search results still need some style love.

davisagli:
Closing this since we should use plone-sphinx-theme instead

PR #1601, opened 2024-01-14
Clarify install methods

### LINKED ISSUE #1598: Clarify install methods to either "Create a project" or "Contribute to a package"

- [Install](https://6.docs.plone.org/install/index.html)
- [Containers](https://6.docs.plone.org/install/containers/index.html)
- [Install Plone from its packages](https://6.docs.plone.org/install/install-from-packages.html)

### Description

The Plone install story lacks clarity.

For Volto we reached the conclusion that you either want to install Volto for a _project_ or to _contribute_ to it and updated its documentation accordingly.

- [Develop Volto core](https://6.docs.plone.org/volto/contributing/developing-core.html)

We want to adopt this framing of installation for the entirety of Plone.

As a side note, we also promote using [Containers](https://6.docs.plone.org/install/containers/index.html) as an installation method. This should be demoted as an alternative installation method until both we clearly define the use cases for using containers over packages and the following issues and pull requests are resolved.

- https://github.com/plone/documentation/issues/1434
- https://github.com/plone/documentation/issues/1360
- https://github.com/plone/documentation/issues/1358
- https://github.com/plone/documentation/issues/1349
- https://github.com/plone/documentation/issues/1314
- https://github.com/plone/documentation/pull/1456

For Plone itself, we cannot make a distinction between Create a Plone project and Contribute to a Plone package because Plone is not a single package but consists of _all_ of its packages, each of which may have its own story for contributing. For Plone itself, we should update [Install](https://6.docs.plone.org/install/index.html) as follows.

- [x] Replace text under under [Install](https://6.docs.plone.org/install/index.html) with the following: "In this part of the documentation, you can find how to try Plone or how to install Plone to either _Create a Plone project_ or _Contribute to a Plone package_."
- [x] Demote Containers to the last item under the Install navigation.
- [x] Demote Containers below Packages.
- [ ] Define use cases for when containers is preferred over packages, and conversely packages over containers.
- [x] In [Getting started](https://6.docs.plone.org/install/index.html#getting-started), move "Try a demo" into narrative text as a sub-heading under [Install](https://6.docs.plone.org/install/index.html).
- [x] In [Getting started](https://6.docs.plone.org/install/index.html#getting-started), remove the Install content.
- [x] In [Getting started](https://6.docs.plone.org/install/index.html#getting-started), create two new cards, "Create a project" and "Contribute to a Plone package". Each should have sufficient information to guide developers on the correct path, mentioning "The [Training](https://training.plone.org/) you do should specify which path to follow."
- [x] "Create a project" should refer to [Install Plone from its packages](https://6.docs.plone.org/install/install-from-packages.html) (and possibly [Containers](https://6.docs.plone.org/install/containers/index.html), although I am not yet convinced this is a good path).
- [x] "Contribute to a Plone package" should be a new section that consists of an introduction and a link to [Contributing to Plone](https://6.docs.plone.org/contributing/index.html).

Feedback appreciated.

stevepiercy:
Fixes #1598.

Note that the new file `create-project.md` consists of content that was formerly at `install-from-packages.md`, and the latter is now a placeholder for the former. I also vastly simplified the content on `install/index.md` as it was not relevant.


docs/install/install-from-packages.md
```diff
@@ -7,453 +8,8 @@ myst:
...
-:alt: Plone login page
-:class: figure
-```
-
-Now you can edit content or configure your Plone site.
-
-You can stop the site with {kbd}`ctrl-c`.
 
-Enjoy!
+This page has been moved and renamed to {doc}`create-project`.
```
davisagli: It looks like https://documatt.com/sphinx-reredirects/index.html might be a nicer way to do this (so that it automatically redirects to the new location)

docs/install/create-project.md
```diff
@@ -0,0 +1,457 @@
...
+    "property=og:title": "Create a Plone project"
+    "keywords": "Plone, Plone 6, create, project, install, cookiecutter"
+---
+
+
+(create-a-project-label)=
+
+# Create a project
+ 
+This chapter describes how you can create a web application project using Plone, with full control over development and deployment.
```
davisagli, line 15: Would it make sense to add something along these lines to redirect people are actually intending to make a contribution?

"This is the correct path to follow if you want to build a website using Plone. If you'd like to make changes to Plone itself, follow the contributor documentation [link] instead."

docs/install/index.md
```diff
@@ -11,96 +11,53 @@ myst:
 
 # Install
```
davisagli, line 12: The content of this page looks good but I think the headings are a bit confusing, because only half of the material here is about installing Plone (the demos do not require installing anything).

I think it would help if you swapped the "Getting started" and "Install" headings (so Getting Started is at the top as h1 and Install is further down as h2).

This will change the label in the nav from Install to Getting Started. I think that's fine?

davisagli (commented):
@stevepiercy I have a few suggestions, but this looks great!

docs/install/install-from-packages.md
davisagli: (I would leave that for another PR though. Not a blocker.)

stevepiercy: Yeah, I was hesitating about that because there is an open PR https://github.com/plone/documentation/pull/1244. I think we can proceed, using that extension anyway.

I will also ask the @plone/ai-team to install a proper 301 redirect on the server after this is merged, as I do not have permission to do that. Once the 301 is installed, I can remove the original file.

`install/install-from-packages.html` to `install/create-project.html`

stevepiercy: Done in a2c12c65b80ca77f578595bd23f6c60ef37e6c59

stevepiercy: Also asked @plone/ai-team in Discord.

docs/install/create-project.md
stevepiercy: Yes, of course, a cross-reference, especially for those who land here first, is needed. Fixed in 070ea528e971db53ff56d08b1ce7e4a5406bd5ff

docs/install/index.md
stevepiercy: Swapping headings means renaming the file and its title would change. I don't want that, as most visitors to _documentation_ (not plone.org) do in fact want to install Plone, and only a few realize, "Oh, I don't need to install it, there's a demo. Maybe I should try it before I buy it?"

This is a common issue in many open source projects. Do we tell people "Getting started", or "Install", or demo, or some combination of the above? I think "Getting started" is not a good primary path in documentation because it buries "Install", and I think that "Install" is more important than the generic catch-all "Getting started". Most of the searches in our Matomo stats since we started collecting data I see are for "Install" (7) not "Getting started" (1). Finally "Getting started" is not the same as trying a demo. Thinking out loud, I don't think we want a separate demo page, either. I'd prefer to direct people to a single page from which decisions are made, as it is now. plone.org handles this better than documentation could, forcing a visitor to make a decision in its menu.

Screenshot 2024-01-15 at 6 21 55 AM

davisagli: @stevepiercy I find these arguments convincing. We're optimizing for what helps people find the right information, not for the structure that makes sense in my head. :)

davisagli (approved):

stevepiercy:
With one approval, I'm going ahead and merging this PR, as it helps unblock other needed changes. We can always revise it later.

PR #1609, opened 2024-01-21
Add glossary terms Guillotina and Nick

stevepiercy:
These terms are mentioned in the Volto docs when creating a creating a project with Volto as a frontend to an existing backend.


stevepiercy:
I need to merge this for the Volto PR.

PR #1620, opened 2024-02-01
Use a custom robots.txt and set headers to block indexing on Netlify …

### LINKED ISSUE #1619: Netlify previews indexed by search engines. Should not, don't they?

-

### Description

sorry, screenshot, as search results are local. You see what I mean…

grafik

stevepiercy:
…preview builds

Fixes #1619.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,8 +235,7 @@ netlify:
 	ln -s ../submodules/volto/docs/source ./docs/volto
 	ln -s ../submodules/plone.restapi ./docs/plone.restapi
 	ln -s ../submodules/plone.api/docs ./docs/plone.api
-	cd $(DOCS_DIR) && sphinx-build -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
-	cp netlify_robots.txt $(BUILDDIR)/html/robots.txt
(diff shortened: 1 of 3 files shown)


stevepiercy:
Got it: https://deploy-preview-1620--6-docs-plone-org.netlify.app/robots.txt

ksuess:
Even here in 'plone/documentation' I think the "branch deploy" option could be set to none. If I am correct, the Netlify deploy is not used for production deploy to docs.plone.org, but the _build dir of the rep is deployed to cloudflare as it is, isn't it?
grafik

stevepiercy:
I think it always deploys at least the production branch, even if set to "None". There's a link to its docs for its settings below the options. That page is confusing.

stevepiercy:
Also that screenshot is for Volto, not Documentation. I set both to None, but the production branch always gets deployed on Netlify. I don't know about Cloudflare.

ksuess:
Yes, one published deploy always seems to be provided by Netlify. I also do not see a way to prevent every deploy but the preview, which is indeed where handy.
Approving.

ksuess (approved):

davisagli:
@ksuess thanks!

PR #1624, opened 2024-02-13
Improve docs for classicui-images: direction parameter

jensens:
Scaling direction docs were incomplete. I took the notes from https://github.com/plone/plone.scale/blob/master/plone/scale/scale.py#L402-L433 and turned them into some more doc-like.

The alternative spelling are in fact kind of deprecated. they still work and are all over the code, but we want people to use the new terms to reduce confusion.

### DIFF
--- a/docs/classic-ui/images.md
+++ b/docs/classic-ui/images.md
@@ -237,6 +237,10 @@ image_scale = scaling_util.publishTraverse(context.REQUEST, groups[1])
 
 ## Scaling `direction`
 
+```{versionchanged} 6.0
+Added new option names to align with CSS `background-size` values, and deprecated previous names.
+```
+
 Scaling is intended for the optimal display of images in a web browser.
 
 To scale an image, you can use the `direction` parameter to control the scaling output.
--- a/docs/classic-ui/images.md
+++ b/docs/classic-ui/images.md
@@ -114,7 +114,7 @@ tag = scale_util.tag(
     scale=None,
     height=None,
     width=None,
-    direction="thumbnail"
+    mode="scale"
 )
 ```
 
@@ -233,9 +233,9 @@ image_scale = scaling_util.publishTraverse(context.REQUEST, groups[1])
 ```
 
 
-(classic-ui-images-scaling-direction-label)=
+(classic-ui-images-scaling-mode-label)=
 
-## Scaling `direction`
+## Scaling `mode`
 
 ```{versionchanged} 6.0
 Added new option names to align with CSS `background-size` values, and deprecated previous names.
@@ -243,13 +243,13 @@ Added new option names to align with CSS `background-size` values, and deprecate
 
 Scaling is intended for the optimal display of images in a web browser.
 
-To scale an image, you can use the `direction` parameter to control the scaling output.
+To scale an image, you can use the `mode` parameter to control the scaling output.
 You must use either `width` or `height`, or both.
 
 Three different scaling options are supported.
 They correspond to the CSS [`background-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size) values.
 
-The possible options for `direction` are listed below, where the default option is `scale`.
+The possible options for `mode` are listed below, where the default option is `scale`.
 
 `scale`
 :   This is the default option.
--- a/docs/classic-ui/images.md
+++ b/docs/classic-ui/images.md
@@ -237,36 +237,36 @@ image_scale = scaling_util.publishTraverse(context.REQUEST, groups[1])
 
 ## Scaling `direction`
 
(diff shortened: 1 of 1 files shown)


stevepiercy:
I'm very confused. It appears that there are two libraries for scaling images in Plone.

- https://github.com/plone/plone.scale as referenced in this PR. My guess is that this package actually resizes the images on the file system.
- [ImageScaling](https://github.com/plone/plone.namedfile/blob/ecf33a2bc7a8c61888909bc383b3e08d80888e43/plone/namedfile/scaling.py#L350) according to [Image handling](https://6.docs.plone.org/classic-ui/images.html). My guess is that this package stored metadata about the image so that when it is rendered in Classic UI, it will display properly by modifying the CSS class attribute in a template.

Can someone clarify my guesses of which packages are actually used, how they are used, and any change history? Specifically in what version of Plone were the options `keep`, `thumbnail`, `scale-crop-to-fit`, `down`, `scale-crop-to-fill`, `up` deprecated? @jensens @mauritsvanrees @gforcada @MrTango 

I can add admonitions for `versionchanged` and `versiondeprecated` as needed.

docs/classic-ui/images.md
```diff
@@ -237,16 +237,35 @@ image_scale = scaling_util.publishTraverse(context.REQUEST, groups[1])
...
+:   This is the default option.
+    `scale` scales to the requested dimensions without cropping.
+    The resulting image may have a different size than requested.
+    This option requires both `width` and `height` to be specified.
+    It does not scale up.
+
+    Deprecated spellings: `keep`, `thumbnail`.
+
+`contain`
+:   `contain` starts by scaling the image either to the smaller dimension when you give both `width` and `height`, or to the only given dimension, then crops to the other dimension if needed.
```
stevepiercy, line 265: I rewrote this to clarify what I think it does, but I am not sure. Please double-check and correct me.

docs/classic-ui/images.md
```diff
@@ -237,16 +237,35 @@ image_scale = scaling_util.publishTraverse(context.REQUEST, groups[1])
...
+
+    Deprecated spellings: `scale-crop-to-fit`, `down`.
+
+`cover`
+:   `cover` scales the image either to the larger dimension when you give both `width` and `height`, or to the only given dimension, up to the size you specify.
```
stevepiercy, line 270: I rewrote this to clarify what I think it does, but I am not sure. Please double-check and correct me.

docs/classic-ui/images.md
stevepiercy, line 1: For the deprecated options, when did they change?

stevepiercy (dismissed):
@jensens I pushed a commit instead of did a review because it was easier than nitpicking formatting and grammar. Nonetheless, I have some questions, and I would appreciate your careful review. Thank you!

MrTango:
@stevepiercy plone.scale and plone.namedfile are both still used. In fact they call each other ;).
plone.scale is more low level and plone.namedfiles provideds some additional browser views with utility functions for imagehandling. And to confuse you a bit more, in classic UI we have plone.outputfilter which uses them to create scales when a page is first time rendered and stores them in annotations.

Regarding the deprecated namings, they where changed first in one package, but not in the other places. But it's safe to say that it came with Plone 6.

MrTango (dismissed):

stevepiercy:
Thanks for the clarification and approval, @MrTango.

mauritsvanrees (changes requested):
In general, this is good, thanks.

But one important thing should change: the `direction` parameter is deprecated, and you should use `mode`.
I can never remember this myself, and always have to look it up in the code:

* `plone.scale` [calls](https://github.com/plone/plone.scale/blob/4.1.1/plone/scale/scale.py#L442) `mode = get_scale_mode(mode, direction)`.  That function [warns](https://github.com/plone/plone.scale/blob/4.1.1/plone/scale/scale.py#L206-L212) when you use the `direction` parameter.
* In `plone.namedfile`, `direction` is only supported in the [`create_scale` method](https://github.com/plone/plone.namedfile/blob/6.2.3/plone/namedfile/scaling.py#L318-L325), where you again get a warning when you use the old `direction` parameter.

stevepiercy:
@mauritsvanrees thanks. Fixed in [85aa436](https://github.com/plone/documentation/pull/1624/commits/85aa43615e58a281589d13bb3e4a6651b5ef86a1), as well as the code example and updated the admonition in https://github.com/plone/documentation/pull/1624/commits/0363feb2073c1b06d8b9d0a1c98217359cc7ce8a.

mauritsvanrees (approved):
Thanks, LGTM now.

stevepiercy:
Thanks everyone! I appreciate it.

PR #1627, opened 2024-02-17
Fix yarn installation

sneridagh:


docs/install/create-project.md
```diff
@@ -134,15 +134,14 @@ For the `fish` shell, see [`nvm.fish`](https://github.com/jorgebucaran/nvm.fish)
 
 (install-prerequisites-yeoman-label)=
 
-#### Yeoman
+#### Yeoman and the Volto boilerplate generator
 
 Install {term}`Yeoman`.
```
stevepiercy, line 139:
```suggestion
Install {term}`Yeoman` and the Volto boilerplate generator.
```

docs/install/create-project.md
```diff
@@ -155,28 +154,11 @@ Install the latest Yarn 3 version (not the Classic 1.x one).
...
-    ```shell
-    yarn set version 3.x
-    yarn -v
-    ```
-    ```console
-    3.2.3
-    ```
-
-    If you see a version that is not v3.x.x, you can try deleting the {file}`yarn.lock` file, and installing again.
+2.  Set the yarn version to stable.
```
stevepiercy, line 157:
```suggestion
2.  Set the Yarn version to `stable`.
```

docs/install/create-project.md
```diff
@@ -155,28 +154,11 @@ Install the latest Yarn 3 version (not the Classic 1.x one).
...
-    rm yarn.lock
-    corepack enable
-    yarn set version 3.x
-    yarn -v
+    yarn set version stable
     ```
-    ```console
-    3.2.3
-    ```
-
```
stevepiercy, line 162: The above command installs Yarn 4.1.0. Are we OK with that?

We should also add a step to verify the minimum supported version.

````suggestion

Verify the installed version of Yarn is 3 or later.

```shell
yarn -v
```
```console
4.1.0
```

````

File: docs/install/create-project.md [hunk 4]
```diff
@@ -134,15 +134,14 @@ For the `fish` shell, see [`nvm.fish`](https://github.com/jorgebucaran/nvm.fish)
...
 
 ```shell
-npm install -g yo
+npm install -g yo @plone/generator-volto
 ```
 
-
 (install-prerequisites-yarn-label)=
 
 #### Yarn 3
```
stevepiercy, line 147: Yarn 3 or 4 or just Yarn?

stevepiercy: Also check line 149. I think we can drop the parenthetical statement by now.

stevepiercy:
`````suggestion
#### Yarn

````important
These instructions will not work if you have used another package manager, such as Homebrew on macOS, to install Yarn.
You can verify where you installed Yarn and its version.

```shell
which yarn
```
```console
/opt/homebrew/bin/yarn
```

```shell
yarn -v
```
```console
3.2.3
```

If the console includes `homebrew` in the path, and the version of Yarn is not supported, then you must uninstall it.

```shell
brew uninstall yarn
```

Now the instructions should work.
````
`````

stevepiercy (changes requested):
More clean up may be necessary to either include or remove Yarn version numbers.

stevepiercy:
You can ignore the linkcheck failures. More sites are blocking HEAD requests for URLs, and now it appears SO has jumped on that bandwagon. 😠

sneridagh:
@stevepiercy let's start again, shall we?

PR #1628, opened 2024-02-17
Fix yarn installation

sneridagh:
let's see now.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -289,6 +289,9 @@ Corepack
     Corepack is distributed by default with all recent Node.js versions.
     Run `corepack enable` to install the required Yarn and pnpm binaries on your path.
 
+Git
+    [Git](https://git-scm.com/) is a free and open source distributed version control system.
+
 Hydration
     After loading an HTML page generated with {term}`SSR` in the browser, React can populate the existing {term}`DOM` elements, and recreate and attach their coresponding components.
 
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -280,7 +280,14 @@ mrs-developer
     As a byproduct of its update operations, it also automatically adjusts `jsconfig.json`, which is used by Volto to configure webpack aliases.
 
 Yarn
-    [Yarn](https://yarnpkg.com/) is a JavaScript package manager.
+    [Yarn](https://yarnpkg.com/) is both a JavaScript package manager and project manager.
+
+Corepack
+    [Corepack](https://github.com/nodejs/corepack) is a zero-runtime-dependency Node.js script that acts as a bridge between Node.js projects and the package managers they are intended to be used with during development.
+    In practical terms, Corepack lets you use {term}`Yarn`, {term}`npm`, and {term}`pnpm` without having to install them.
+
+    Corepack is distributed by default with all recent Node.js versions.
+    Run `corepack enable` to install the required Yarn and pnpm binaries on your path.
 
 Hydration
     After loading an HTML page generated with {term}`SSR` in the browser, React can populate the existing {term}`DOM` elements, and recreate and attach their coresponding components.
--- a/docs/install/create-project.md
+++ b/docs/install/create-project.md
@@ -11,7 +11,7 @@ myst:
 (create-a-project-label)=
 
 # Create a project
- 
+
 This chapter describes how you can create a web application project using Plone, with full control over development and deployment.
 
 If instead you want to contribute to a Plone package, see {doc}`/contributing/index`.
@@ -62,6 +62,7 @@ To avoid RAM and disk swap limitations, we recommend either temporarily resizing
 -   {term}`Node.js` LTS 20.x
 -   {term}`Yeoman`
 -   {term}`Yarn`
+-   {term}`git`
 -   {term}`GNU make`
 -   {term}`Docker`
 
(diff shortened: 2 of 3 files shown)


sneridagh:
@stevepiercy I tested it in a plain ubuntu docker container, and it works as expected. I also noticed that `git` is a requirement too. I also noticed a thing, putting the comment side by side to the code.

sneridagh:
@stevepiercy meh, I can't... so:

https://github.com/plone/documentation/pull/1628/files#diff-c5520082ca7976f8c1d2ec06e4fe7f137c9a1fd6b4032cbde75c433a0aa83a7aR102-R106

the mention to `~/bash_profile` this is a mac thing, not a Debian/Ubuntu practice nowadays. So it does not work there... Could you fix it to be more generic? I guess mention the nvm installation is fine, or just allow people to restart the current session.

sneridagh:
btw: https://community.plone.org/t/plone-6-volto-generate-app-cannot-find-module-is-ci/15886/3

docs/install/create-project.md
```diff
@@ -134,49 +135,48 @@ For the `fish` shell, see [`nvm.fish`](https://github.com/jorgebucaran/nvm.fish)
...
+Use {term}`Corepack` to enable Yarn, which was already installed with the {ref}`supported version of Node.js <install-packages-prerequisites-label>`.
 
 1.  Open a terminal and type:
 
     ```shell
     corepack enable
     ```
 
-2.  Verify that Yarn v3.x.x is installed and activated.
+In your generated Volto project from the previous step, you can find the pinned version of Yarn in its {file}`.yarnrc.yml`.
```
stevepiercy: @sneridagh can you verify my guess about how Yarn is pinned? This may be one of those things that seasoned JS developers "just know", but is a mystery to Python developers.

Also should we do a `yarn -v` to get the enabled version of Yarn, then tell the reader to compare that against what exists in the pin?

docs/install/create-project.md
```diff
@@ -62,6 +62,7 @@ To avoid RAM and disk swap limitations, we recommend either temporarily resizing
 -   {term}`Node.js` LTS 20.x
 -   {term}`Yeoman`
 -   {term}`Yarn`
+-   {term}`git`
```
stevepiercy: If we make `git` a requirement, then we also need to provide a new section and reference to how to install it. We can use a tabbed interface for various OSes using `sphinx-design`, or just give a link to it. I'm more inclined to just a link because developers have probably already installed git at some point in their lives.

stevepiercy (changes requested):
I pushed a few commits, and added some questions in my review. Would you please take a look and respond when you get a moment?

After our chat yesterday, it occurred to me that it might encourage the "Python recidivist" to try Volto would be to describe for what does Volto use each of the tools that are required for installation? Although we already define them in the Glossary by pulling the descriptions of what they are and do from their authoritative documentation or marketing, we don't give them context within Volto. I think a good spot for that explanatory text might be https://6.docs.plone.org/volto/contributing/developing-core.html.

stevepiercy:
@sneridagh I think we also need to update https://6.docs.plone.org/volto/contributing/developing-core.html to get it in synch with this pull request. I'll create a separate PR for that and we discuss details there.

docs/install/create-project.md
stevepiercy: Done in https://github.com/plone/volto/pull/5769 and [07f76da](https://github.com/plone/documentation/pull/1628/commits/07f76dab395be4c561f5d73b225e8fc3caca56c6)

stevepiercy:
@sneridagh here's the other PR, which needs to be reviewed and merged before this one can be merged: https://github.com/plone/volto/pull/5769

stevepiercy:
We can ignore the SO linkcheck failures here for now. The Sphinx linkcheck builder might address this in version 5 or greater.

docs/install/create-project.md
sneridagh: I don't think that's necessary, it's pinned. Do you test if a pinned version in requirements.txt is the one that is really installed?

sneridagh: Regarding the placement of the pin, it's in the `package.json` as well. `packageManager` key.

```
"packageManager": "yarn@3.2.3",
```

docs/glossary.md
```diff
@@ -280,7 +280,14 @@ mrs-developer
     As a byproduct of its update operations, it also automatically adjusts `jsconfig.json`, which is used by Volto to configure webpack aliases.
 
 Yarn
-    [Yarn](https://yarnpkg.com/) is a JavaScript package manager.
+    [Yarn](https://yarnpkg.com/) is both a JavaScript package manager and project manager.
+
+Corepack
+    [Corepack](https://github.com/nodejs/corepack) is a zero-runtime-dependency Node.js script that acts as a bridge between Node.js projects and the package managers they are intended to be used with during development.
+    In practical terms, Corepack lets you use {term}`Yarn`, {term}`npm`, and {term}`pnpm` without having to install them.
```
sneridagh, line 287: Do we have an `pnpm` term?

docs/install/create-project.md
stevepiercy: > I don't think that's necessary, it's pinned. Do you test if a pinned version in requirements.txt is the one that is really installed?

That assumes everything goes right. When something goes wrong, such as the Homebrew conflict, I would compare what is pinned against what is actually used. I can remove the training wheels, and leave developers to the subsequent admonition.

Interestingly, when I ran through the installation docs:

```shell
$ yarn -v
1.22.19
$ which yarn                                                    
/Users/stevepiercy/.yarn/bin/yarn
```

Why? I have no idea. It's some remnant from 2022. So checking the system version of Yarn is misleading for a comparison anyway, as everything still works.

Fixed in cf167b78568ad21c93b787bfbb0f6f38cb10a692.

docs/glossary.md
stevepiercy: Yup, near the end of the Glossary, which automatically sorts on rendering to HTML.

```markdown
pnpm
    [pnpm](https://pnpm.io/) is a fast, disk space efficient package manager.
```

stevepiercy:
@sneridagh one more review pass, and I would appreciate your thoughts about me creating a separate issue or PR for my comment in https://github.com/plone/documentation/pull/1628#pullrequestreview-1887051017. We say "use this package manager", but we don't give more explanation. Now with 3 (or is it 4?) package managers in Volto, I think context is helpful.

sneridagh:
@stevepiercy https://github.com/plone/documentation/pull/1628/commits/cf167b78568ad21c93b787bfbb0f6f38cb10a692

LGTM

PR #1629, opened 2024-02-22
Add predicate to glossary

stevepiercy:
See https://github.com/plone/volto/pull/5775. This should be merged before the other so that its docs build without an error for its new term.


sneridagh (approved):
🎉

PR #1632, opened 2024-02-29
Fix for plone/cookiecutter-zope-instance 2.x: naming zcml variables

ksuess:
See https://github.com/plone/cookiecutter-zope-instance?tab=readme-ov-file#upgrading

Fix needed. Otherwise the build (`make build-backend`) fails.

### DIFF
--- a/docs/install/manage-add-ons-packages.md
+++ b/docs/install/manage-add-ons-packages.md
@@ -173,8 +173,7 @@ collective.easyform
 
 Add it to {file}`instance.yaml` to let Zope know that this add-on should be loaded:
 
-```{code-block} yaml
-:emphasize-lines: 3-6
+```yaml
 default_context:
     zcml_package_includes: project_title, collective.easyform
 ```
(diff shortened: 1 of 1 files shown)


stevepiercy (approved):
Thank you! Will merge right away.

stevepiercy:
I'll fix the broken links in another PR.

jensens:
No idea how its done best, but we need version-management for cookiecutters.
The call supports option `cookiecutter --checkout=1.0.1 ...`  for those who do not want the new 2.0.0 major release.

stevepiercy:
@jensens are you talking about adding support for that call in the cookiecutter itself or in documentation, or both?

We could add a note to the documentation under [Install Plone 6](https://6.docs.plone.org/install/create-project.html#install-plone-6), but the assumption is that the user is creating a new project and should use the latest supported versions of everything. I think adding a note here would add confusion.

But in management of existing projects, we have a note to see the readme for [Configuration with `cookiecutter-zope-instance`](https://6.docs.plone.org/install/manage-add-ons-packages.html#configuration-with-cookiecutter-zope-instance). Would that be a good place to add a note about specifying the version of this cookiecutter?

jensens:
We should change cookiecutter-plone-starter to pin it and update the Makefile snippet accordingly.

davisagli:
@jensens I added an issue over there: https://github.com/collective/cookiecutter-plone-starter/issues/123

PR #1636, opened 2024-03-07
Do not call virtualenv --clear in our Makefile.

### LINKED ISSUE #1635: `make livehtml` removed the current directory

_No response_

### Description

I created a documentation PR last week, and now wanted to continue on it:

```
$ make livehtml 
python3 -m venv . || virtualenv --clear --python=python3 .
Error: [Errno 2] No such file or directory: '/Users/maurits/community/plone-coredev/6.0/documentation/docs/bin/python3'
created virtual environment CPython3.11.8.final.0-64 in 4533ms
  creator CPython3Posix(dest=/Users/maurits/community/plone-coredev/6.0/documentation/docs, clear=True, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, wheel=bundle, setuptools=bundle, via=copy, app_data_dir=/Users/maurits/Library/Application Support/virtualenv)
    added seed packages: pip==21.2.3, setuptools==57.4.0, wheel==0.37.0
  activators PythonActivator,CShellActivator,FishActivator,PowerShellActivator,BashActivator
bin/pip install -r requirements-initial.txt
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory
bash: line 1: bin/pip: No such file or directory
make: *** [bin/python] Error 127
```

So for whatever reason (no idea), `python3 -m venv .` failed, so the Makefile fell back to calling `virtualenv --clear`.  Since a few years, `--clear` actually removes the target directory.  So my checkout of the documentation repo was gone.

No worries: no work was lost.
Still, I would like to avoid this. :-D

Two possible solutions, for [this line in the Makefile](https://github.com/plone/documentation/blob/6.0/Makefile#L38):

1. Remove the `--clear` option.
2. Never call `virtualenv`, only use `python3 -m venv .`.  I suppose the `virtualenv` fallback was only there to support Python versions that did not yet have the `venv` module.  This should not be an issue anymore.

I would be happy to make a PR if wanted, but it would be good to know the preference first.

mauritsvanrees:
This would remove the current directory if called as fallback.

This fixes https://github.com/plone/documentation/issues/1635

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ distclean:  ## Clean docs build directory and Python virtual environment
 
 
 bin/python:
-	python3 -m venv . || virtualenv --clear --python=python3 .
+	python3 -m venv .
 	bin/pip install -r requirements-initial.txt
 	bin/pip install -r requirements.txt
 


stevepiercy (approved):
Thank you!

PR #1637, opened 2024-03-08
feat(upgrade): mention the boolean fields changes

gforcada:
See the context in https://github.com/zopefoundation/zope.schema/issues/104

### DIFF
--- a/docs/backend/upgrading/version-specific-migration/upgrade-to-60.md
+++ b/docs/backend/upgrading/version-specific-migration/upgrade-to-60.md
@@ -653,5 +653,6 @@ It renders the portlets from the `plone.footerportlets` portlet manager.
 
 ## Boolean fields
 
-Since `zope.schema==6.1.0` all `zope.schema.Bool` fields need to have a `required=False` attribute,
+Since `zope.schema==6.1.0`, all `zope.schema.Bool` fields must have a `required=False` attribute.
+This allows you to either tick or not tick the checkbox, submit the form, and process the field with either its value when ticked or `None` when unticked.
 otherwise, you will not be able to save the form without marking the checkbox, which effectively turns that field to be always `True`.
--- a/docs/backend/upgrading/version-specific-migration/upgrade-to-60.md
+++ b/docs/backend/upgrading/version-specific-migration/upgrade-to-60.md
@@ -650,3 +650,8 @@ Plone 6.0 makes changes to two viewlet managers:
 One final change is that Plone 6.0 moves the `plone.footer` viewlet from `plone.app.layout/viewlets` to `plone.app.portlets`.
 The viewlet remains in manager `plone.portalfooter`.
 It renders the portlets from the `plone.footerportlets` portlet manager.
+
+## Boolean fields
+
+Since `zope.schema==6.1.0` all `zope.schema.Bool` fields need to have a `required=False` attribute,
+otherwise, you will not be able to save the form without marking the checkbox, which effectively turns that field to be always `True`.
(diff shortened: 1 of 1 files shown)


docs/backend/upgrading/version-specific-migration/upgrade-to-60.md
```diff
@@ -650,3 +650,8 @@ Plone 6.0 makes changes to two viewlet managers:
 One final change is that Plone 6.0 moves the `plone.footer` viewlet from `plone.app.layout/viewlets` to `plone.app.portlets`.
 The viewlet remains in manager `plone.portalfooter`.
 It renders the portlets from the `plone.footerportlets` portlet manager.
+
+## Boolean fields
+
+Since `zope.schema==6.1.0` all `zope.schema.Bool` fields need to have a `required=False` attribute,
+otherwise, you will not be able to save the form without marking the checkbox, which effectively turns that field to be always `True`.
```
stevepiercy: I think there is a use case to set `required=True` or not add the attribute `required=False`. For example:

- [ ] I agree to the Terms and Conditions of Service-o-rama.

The user is required to check the box to use Service-o-rama, else the form data would not be saved.

stevepiercy (commented):
Please see review question.

docs/backend/upgrading/version-specific-migration/upgrade-to-60.md
davisagli: At the schema level, `required=True` for a boolean field means that the field must have a value that is boolean (True or False) rather than missing (None).

If some widget library is treating it as meaning that a checkbox based on the field needs to be checked, that's a bug in that library, in my opinion.

For a checkbox that must have its value set to True, the way to do that at the schema level would be with a constraint (`constraint=lambda value: value is True`)

stevepiercy: OK, thanks for the explanation. I couldn't comprehend that issue's long discussion.

docs/backend/upgrading/version-specific-migration/upgrade-to-60.md
```diff
@@ -650,3 +650,8 @@ Plone 6.0 makes changes to two viewlet managers:
 One final change is that Plone 6.0 moves the `plone.footer` viewlet from `plone.app.layout/viewlets` to `plone.app.portlets`.
 The viewlet remains in manager `plone.portalfooter`.
 It renders the portlets from the `plone.footerportlets` portlet manager.
+
+## Boolean fields
+
+Since `zope.schema==6.1.0` all `zope.schema.Bool` fields need to have a `required=False` attribute,
```
stevepiercy:
```suggestion
Since `zope.schema==6.1.0`, all `zope.schema.Bool` fields must have a `required=False` attribute.
This allows you to either tick or not tick the checkbox, submit the form, and process the field with either its value when ticked or `None` when unticked.
```

docs/backend/upgrading/version-specific-migration/upgrade-to-60.md
stevepiercy:
```suggestion
Otherwise, you can't save the form without ticking the checkbox, which effectively makes the field value always `True`.
```

stevepiercy (changes requested):
Added another explanatory sentence, and broke original sentences into two, one per line.

gforcada:
Thanks for the wording improvements, I hope this is now ready to be merged 🤞🏾 🍀

stevepiercy (approved):
Thank you! Will merge after other PR passes CI, then update branch with `main`.

stevepiercy:
GitHub still not well. Merging.

PR #1642, opened 2024-03-17
Add `pnpm workspace` to glossary.

stevepiercy:


stevepiercy:
I'm going to merge this PR when it turns green, and if y'all want to add review comments, I will circle back and make modifications. I just want to get an entry in the glossary for `pnpm workspace` now, so I can continue work on the other PR https://github.com/plone/volto/pull/5882

stevepiercy:
The failure is OK, GitHub is having intermittent issues at the moment. Merging now.

docs/glossary.md
```diff
@@ -701,4 +701,13 @@ Nick
 predicate
 predicates
     In programming, a predicate is a test which returns `true` or `false`.
+
+pnpm workspace
+workspace
+    pnpm has built-in support for monorepositories (also known as multi-package repositories, multi-project repositories, or monolithic repositories).
+    Workspaces provide support to manage multiple packages from your local file system from within a singular top-level, root package.
+    By defining a single `package.json` file at the root of the repository and specifying the individual packages as workspaces, pnpm treats them as interdependent projects.
```
ichim-david, line 709: @stevepiercy 
the packages that are defined as part of the pnpm repo are defined in https://github.com/plone/volto/blob/main/pnpm-workspace.yaml
If you look at package.json there is no reference to any of the packages as dependencies.
Also there is no specification of individual packages as workspace package as we are using the glob syntax to specify all packages from packages/* or apps/*

It's something that we have an entry but I think we need todo some tweaking to it

stevepiercy: Thanks. See https://github.com/plone/documentation/pull/1643. I copy-pasted from a page that had a decent definition of pnpm workspaces because the official pnpm docs don't have a good one.

PR #1643, opened 2024-03-17
Use correct filename for pnpm workspace definition and add glossary terms for Volto

stevepiercy:


stevepiercy:
For this round, I'll defer merging until I get an approval. There may be other changes that need to happen as part of my review of https://github.com/plone/volto/pull/5882

ichim-david (dismissed):

PR #1652, opened 2024-04-19
Initial skeleton of deployment documentation

stevepiercy:
This is the initial skeleton of our shiny new Plone 6 Deployment documentation. 🥳


docs/deployment/components.md
```diff
@@ -0,0 +1,58 @@
...
+
+You need the following components to deploy a Plone application.
+
+-   {term}`Plone backend`
+-   {term}`Volto` as the default {term}`Frontend` for Plone
+-   {term}`TLS termination proxy`
+-   {term}`Load balancer` or router
+-   Database
+-   Mail service
+-   Optimization tasks
```
davisagli: This is not a component. We should have a separate chapter "Optimize and tune your Plone site" which explains how to monitor and tune.

docs/deployment/components.md
```diff
@@ -0,0 +1,58 @@
...
+    -   Database parameters
+    -   https://5.docs.plone.org/manage/deploying/performance/index.html
+-   Maintenance tasks
+    -   Database backup
+    -   Packing database - https://5.docs.plone.org/manage/deploying/packing.html
+    -   Monitoring
```
davisagli: Monitoring should go under recommended or optional ... probably optional since we aren't currently making specific recommendations about how to do it

docs/deployment/components.md
```diff
@@ -0,0 +1,58 @@
...
+    -   Log rotation
+    -   Firewall
+    -   Attack protection
+    -   DNS
+    -   Host name resolution
```
davisagli: Everything from firewall on down is not a maintenance task, but an area of concern that should be addressed when setting up the underlying system that hosts Plone.

I think these are mostly out of scope for the Plone docs, except:
- we can say that you ought to think about them
- in the server-enviroment chapter, we'll show one way to address them

It might make sense to move mentioning them from here to server-environment.

Also, aren't DNS and host name resolution the same thing?

davisagli (commented):
Thanks for getting this started @stevepiercy. I'm just reviewing the high-level organization for now.

docs/deployment/components.md
stevepiercy: Done in e3f4a1666941487da502b02607eb0a33d54ef103

docs/deployment/components.md
stevepiercy: Done in 3d97d4148c04c6cada95a4b2ff19e681f443765e

docs/deployment/components.md
stevepiercy: I can't remember the term where you configure the web server. Is that "listen on IP address" or "bind" or something else?

stevepiercy: Mostly addressed in 9d29c455f12282530f29200a467394da0f80e4b5, but might need some polishing for the web server configuration bit.

stevepiercy:
@sneridagh @ericof can I get reviews from y'all?

sneridagh (approved):
So far, it looks good to me, as a first iteration. Let's continue I'd say.

docs/deployment/components.md
stevepiercy: @davisagli please let me know if this change is OK. I'd like to merge this PR soon to replace the current Deployment skeleton.

ericof (approved):

davisagli (approved):

stevepiercy:
Thank you, everyone! Now the hard part begins...

PR #1661, opened 2024-05-10
Add Sphinx examples extension

stevepiercy:
See also https://github.com/plone/volto/pull/6011

Simplify example MyST syntax with extension in use.


stevepiercy:
We can ignore the failing linkcheck, it was temporary.

davisagli (approved):

PR #1664, opened 2024-05-14
Add supported languages section to i18n and l10n

stevepiercy:
Addresses comment in https://github.com/plone/volto/pull/6016#pullrequestreview-2054446691


erral (approved):

stevepiercy:
@davisagli does this PR align with what you had in mind from your comment?

davisagli (approved):
@stevepiercy This is good, but what I had in mind is something more for a newcomer to Plone, rather than a developer, so they are aware of what languages are available.

stevepiercy:
@davisagli that sounds like something for plone.org, or maybe the user manuals? Can you create an issue in the appropriate repos for what you want?

PR #1685, opened 2024-07-23
Opsworks was EOL on May 26, 2024

stevepiercy:
See https://aws.amazon.com/blogs/mt/seamlessly-off-board-from-aws-opsworks-stacks-by-detaching-resources/

----
📚 Documentation preview 📚: https://plone6--1685.org.readthedocs.build/


stevepiercy:
This reference was never used in docs, but it was used in 2022 training.

PR #1713, opened 2024-09-25
Disable VSCode's format-on-save feature for markdown

davisagli:
This should help avoid contributors accidentally reformatting (the default formatter makes changes we don't want, like converting to the wrong quotes)

----
📚 Documentation preview 📚: https://plone6--1713.org.readthedocs.build/


stevepiercy (approved):
Heck yeah! We should port this to Volto, plone.restapi, and plone.api submodules as well.

PR #1743, opened 2024-10-25
Fix broken links

stevepiercy:
----
📚 Documentation preview 📚: https://plone6--1743.org.readthedocs.build/

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -523,7 +523,7 @@ manual `.po` entries
     Entries which cannot be detected by an automatic code scan.
 
 react-intl
-    A library that is part of [Format.JS](https://formatjs.github.io/docs/getting-started/installation) which helps developers set up their applications for internationalization.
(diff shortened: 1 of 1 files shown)


davisagli (changes requested):
@stevepiercy I just changed this in https://github.com/plone/documentation/pull/1741 to fix a linkcheck error. Is it redirecting inconsistently?

stevepiercy:
The advice in linkcheck was wrong, as it only detects the response from the server. I manually researched the new link based on the existing link.

- Original: https://formatjs.io/docs/getting-started/installation
- Current: https://formatjs.github.io/docs/getting-started/installation

It could be argued that the better link would be to the root of the site instead of how to install it. We actually need to improve this Glossary entry, now that I think about it. Update coming.

davisagli:
@stevepiercy Oh I see, I used the new redirect but it didn't go to the same content as before. Mea culpa.

stevepiercy:
@davisagli I think we're good now. One more review?

davisagli (approved):

PR #1760, opened 2024-11-06
Improve install docs

MrTango:
This corrects the creation of Python venv. We shoud never mix the venv with other configuration like the buildout configuration or others. This can lead to losing data or configuration when updating the venv later.

so instead of 

`python3 -m venv .`

we should  use 

`python3 -m venv venv`

or similar to create a venv.
Then either activate the venv or use it like this

`./venv/bin/pip install Plone`

The buildout bootstrap command is useful to avoid the need to always prefix the buildout run with the venv path later.
It will create the `bin/buildout` script and only need to run once.

### DIFF
--- a/docs/admin-guide/index.md
+++ b/docs/admin-guide/index.md
@@ -11,7 +11,7 @@ myst:
 
 # Admin guide
 
-In this part of the documentation, you can find how to install, operate, configure, and deploy Plone.
+In this part of the documentation, you can find how to install, operate and configure Plone.
 
 
 ```{toctree}
--- a/docs/admin-guide/index.md
+++ b/docs/admin-guide/index.md
@@ -11,7 +11,7 @@ myst:
 
 # Admin guide
 
-In this part of the documentation, you can find how to install, operate and configure Plone.
+In this part of the documentation, you can find how to install, operate, and configure Plone.
 
 
 ```{toctree}
--- a/docs/admin-guide/install-buildout.md
+++ b/docs/admin-guide/install-buildout.md
@@ -25,7 +25,7 @@ For other installation options, see {ref}`get-started-install-label`.
 ## Prerequisites for installation
 
 -   For Plone 6.0, Python {SUPPORTED_PYTHON_VERSIONS_PLONE60}
-% TODO: These instructions install Plone 6.0.x. Uncomment next line and change the subsequent include when Plone 6.1 is released and "latest". 
+% TODO: These instructions install Plone 6.0.x. Uncomment next line and change the subsequent include when Plone 6.1 is released and "latest".
 % -   For Plone 6.1, Python {SUPPORTED_PYTHON_VERSIONS_PLONE61}
 
 
@@ -47,14 +47,16 @@ cd <my_projects>/plone
 Create a Python virtual environment.
 
 ```shell
-python3 -m venv .
+python3 -m venv venv
 ```
 
 Install the minimal Python packages needed in order to run Buildout.
 
 ```shell
-bin/pip install -r https://dist.plone.org/release/6-latest/requirements.txt
+venv/bin/pip install -r https://dist.plone.org/release/6-latest/requirements.txt
 ```
+
+
 
 Create a {file}`buildout.cfg` file in your directory with the following contents.
 
@@ -75,6 +77,12 @@ eggs =
     Plone
 ```
 
+Install Buildout script in bin directory.
+
+```shell
+venv/bin/buildout bootstrap
+```
+
 Run Buildout.
 
 ```shell
@@ -83,6 +91,7 @@ bin/buildout
 
 This may take a few minutes.
 
+When ever the buildout configuration has changed, re-run ./bin/buildout
 
 ## Start Plone in foreground mode
 
--- a/docs/admin-guide/install-buildout.md
+++ b/docs/admin-guide/install-buildout.md
@@ -77,7 +77,7 @@ eggs =
     Plone
 ```
 
-Install Buildout script in bin directory.
+Use Buildout's [`bootstrap` command](https://www.buildout.org/en/latest/topics/bootstrapping.html) to install a local `buildout` script in the {file}`bin` directory.
 
 ```shell
 venv/bin/buildout bootstrap
@@ -91,7 +91,7 @@ bin/buildout
 
 This may take a few minutes.
 
-When ever the buildout configuration has changed, re-run ./bin/buildout
+Whenever you change the Buildout configuration, run `./bin/buildout` again.
 
 ## Start Plone in foreground mode
 
--- a/docs/admin-guide/install-pip.md
+++ b/docs/admin-guide/install-pip.md
@@ -53,7 +53,7 @@ python3 -m venv venv
 Install Plone and a helper package, {term}`pipx`.
 
 ```shell
-./venv/bin/pip install -c https://dist.plone.org/release/6.0-latest/constraints.txt Plone pipx
+venv/bin/pip install -c https://dist.plone.org/release/6.0-latest/constraints.txt Plone pipx
 ```
 
 
--- a/docs/admin-guide/install-pip.md
+++ b/docs/admin-guide/install-pip.md
@@ -25,7 +25,7 @@ For other installation options, see {ref}`get-started-install-label`.
 ## Prerequisites for installation
 
 -   For Plone 6.0, Python {SUPPORTED_PYTHON_VERSIONS_PLONE60}
-% TODO: These instructions install Plone 6.0.x. Uncomment next line and change the subsequent include when Plone 6.1 is released and "latest". 
+% TODO: These instructions install Plone 6.0.x. Uncomment next line and change the subsequent include when Plone 6.1 is released and "latest".
 % -   For Plone 6.1, Python {SUPPORTED_PYTHON_VERSIONS_PLONE61}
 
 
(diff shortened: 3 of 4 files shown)


docs/admin-guide/index.md
```diff
@@ -11,7 +11,7 @@ myst:
 
 # Admin guide
 
-In this part of the documentation, you can find how to install, operate, configure, and deploy Plone.
+In this part of the documentation, you can find how to install, operate and configure Plone.
```
davisagli: @MrTango Deployment belongs in the admin guide, we just haven't gotten to moving it yet. I'm okay with this edit in the meantime.

docs/admin-guide/install-buildout.md
```diff
@@ -47,15 +47,17 @@ cd <my_projects>/plone
 Create a Python virtual environment.
 
 ```shell
-python3 -m venv .
+python3 -m venv venv
```
davisagli, line 50: Why? This requires more typing when running commands.

docs/admin-guide/install-buildout.md
```diff
@@ -75,21 +77,28 @@ eggs =
...
+
+```shell
+./venv/bin/buildout bootstrap
+```
+
 Run Buildout.
 
 ```shell
-bin/buildout
+./bin/buildout
```
davisagli: Why? It's more typing and unnecessary.

davisagli (changes requested):

docs/admin-guide/install-buildout.md
```diff
@@ -75,21 +77,28 @@ eggs =
     Plone
 ```
 
+Install Buildout script in bin directory.
+
+```shell
+./venv/bin/buildout bootstrap
```
davisagli: This isn't needed. Buildout was already installed as part of the `pip install` command on line 56.

davisagli (changes requested):

docs/admin-guide/index.md
MrTango: ok, didn't know that. If we have it in the admin guide, then probably still not mixed with installation right.
As with all the options and details, it get;s a bit to much.

docs/admin-guide/install-buildout.md
MrTango: it does not install buildout, but it creates the bin/buildout script, so that you don't need to prefix the buildout run with the venv folder.all the time.

docs/admin-guide/install-buildout.md
MrTango: in my experience it works better than without. Sometimes on Linux calling a script without `./` will not find it.
Thats why i always use it, but I'm ok if we leave this out.

docs/admin-guide/install-buildout.md
MrTango: you will lose data if you mix it like this.
Only happens since Python3, but i did see this several times when the venv was rebuild with `--clear`, which you have to do when the Python binary changed. I was using the mixed version for years, but in Python3 this is dangerous.

docs/admin-guide/index.md
davisagli: See the landing page https://6.docs.plone.org/admin-guide/index.html -- there are different categories. But we want a pretty flat structure in the navigation so that people can see what's there at a glance without guessing what might be nested.

docs/admin-guide/install-buildout.md
davisagli: I can see how it makes it harder to delete and rebuild the virtualenv. I agree with recommending a separate venv folder.

docs/admin-guide/install-buildout.md
davisagli: I don't feel strongly about it, but I don't think I ever had problems with this.

docs/admin-guide/install-buildout.md
davisagli: Right, so it's necessary (or at least convenient) if the virtualenv is in a different location.

The other option would be activating the virtualenv, but that's always a can of worms to explain to people, so let's go with your way.

docs/admin-guide/run-plone.md
```diff
@@ -18,7 +18,7 @@ There are different commands to run Plone, depending on which method you used to
 ## Run Plone in foreground mode
 
 Running Plone in foreground mode will show output in the terminal.
-This is recommended while developing a Plone site.
+This is recommended while developing a Plone site. The command to be used, depends on the used install method.
```
davisagli:
```suggestion
This is recommended while developing a Plone site. The command to be used depends on the install method you used.
```

davisagli: @stevepiercy help with wording please...my suggestion is also awkward

davisagli (dismissed):
Basically LGTM now.

docs/admin-guide/index.md
stevepiercy: Just to add to this conversation, there is a longer term goal to implement Guides as shown in https://pydata-sphinx-theme.readthedocs.io/en/stable/ and https://numpy.org/doc/stable/release.html in their top horizontal navigation. We're in the first phase of reorganization, and that may be part of a second or third phase.

docs/admin-guide/install-buildout.md
stevepiercy: I think it is only required when trying to run a script in the current working directory, but not a binary. In other words, the following example fails:

```
> script
```

Whereas the following will succeed.

```
> ./script
```

However, as long as a path to the script is specified—whether that is `./script`, `dir1/script`, or `dir1/dir2/script`—then it will be invoked.

More [discussion from SO](https://stackoverflow.com/a/6331085/2214933) explains how `/` and `$PATH` work together.

As far as in this example, I would remove `./` because there is a `/` in `bin/buildout`, making `./` optional.

stevepiercy:
> This corrects the creation of Python venv. We shoud never mix the venv with other configuration like the buildout configuration or others. This can lead to losing data or configuration when updating the venv later.
> 
> so instead of
> 
> `python3 -m venv .`
> 
> we should use
> 
> `python3 -m venv venv`
> 
> or similar to create a venv. Then either activate the venv or use it like this
> 
> `./venv/bin/pip install Plone`
> 
> The buildout bootstrap command is useful to avoid the need to always prefix the buildout run with the venv path later. It will create the `bin/buildout` script and only need to run once.

I just want to say that I really appreciate this explanation. I have long felt that creating Python virtual environments at the project root was "not right", both in Plone documentation and our tools. Coming from Pyramid, I always used `env/bin/thingie` but never activated it. In their docs, they used an additional step as demonstrated in [Set an environment variable](https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/requirements.html#set-an-environment-variable) and explained in [Why use `$VENV/bin/pip` instead of source `bin/activate`, then `pip`?](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/install.html#venv-bin-pip-vs-source-bin-activate). In practice, I felt that last bit was too much, because anything that requires me to hold down the SHIFT key slows down my typing.

docs/admin-guide/index.md
stevepiercy: Oxford comma

```suggestion
In this part of the documentation, you can find how to install, operate, and configure Plone.
```

docs/admin-guide/install-buildout.md
```diff
@@ -75,6 +77,12 @@ eggs =
     Plone
 ```
 
+Install Buildout script in bin directory.
```
stevepiercy: Because "Buildout" is the package, whereas `buildout` is the script, which tripped up both me and @davisagli.

```suggestion
Use Buildout's [`bootstrap` command](https://www.buildout.org/en/latest/topics/bootstrapping.html) to install a local `buildout` script in the {file}`bin` directory.
```

docs/admin-guide/install-buildout.md
```diff
@@ -83,6 +91,7 @@ bin/buildout
 
 This may take a few minutes.
 
+When ever the buildout configuration has changed, re-run ./bin/buildout
```
stevepiercy: Change to active voice, because the configuration doesn't change on its own.

```suggestion
Whenever you change the Buildout configuration, run `./bin/buildout` again.
```

docs/admin-guide/install-pip.md
```diff
@@ -47,13 +47,13 @@ cd <my_projects>/plone
...
 ```shell
-python3 -m venv .
+python3 -m venv venv
 ```
 
 Install Plone and a helper package, {term}`pipx`.
 
 ```shell
-bin/pip install -c https://dist.plone.org/release/6.0-latest/constraints.txt Plone pipx
+./venv/bin/pip install -c https://dist.plone.org/release/6.0-latest/constraints.txt Plone pipx
```
stevepiercy:
```suggestion
venv/bin/pip install -c https://dist.plone.org/release/6.0-latest/constraints.txt Plone pipx
```

docs/admin-guide/run-plone.md
```diff
@@ -18,7 +18,7 @@ There are different commands to run Plone, depending on which method you used to
 ## Run Plone in foreground mode
 
 Running Plone in foreground mode will show output in the terminal.
-This is recommended while developing a Plone site.
+This is recommended while developing a Plone site. The command to be used depends on the install method you used.
```
stevepiercy:
```suggestion
This is recommended while developing a Plone site.
The command you use depends on the installation method you used.
```

stevepiercy (changes requested):
Minor tweaks to grammar, sprinkle some MyST syntax and reference links, and reword a few things.

stevepiercy (approved):
Thanks @MrTango and @davisagli! I'll merge this as soon as it's green.

PR #1806, opened 2024-12-05
add "Customizing TinyMCE" chapter to Classic UI docs

szakitibi:
## First-time contributors

You **must** read and follow our [First-time contributors](https://6.docs.plone.org/contributing/first-time.html).

---

## Submit a pull request

Thank you for your contribution to the Plone Documentation.

Before submitting this pull request, please make sure you follow our guides:

-   [Contributing to Plone documentation](https://6.docs.plone.org/contributing/index.html)
-   [Building and checking the quality of documentation](https://6.docs.plone.org/contributing/setup-build.html)
-   [Authors guide](https://6.docs.plone.org/contributing/authors.html)
-   [MyST reference](https://6.docs.plone.org/contributing/myst-reference.html)

## Issue number

- Fixes #

## Description

_Write a description of the fixes or improvements._

## Add screenshots or links to a preview of the changes

----
📚 Documentation preview 📚: https://plone6--1806.org.readthedocs.build/

PR #1807, opened 2024-12-05
Tinymce customization

stevepiercy:
Closes #1806.

----
📚 Documentation preview 📚: https://plone6--1807.org.readthedocs.build/

### DIFF
--- a/docs/classic-ui/tinymce-customization.md
+++ b/docs/classic-ui/tinymce-customization.md
@@ -13,3 +13,34 @@ myst:
 
 This chapter is a developer reference manual for customizing {term}`TinyMCE`.
 
+
+(tinymce-remove-formats-label)=
+
+## Remove formats
+
+In Plone 6, TinyMCE has a special logic that automatically reads registered files named {file}`tinymce-formats.css` and adds the CSS classes defined in those files to TinyMCE's {menuselection}`Format --> Formats` menu.
+
+Plone 6 Classic UI ships with the Barceloneta theme which includes two custom formats,`highlight-inline` and `p.highlight-paragraph`, in the TinyMCE {menuselection}`Format --> Formats` menu.
+You can remove these formats through the TinyMCE control panel.
+
+1.  Navigate to {menuselection}`Site Setup --> TinyMCE --> Default`, or append `@@tinymce-controlpanel` to the root of your website in your browser's location bar.
+1.  Scroll down to {guilabel}`Choose the CSS used in WYSIWYG Editor Area`.
+1.  Remove the entry `++theme++barceloneta/tinymce/tinymce-formats.css`.
+1.  Click the {guilabel}`Save` button.
+
+Once removed, the custom formats will no longer appear in the menu.
+
+
+## Add formats
+
+To add custom formats, you can provide your own files.
+The file must satisfy the following requirements.
+
+1.  named {file}`tinymce-formats.css`
+1.  registered as a resource in your Plone site
+1.  included in the TinyMCE control panel in the textarea input {guilabel}`Choose the CSS used in WYSIWYG Editor Area`.
+    See {ref}`tinymce-remove-formats-label`.
+
+CSS styles defined in this file will automatically be added to the top level of TinyMCE's {menuselection}`Format --> Formats` menu.
+
+Alternatively, you can customize TinyMCE by adding the styles through the [`formats` option in the JSON configuration](https://www.tiny.cloud/docs/tinymce/7/content-formatting/).
--- a/docs/classic-ui/tinymce-customization.md
+++ b/docs/classic-ui/tinymce-customization.md
@@ -14,18 +14,19 @@ myst:
 This chapter is a developer reference manual for customizing {term}`TinyMCE`.
 
 
-(tinymce-remove-formats-label)=
+## Customize {menuselection}`Format` menu
 
-## Customize format menu
+There are several options to customize TinyMCE's {menuselection}`Format --> Formats` menu.
 
-There are two options to customize TinyMCEs {menuselection}`Format --> Formats` menu.
+The first two options use TinyMCE's [`formats` JSON configuration](https://www.tiny.cloud/docs/tinymce/latest/content-formatting/).
+They are less complicated to implement than the third option.
 
-### Adjust the `formats` option in the JSON configuration (recommended).
 
-See the [`formats` JSON configuration](https://www.tiny.cloud/docs/tinymce/latest/content-formatting/).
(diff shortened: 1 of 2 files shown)


docs/classic-ui/tinymce-customization.md
```diff
@@ -0,0 +1,46 @@
...
+# TinyMCE customization
+
+This chapter is a developer reference manual for customizing {term}`TinyMCE`.
+
+
+(tinymce-remove-formats-label)=
+
+## Remove formats
+
+In Plone 6, TinyMCE has a special logic that automatically reads registered files named {file}`tinymce-formats.css` and adds the CSS classes defined in those files to TinyMCE's {menuselection}`Format --> Formats` menu.
```
petschki:
```suggestion
In Plone 6, TinyMCE has a special logic that automatically reads registered files named {file}`tinymce-formats.css` and adds the CSS classes defined in those files to TinyMCE's {menuselection}`Format --> Formats` menu by using the [`importcss_file_filter` option](https://www.tiny.cloud/docs/tinymce/latest/importcss/#importcss_file_filter).
```

petschki (changes requested):

stevepiercy:
Direct link to preview https://plone6--1807.org.readthedocs.build/classic-ui/tinymce-customization.html

petschki:
@stevepiercy I've changed the order of information and enhanced it:

1. customize formats (with recommended JSON config option and example how to update)
2. remove formats

stevepiercy:
@petschki I made a few more revisions to better explain why someone should choose one option over another, and turned the .css file option into more of a how-to-guide.

petschki (approved):
👍🏼

szakitibi:
Please note following the instructions of [Edit the Formats option in the TinyMCE control panel](https://plone6--1807.org.readthedocs.build/classic-ui/tinymce-customization.html#edit-the-formats-option-in-the-tinymce-control-panel) will not add the new item to the "Format" menu. Editing the "Formats" JSON will only add the format to the configurations, but it will be still invisible in the TinyMCE's menu. E.g. "discreet" for classic demo portal is registered as one of the "Formats" in the JSON, but it is nowhere to be found in the TinyMCE menu.

As a last step "Formats" must be registered under one of styles (in the case of "myformat", under "Block styles" of the controlpanel), only then will it pop up in the TinyMCE's menu.

docs/classic-ui/tinymce-customization.md
```diff
@@ -0,0 +1,87 @@
...
+They are less complicated to implement than the third option.
+
+
+### Add-on GenericSetup configuration file
+
+This option is best for system administrators and developers who write their own add-ons to ease reproducibility.
+
+You can add a GenericSetup configuration file to your add-on, such as {file}`profiles/default/registry/tinymce.xml`, with the following content.
+
+```xml
```
yurj, line 31: What about
```xml
  <records interface="Products.CMFPlone.interfaces.ITinyMCESchema" prefix="plone">
    <value key="other_settings">
      {
        "extended_valid_elements":"i[class],script[type|src]",
        "style_formats": [
          {
            "title": "Highlight",
            "block": "p",
            "attributes": {
              "class": "featured"
            }
          },
          {
            "title": "Alert",
            "block": "p",
            "attributes": {
              "class": "redalert"
            }
          }
        ],
        "style_formats_merge": true
      }
    </value>
  </records>
```

docs/classic-ui/tinymce-customization.md
```diff
@@ -0,0 +1,87 @@
...
+1.  Name the file {file}`tinymce-formats.css`.
+1.  Register the file as a resource in your Plone site.
+1.  Include the file in the TinyMCE control panel in the textarea input {guilabel}`Choose the CSS used in WYSIWYG Editor Area`.
+    1.  Navigate to {menuselection}`Site Setup --> TinyMCE --> Default`, or append `@@tinymce-controlpanel` to the root of your website in your browser's location bar.
+    1.  Scroll down to {guilabel}`Choose the CSS used in WYSIWYG Editor Area`.
+
+CSS styles defined in this file will automatically be added to the top level of TinyMCE's {menuselection}`Format --> Formats` menu.
+
+
+## Remove imported formats
```
yurj, line 77: is there a generic setup xml example to do this? This would pair with the above example of a GS xml.

petschki:
@szakitibi yes you're right. forgot about that. But depending on the format you can also add it to the "Inline formats"

docs/classic-ui/tinymce-customization.md
petschki: you mean removing it? sure... just add the updated record to the xml and add `purge="true"` as parameter.

```
<record field="content_css"
          interface="plone.base.interfaces.controlpanel.ITinyMCESchema"
          name="plone.content_css"
  >
    <value purge="true">
      <element>++theme++barceloneta/tinymce/tinymce-ui-content.css</element>
    </value>
  </record>
```

feel free to open a PR and enhance it ...

szakitibi:
> @szakitibi yes you're right. forgot about that. But depending on the format you can also add it to the "Inline formats"

Exactly. That is why I used "under one of the styles". The official documentation should be clear about this part, that the registered "formats" can be registered under different styles, and that will determine their position in the TinyMCE menu.

petschki:
Last but not least there should be a warning, that the provided genericsetup xml will kill the default formats. If those should be preserved, you need to copy them over to your `tinymce.xml`

docs/classic-ui/tinymce-customization.md
petschki: Beware: `other_settings` will override all previously made settings. This is meant to inject settings which are not populated in other config fields.

And the `style_formats_merge` options provides the default "tinymce styles" but that's not exactly the same as the plone default tinymce styles ...

szakitibi:
Also another comment regarding the use of "formats" JSON. I think it is not possible to add top level items (like the `highlight-inline`) to "Formats" dropdown menu with the available "* styles" configurations. It seems currently only the `tinymce-formats.css` is able to do that trick.

stevepiercy:
I would be happy to accept a pull request with revisions to this document. These comments are great, but they won't get into the documentation unless someone who has the technical knowledge writes them up in PR.

I do not have this technical knowledge. I am only a technical writer who translates Plone-ish to English and sprinkles some MyST syntax on top to make it pretty. Thanks for your collaboration.

docs/classic-ui/tinymce-customization.md
stevepiercy: Done in 43807fa17a507b407dcc353ec812461eeadccbf9

PR #1808, opened 2024-12-05
Continue tinymce customization docs

petschki:
Continued from #1807 and implemented the comments.

@szakitibi not shure how and where to write the info, that the formats can only be placed on top if you use the third approach ... I'm open for suggestions.

----
📚 Documentation preview 📚: https://plone6--1808.org.readthedocs.build/

### DIFF
--- a/docs/classic-ui/tinymce-customization.md
+++ b/docs/classic-ui/tinymce-customization.md
@@ -103,6 +103,27 @@ CSS styles defined in this file will automatically be added to the top level of 
 
 ## Remove imported formats
 
+Similar to adding formats, you can remove formats in a couple of ways.
+
+
+### Add-on GenericSetup configuration file
+
+Alternatively, you can use GenericSetup in your add-on.
+
+```xml
+<record field="content_css"
+          interface="plone.base.interfaces.controlpanel.ITinyMCESchema"
+          name="plone.content_css"
+  >
+    <value purge="true">
+      <element>++theme++barceloneta/tinymce/tinymce-ui-content.css</element>
+    </value>
+  </record>
+```
+
+
+### Configure the TinyMCE control panel
+
 Plone 6 Classic UI ships with the Barceloneta theme which includes two custom formats, `highlight-inline` and `p.highlight-paragraph`, in the TinyMCE {menuselection}`Format --> Formats` menu.
 You can remove these formats through the TinyMCE control panel.
 
--- a/docs/classic-ui/tinymce-customization.md
+++ b/docs/classic-ui/tinymce-customization.md
@@ -26,7 +26,7 @@ They are less complicated to implement than the third option.
 
 This option is best for system administrators and developers who write their own add-ons to ease reproducibility.
 
-You can add a GenericSetup configuration file to your add-on, such as {file}`profiles/default/registry/tinymce.xml`, with the following content.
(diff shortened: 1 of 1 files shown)


stevepiercy:
@petschki I made a few more revisions to bring it all together.

Direct link to preview at https://plone6--1808.org.readthedocs.build/classic-ui/tinymce-customization.html

szakitibi:
> @szakitibi not shure how and where to write the info, that the formats can only be placed on top if you use the third approach ... I'm open for suggestions.

Ok, I think we have only *two solutions* here. The "Formats+styles" and the `tinymce-formats.css` configuration are more like complementary to each other, rather then being alternative solutions.

The "Formats" JSON + registration in "Header/Inline/Block/Alignment/Table styles"  field will insert the new format in the respective "Formats" submenu like this:

image

On the other hand the `tinymce-formats.css` registered and configured will add the styles at the top level of the "Formats" menu:

image

---

The "Formats+styles" solution can be configured two ways:

1. TTW in the TinyMCE controlpanel - **easy for editors**.
2. Using generic setup - **meant for plone developers**.

The `tinymce-formats.css` solution is for **plone deveopers only**, since it requires access to the product source code, its steps are:

1. Adding the CSS rules in a `tinymce-formats.css` file.
2. Registering the file as a plone resource using zcml.
3. Adding it to the `ITinyMCESchema.content_css` field in registry using generic setup.

Note: Step 3 is possible via TTW, but **the other two steps happen in product code**! As far as I see only the generic setup should be documented for `tinymce-formats.css` solution, since the whole thing is meant for **plone developers**.

I see no point for mixing TTW into it in this subchapter discussing how to add new formats using `tinymce-formats.css`. However...

---

There is the topic of cleaning up the default barceloneta formats, where we have again two options:

1. The TTW option - **easy for editors**. Remove the `tinymce-formats.css` in the controlpanel and edit the "Formats" JSON + styles to remove any formats you do not want.
2. The generic setup solution - **for plone developers**. Update the `ITinyMCESchema` fields as you wish using generic setup in your product.

---

So the question is more like: What is the target group of this page? Editors? Plone developers? Both?

Depending on the answer the documentation should be split into parts. The GS stuff should be in the plone developers section and the TTW in the editors section of the page.

stevepiercy:
@szakitibi I see what you're saying, but... read on.

---

Cleaning up Barceloneta is up for discussion within the Classic UI team for inclusion in Plone 7 or later. That's beyond this PR's purpose, which is focused on Plone 6 Documentation.

---

"TinyMCE customization" is for all audiences. We try not to cast any roles in Plone 6 Documentation.

For example, let's say I'm performing the role of developer. I'm in a rush and I don't want to take the time to create an add-on. I can just copy-paste stuff from one site to another, as I only maintain two Plone sites, and complete the task in a few minutes.

People search for how to perform a task, not their role. It's a principle from the [Diátaxis framework](https://www.diataxis.fr/).

It's our job as documentarians to inform people toward making a decision that is appropriate to them. For example, we recently updated documentation for how to [Install Plone](https://6.docs.plone.org/install/index.html#install-plone) to do that. As such, we can offer the options and inform readers of the requirements to perform the task, regardless of their role.

---

After all that, is any of the information in this pull request not accurate? What specifically is missing or needs correction?

stevepiercy:
@petschki one final review from you, please. I think this is ready to merge.

petschki:
I cannot review, but approved 👍🏼

Issue #1611, opened 2024-01-21
Automate screenshots and videos


stevepiercy:
Documentation should have current screenshots of its user interface for its end user manuals, for both Classic UI and Volto. We also want to detect when there is a change in the UI, where the current image differs from the previous one, and have it reported as a warning or error to get the attention of the contributor in a pull request or earlier during testing.

This used to be a feature in Plone 5 documentation, but has since become unsupported. Now there are better tools and there are some efforts already underway.

- At PloneConf 2023, @polyester developed a proof of concept for using Cypress to collect screenshots of the Volto frontend Plone demo at https://demo.plone.org/. Volto already uses Cypress for its GitHub Workflows and automated testing. Further help on bridging the gap to make it use a Docker container would be much appreciated.
- At the Midsummer Sprint 2023, and subsequently at PloneConf 2023, a migration of Classic UI's robot framework tests to [Playwright](https://playwright.dev/) is well under way. @gforcada and @datakurre are the main contributors, along with @thet and @petschki.
- A member from the Python Documentation team suggested using a tool based on Playwright called [shot-scraper](https://github.com/simonw/shot-scraper?tab=readme-ov-file).
- @stevepiercy is working on a proposal for Google Season of Docs that would include this issue. We did not get accepted in 2024. We'll try again in 2025.
- https://github.com/plone/Products.CMFPlone/issues/3987

Manas-Kenge:
@stevepiercy what do you think about https://heroshot.sh/? I was doing some research and came across this https://heroshot.sh/docs/guide/automated-updates.html

stevepiercy:
Interesting, but I don't see an integration with Sphinx in its docs. Can you get it to work with Sphinx?

If it doesn't work with Sphinx, would you like to make a feature request?

https://github.com/omachala/heroshot/issues

Manas-Kenge:
I raised an issue, let's see if they add the integration

Manas-Kenge:
Hey @stevepiercy, just to give you an update — heroshot added support for Sphinx: https://github.com/omachala/heroshot/issues/34. 
Would you take a look at their framework and decide if it’s something Plone would use? Since the scope of the PR will be pretty big, I’d like to get a green light before I put in the effort.

stevepiercy:
[...]

PR #1812, opened 2024-12-09
Recording videos and images for documentation

stevepiercy:
Documentation includes:

- Initialize and install environment
- Run servers

See #1611 and https://github.com/plone/Products.CMFPlone/issues/3987

----
📚 Documentation preview 📚: https://plone6--1812.org.readthedocs.build/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 VALEFILES       := $(shell find -L $(DOCS_DIR) -type d \( -path $(DOCS_DIR)/plone.restapi/lib/* -o  -path $(DOCS_DIR)"/plone.restapi/performance/*" \) -prune -false -o -type f -name "*.md" -print)
+RECORDINGS		= ./recordings
 
 # Add the following 'help' target to your Makefile
 # And add help text after each target name starting with '\#\#'
@@ -245,6 +246,11 @@ rtd-pr-preview:  ## Build pull request preview on Read the Docs
 .PHONY: storybook
 storybook:
 	cd submodules/volto && pnpm i && pnpm build:registry && pnpm --filter @plone/volto build-storybook -o ../../../../_build/html/storybook
+
+.PHONY: recording-init  ## Initializes the image and video recording environment
+recording-init:
+	mkdir "$(RECORDINGS)" && cd "$(RECORDINGS)" && pipx run cookieplone --no-input project
+	cd "$(RECORDINGS)/project-title" && make install
 
 .PHONY: all
 all: clean vale linkcheck html  ## Clean docs build, then run vale and linkcheck, and build html
--- a/RECORDINGS.md
+++ b/RECORDINGS.md
@@ -33,7 +33,15 @@ module.exports = defineConfig({
 });
 ```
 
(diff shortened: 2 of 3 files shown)


stevepiercy:
I've pushed this as far as I can, the sample test failed, and I am stuck. I could use an assist to push this forward. See the docs in this PR for details.

I'm also not convinced that Cypress Recorder is the best way forward to create tests, after seeing how slick Playwright is.

Makefile
```diff
@@ -246,5 +247,10 @@ rtd-pr-preview:  ## Build pull request preview on Read the Docs
 storybook:
 	cd submodules/volto && pnpm i && pnpm build:registry && pnpm --filter @plone/volto build-storybook -o ../../../../_build/html/storybook
 
+.PHONY: recording-init  ## Initializes the image and video recording environment
+recording-init:
+	mkdir "$(RECORDINGS)" && cd "$(RECORDINGS)" && pipx run cookieplone --no-input project
+	cd "$(RECORDINGS)/project-title" && make install
```
davisagli, line 253: Oh, clever way to get an up-to-date project ready for recording!

RECORDINGS.md
```diff
@@ -0,0 +1,73 @@
...
+A new browser window will pop up.
+
+Select {guilabel}`E2E Testing`.
+
+Select {guilabel}`Chrome`, because it is the only browser that supports Cypress Studio.
+
+Click {guilabel}`Start E2E Testing in Chrome`.
+
+And that's pretty much as far as I got.
+Even though I created a page, `/document`, and ran the sample test, it failed.
```
davisagli, line 53: You don't have to create a page, the example test does that. But the example test was broken. 😞 Fixed in https://github.com/plone/cookieplone-templates/pull/138

davisagli:
> I'm also not convinced that Cypress Recorder is the best way forward to create tests, after seeing how slick Playwright is.

I don't have experience with either one (only writing Cypress tests in code). I think you can pick whichever one looks more promising. You can still use the same approach for installing a project and starting the backend and frontend. The main differences between the "acceptance" commands for starting Plone and the normal ones are:
- it runs on port 55001 to avoid conflicting if you've already got a non-test instance running
- the default credentials are admin/secret instead of admin/admin
- it runs under bin/robot-server from plone.app.robotframework, which means there are some endpoints exposed so that there's a way for the test runner to reset the instance after each test. I don't recall the details but can find a pointer for you if that sounds useful

davisagli:
> there are some endpoints exposed so that there's a way for the test runner to reset the instance after each test. I don't recall the details but can find a pointer for you if that sounds useful

That's what the Cypress infrastructure in Volto is doing here:
- https://github.com/plone/cookieplone-templates/blob/main/frontend_addon/%7B%7B%20cookiecutter.__folder_name%20%7D%7D/cypress/support/e2e.js
- https://github.com/plone/volto/blob/main/packages/volto/cypress/support/reset-fixture.js

RECORDINGS.md
```diff
@@ -0,0 +1,53 @@
...
+
+This project will serve as baseline for the tests. Eventually, we could pull a specific branch for generating the tests.
+
+Now you will start the backend, frontend, and acceptance test servers, one each in its own terminal session.
+
+```{note}
+None of these commands are documented anywhere.
+You could run `make help`, and get a dump of the commands, but that lacks context for usage.
+See https://github.com/plone/documentation/issues/1758.
+```
```
sneridagh, line 24: @stevepiercy this is not true, the commands do what you say below, and in the Makefile help.

RECORDINGS.md
sneridagh: This is fixed now, since I created a dedicated Cypress folder in `recordings/cypress`, the tests are in there. They will use the generated project as baseline to run. It uses the same approach than the Cookieplone setup.

@stevepiercy let's review it together tomorrow.

PR #1823, opened 2024-12-20
Plone 6.0.14 and 6.1.0b2 are out with Python 3.13 support.

mauritsvanrees:
Marked as draft PR, because the `plone-backend` Docker images are not ready yet.

----
📚 Documentation preview 📚: https://plone6--1823.org.readthedocs.build/

### DIFF
--- a/docs/conceptual-guides/package-management.md
+++ b/docs/conceptual-guides/package-management.md
@@ -32,11 +32,11 @@ If you want to check out a Plone core package for development, or want to overri
 
 ```text
 # constraints.txt with unresolvable version conflict
--c https://dist.plone.org/release/{PLONE_BACKEND_PATCH_VERSION}/constraints.txt
-plone.api>=2.0.0a3
+-c https://dist.plone.org/release/6.0.9/constraints.txt
+plone.api>=2.1.0
 ```
 
-Unfortunately `pip` does not allow overriding constraints this way. 
+Unfortunately `pip` does not allow overriding constraints this way.
 {term}`mxdev` solves this issue.
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -441,7 +441,7 @@ def source_replace(app, docname, source):
 # Dict of replacements.
 source_replacements = {
     "{PLONE_BACKEND_MINOR_VERSION}": "6.0",
-    "{PLONE_BACKEND_PATCH_VERSION}": "4",
+    "{PLONE_BACKEND_PATCH_VERSION}": "6.0.14",
 }
 
 # Finally, configure app attributes.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -441,7 +441,6 @@ def source_replace(app, docname, source):
 # Dict of replacements.
 source_replacements = {
     "{PLONE_BACKEND_MINOR_VERSION}": "6.0",
-    "{PLONE_BACKEND_PATCH_VERSION}": "6.0.14",
 }
 
 # Finally, configure app attributes.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -318,8 +318,8 @@ myst_substitutions = {
     "postman_response": "![](../_static/img/postman_response.png)",
     "postman_retain_headers": "![](../_static/img/postman_retain_headers.png)",
     "fawrench": '<span class="fa fa-wrench" style="font-size: 1.6em;"></span>',
-    "SUPPORTED_PYTHON_VERSIONS_PLONE60": "3.8, 3.9, 3.10, 3.11, or 3.12",
-    "SUPPORTED_PYTHON_VERSIONS_PLONE61": "3.10, 3.11, or 3.12",
+    "SUPPORTED_PYTHON_VERSIONS_PLONE60": "3.9, 3.10, 3.11, 3.12, or 3.13",
+    "SUPPORTED_PYTHON_VERSIONS_PLONE61": "3.10, 3.11, 3.12, or 3.13",
 }
 
 
@@ -441,7 +441,7 @@ def source_replace(app, docname, source):
 # Dict of replacements.
 source_replacements = {
     "{PLONE_BACKEND_MINOR_VERSION}": "6.0",
-    "{PLONE_BACKEND_PATCH_VERSION}": "6.0.13",
+    "{PLONE_BACKEND_PATCH_VERSION}": "4",
 }
 
 # Finally, configure app attributes.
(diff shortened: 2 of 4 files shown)


docs/conf.py
```diff
@@ -441,7 +441,7 @@ def source_replace(app, docname, source):
 # Dict of replacements.
 source_replacements = {
     "{PLONE_BACKEND_MINOR_VERSION}": "6.0",
-    "{PLONE_BACKEND_PATCH_VERSION}": "6.0.13",
+    "{PLONE_BACKEND_PATCH_VERSION}": "4",
```
stevepiercy: This line gets rendered in the code block at https://plone6--1823.org.readthedocs.build/conceptual-guides/package-management.html#manage-backend-python-packages, which becomes this URL https://dist.plone.org/release/4/constraints.txt. Can you verify?

```suggestion
    "{PLONE_BACKEND_PATCH_VERSION}": "6.0.14",
```

docs/conf.py
```diff
@@ -318,8 +318,8 @@
     "postman_response": "![](../_static/img/postman_response.png)",
     "postman_retain_headers": "![](../_static/img/postman_retain_headers.png)",
     "fawrench": '<span class="fa fa-wrench" style="font-size: 1.6em;"></span>',
-    "SUPPORTED_PYTHON_VERSIONS_PLONE60": "3.8, 3.9, 3.10, 3.11, or 3.12",
-    "SUPPORTED_PYTHON_VERSIONS_PLONE61": "3.10, 3.11, or 3.12",
+    "SUPPORTED_PYTHON_VERSIONS_PLONE60": "3.9, 3.10, 3.11, 3.12, or 3.13",
```
stevepiercy, line 321: When we drop a Python version, I think of all the places where we should add an admonition. For the drop of Python 3.8 in 6.0.14, where would be good places?

I'll push a couple of commits for discussion. If they're not appropriate, we can change or drop them.

stevepiercy (changes requested):

docs/conf.py
stevepiercy: Done in b9056cf2cb3d3b906d01f8155ab095e378f9555b

docs/conf.py
stevepiercy: It looks like the include was the place for this, and it gets pulled into three places.

stevepiercy:
Preview changes:

- https://plone6--1823.org.readthedocs.build/backend/upgrading/intro.html#in-place-migrations
- https://plone6--1823.org.readthedocs.build/conceptual-guides/package-management.html#manage-backend-python-packages
- https://plone6--1823.org.readthedocs.build/install/create-project-cookieplone.html#python
- https://plone6--1823.org.readthedocs.build/admin-guide/install-buildout.html#python
- https://plone6--1823.org.readthedocs.build/admin-guide/install-pip.html#python
- https://plone6--1823.org.readthedocs.build/install/create-project.html#python

docs/conf.py
mauritsvanrees: Oops, it was late at night when I did this, and I did not catch that I replaced the whole version number instead of only the last digit... Your change is good.

docs/conf.py
mauritsvanrees: LGTM, thanks!

mauritsvanrees:
I saw one place where we were already using `PLONE_BACKEND_PATCH_VERSION` and we actually should not:

https://plone6--1823.org.readthedocs.build/conceptual-guides/package-management.html#manage-backend-python-packages

It has this example of an unresolvable version conflict:

```
# constraints.txt with unresolvable version conflict
-c https://dist.plone.org/release/6.0.14/constraints.txt
plone.api>=2.0.0a3
```

But that is actually not a conflict!  The plone.api version in the 6.0.14 constraints (and actually already in 6.0.0) is higher than 2.0.0a3, so there is no conflict.

I have pushed a commit to use different versions, which are fine to be hardcoded:

```
-c https://dist.plone.org/release/6.0.9/constraints.txt
plone.api>=2.1.0
```

Ah, the preview has just updated.

mauritsvanrees:
I will mark this as ready for review/merge now, as the Docker images are ready.

docs/conceptual-guides/package-management.md
```diff
@@ -32,11 +32,11 @@ If you want to check out a Plone core package for development, or want to overri
 
 ```text
 # constraints.txt with unresolvable version conflict
--c https://dist.plone.org/release/{PLONE_BACKEND_PATCH_VERSION}/constraints.txt
-plone.api>=2.0.0a3
+-c https://dist.plone.org/release/6.0.9/constraints.txt
+plone.api>=2.1.0
```
stevepiercy, line 36: Dang, now that right there is some _serious_ institutional knowledge!

docs/conf.py
```diff
@@ -441,7 +441,7 @@ def source_replace(app, docname, source):
 # Dict of replacements.
 source_replacements = {
     "{PLONE_BACKEND_MINOR_VERSION}": "6.0",
-    "{PLONE_BACKEND_PATCH_VERSION}": "6.0.13",
+    "{PLONE_BACKEND_PATCH_VERSION}": "6.0.14",
```
stevepiercy: Actually, the only place this was used was in that serious institutional knowledge bit, so let's remove this.

```suggestion
```

stevepiercy (changes requested):

stevepiercy (approved):
Will merge when green.

docs/conceptual-guides/package-management.md
mauritsvanrees: My secret is to grep through local copies that I have of all `versions.cfg` files for all Plone versions. :-)

docs/conf.py
mauritsvanrees: No: you also added this in `docs/backend/upgrading/intro.md`:

```
Then upgrade from Plone 5 to the latest Plone 6 version ({PLONE_BACKEND_PATCH_VERSION}).
```

So please bring that definition back.

Alternatively, we could remove the details from those lines, which I think is fine, and no longer mention Plone 2.5 and 3:

```
If you are at Plone 4 and want to upgrade to the latest Plone 6, you should approach this in several steps:

-   First upgrade from Plone 4 to the latest Plone 5 version.
-   Then upgrade from Plone 5 to the latest Plone 6 version.
```

mauritsvanrees: You merged the PR while I was adding this comment.
I will make a new PR with my suggestion.

stevepiercy: Sorry about that. Thank you for your diligence.

PR #1829, opened 2025-01-03
Add terms to Glossary for block extension examples

stevepiercy:
This needs to be merged before https://github.com/plone/volto/pull/6560 so that that PR can pull in the Glossary terms.

----
📚 Documentation preview 📚: https://plone6--1829.org.readthedocs.build/

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -871,4 +871,18 @@ variation
     Extending this example, a developer can add a boolean field to the block that toggles the display of the link.
     Thus an editor can select between `title + link` or just `title`.
 
+HOC
+Higher-Order Component
+    A higher-order component (HOC) is an advanced technique in React for reusing component logic.
+    HOCs are not part of the React API, per se.
+    They are a pattern that emerges from React's compositional nature.
+    Concretely, a higher-order component is a function that takes a component and returns a new component.
+    
+    ```{important}
+    Higher-order components are not commonly used in modern React code.
+    ```
+    ```{seealso}
+    https://legacy.reactjs.org/docs/higher-order-components.html
+    ```
+
 ```
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -863,9 +863,12 @@ schema enhancer
     This function receives an object with `formData`, which is the block data; `schema`, which is the original schema to modify; and the injected `intl`, which aids with internationalization.
 
 variation
-    A variation is a common development pattern that provides versions of a given block.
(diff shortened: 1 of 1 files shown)


ichim-david (dismissed):
@stevepiercy I am ok with these additions.
I see that we have this info here:
https://github.com/plone/volto/blob/374dfbcfe6a798fa66f8257e0f560fa1963d2646/docs/source/blocks/extensions.md?plain=1#L12

and 
https://github.com/plone/volto/blob/374dfbcfe6a798fa66f8257e0f560fa1963d2646/docs/source/blocks/extensions.md?plain=1#L102

Ppl might not understand what InlineForm is and what it does but it can be a starting point for them to search for what is that component

docs/glossary.md
```diff
@@ -857,4 +857,15 @@ Vale
...
+    A schema enhancer uses the `schemaEnhancer` function, which can modify the schema used by the `InlineForm` component.
+    Any registered extension plugin can provide a `schemaEnhancer` function.
+    This function receives an object with `formData`, which is the block data; `schema`, which is the original schema to modify; and the injected `intl`, which aids with internationalization.
+
+variation
+    A variation is a common development pattern that provides versions of a given block.
+    You can use a variation to enhance the default behavior of a block without shadowing its stock components.
+    These enhancements can either be at the settings level via schema enhancers or, if the code of your block allows it, use alternative renderers showing the enhanced fields, modifying the block behavior or its look and feel.
+    Editors can select a variation on demand.
+
```
erral, line 887: I would make this last definition simpler.

The standard pattern for variations is to offer "alternative views" for the same data. Example: A variation shows  news items with "title + link". Another variation shows "title + description + link", another variations shows "title + image + link".

As an advanced pattern we can say that a variation can enhance the block adding additional fields to the block. Example: 1 variation shows news items with "title + link". Another variation adds an additional boolean field to the block to decide whether to show the link. So in the end the editor can select between "title + link" and "title + link (if the boolean is set to true)".

erral (changes requested):

docs/glossary.md
stevepiercy: > "title + link" and "title + link (if the boolean is set to true)".

Aren't these the same?

stevepiercy:
@ichim-david yup, that's the source of the text I cobbled together.

@erral I simplified it in 05a3bd5e48406301df7f7715fdd34273b65c0c8e.

erral (approved):

PR #1832, opened 2025-01-08
Fix link to Portal Modelo

stevepiercy:
https://www.interlegis.leg.br/produtos-servicos/portal-modelo/ redirects to https://www12.senado.leg.br/interlegis

I am not sure whether it is still called Portal Modelo. Help!

----
📚 Documentation preview 📚: https://plone6--1832.org.readthedocs.build/

PR #1833, opened 2025-01-09
Distribution: add example for hiding fields from the form.

mauritsvanrees:
I am making a distribution for a client.

* I don't want to show the description field because it should remain empty (at least that is how it is on their current Plone 4 site).
* I don't want to show the default_language field, because I am setting this (and four available languages) in the Generic Setup profile.
* Same for the portal_timezone: it is a site for Belgium, so I am setting Europe/Brussels in the Generic Setup profile, so this timezone is always set, even when a colleague in Portugal creates a site on his laptop.

Unfortunately currently these fields have to be in the schema, as documented, otherwise you get an error. So instead  I found out how to hide these fields.

----
📚 Documentation preview 📚: https://plone6--1833.org.readthedocs.build/

### DIFF
--- a/docs/developer-guide/create-a-distribution.md
+++ b/docs/developer-guide/create-a-distribution.md
@@ -245,6 +245,30 @@ You may have noticed the entries for both `default_language` and `portal_timezon
 `plone.distribution` adds both definitions at runtime, providing a list of languages and timezones available on the installation.
 ````
 
+If you want to hide fields in the form, you can use the `uischema`.
+This is especially useful for fields that must be in the schema, but for which you always want the default value.
+The next lines hide three fields:
+
+```json
+{
+  "schema": {
+    ...
+  },
+  "uischema": {
+      "description": {
+          "ui:widget": "hidden"
+      },
+      "default_language": {
+          "ui:widget": "hidden"
+      },
+      "portal_timezone": {
+          "ui:widget": "hidden"
+      }
+  }
+}
+```
+
+
 ## Add a dependency on an add-on
 
 If you want to add a Plone backend add-on to your Plone distribution, then you must perform the following steps.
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
Direct link to the updated section in the preview:

https://plone6--1833.org.readthedocs.build/developer-guide/create-a-distribution.html#schema-json

docs/developer-guide/create-a-distribution.md
```diff
@@ -245,6 +245,30 @@ You may have noticed the entries for both `default_language` and `portal_timezon
...
 ````
 
+If you want to hide fields in the form, you can use the `uischema`.
+This is especially useful for fields that must be in the schema, but for which you always want the default value.
+The next lines hide three fields:
+
+```json
+{
+  "schema": {
+    ...
```
stevepiercy: I think this makes the JSON snippet valid.

```suggestion
    "…": "…"
```

docs/developer-guide/create-a-distribution.md
```diff
@@ -245,6 +245,30 @@ You may have noticed the entries for both `default_language` and `portal_timezon
...
+  "uischema": {
+      "description": {
+          "ui:widget": "hidden"
+      },
+      "default_language": {
+          "ui:widget": "hidden"
+      },
+      "portal_timezone": {
+          "ui:widget": "hidden"
+      }
```
stevepiercy: The above 9 lines are indented 4 spaces, instead of 2 as in the rest of the sample. Can you make indents consistent? I don't know whether we have a preference for 2 or 4 spaces in JSON, so whatever your editor does is fine with me.

stevepiercy (changes requested):

docs/developer-guide/create-a-distribution.md
mauritsvanrees: Ah, right.  The main example has an indent of two, so let's use that.  Done.

stevepiercy (approved):
Thank you!

stevepiercy:
The failing test is addressed in a separate PR https://github.com/plone/documentation/pull/1832. Merging, and thank you!

PR #1845, opened 2025-02-07
Prepare for Plone 6.1 release

stevepiercy:
TODO:

- [x] After the release, we need to run `pipx run cookieplone project --no-input` and replace the current console output with the command's in `docs/install/create-project-cookieplone.md`.

Refs https://github.com/plone/Products.CMFPlone/issues/4101

----
📚 Documentation preview 📚: https://plone6--1845.org.readthedocs.build/

PR #1849, opened 2025-02-08
Use `6-latest` for the truly latest 6

stevepiercy:
- Thus we don't have to update it for 6.2 🥳

----
📚 Documentation preview 📚: https://plone6--1849.org.readthedocs.build/

### DIFF
--- a/docs/admin-guide/install-pip.md
+++ b/docs/admin-guide/install-pip.md
@@ -58,7 +58,7 @@ python3 -m venv venv
 Install Plone and a helper package, {term}`pipx`.
 
 ```shell
-venv/bin/pip install -c https://dist.plone.org/release/6.1-latest/constraints.txt Plone pipx
+venv/bin/pip install -c https://dist.plone.org/release/6-latest/constraints.txt Plone pipx
 ```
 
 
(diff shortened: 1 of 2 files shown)


davisagli (changes requested):
No, please don't do that. Getting patch releases automatically is probably a good thing. Getting minor updates with new features without opting into it is too surprising.

stevepiercy:
@davisagli the context is for install only, so I don't understand your comment. If I install Plone, I expect the latest possible versions. Can you elaborate?

davisagli (commented):
@stevepiercy I was overlooking the context of where you changed it. It should be okay to use 6-latest in these places. I wouldn't want to use it in the buildout instructions, since the buildout could be re-run in the future and we wouldn't want an accidental upgrade at that point.

davisagli (approved):

stevepiercy:
@davisagli OK, yeah, that's what I had in mind. There are some places I didn't change, for example in the version upgrade guides or this:

https://github.com/plone/documentation/blob/d633bbb527fad10338f16aa61c99610022352278/docs/install/create-project.md?plain=1#L196

Or updating documentation's own requirements to avoid those unwanted surprises.

https://github.com/plone/documentation/pull/1845/files#diff-54c8b05644cf835f41801972f716dfd4eb526ae9be79c33297be600dfeb7c259

I tried to keep the context in mind when doing these recent updates, but it's possible I goofed.

PR #1852, opened 2025-02-11
Clarified steps in create-project-cookieplone.md for better understan…

codehashira99:
### Why:
The setup instructions for creating a new cookieplone project were unclear about the version selection for Plone and Volto, especially for newcomers. This caused confusion during the setup process.

### What:
- Clarified the instructions regarding the version selection in the `create-project-cookieplone.md` file.
- Provided default values for Plone and Volto versions when the user presses Enter, making it easier for them to proceed without manual input.
- Many users may either make a typo mistake while typing the versions or search for which version they have to type , so i just mentioned to press an enter for both (Plone and Volto versions ) so that their project is build on the stable current versions.

### How:
- Updated the documentation to explicitly mention that pressing Enter will select the latest stable version for both Plone and Volto.
- Added a sample command to demonstrate the version selection process.

### Testing:
- No functional changes, only documentation updates. Read through the file to ensure the instructions are clearer.
  
### Notes:
- This is part of the ongoing documentation improvements for Plone 6 setup.

----
📚 Documentation preview 📚: https://plone6--1852.org.readthedocs.build/

### DIFF
--- a/docs/install/create-project-cookieplone.md
+++ b/docs/install/create-project-cookieplone.md
@@ -124,7 +124,7 @@ pipx run cookieplone project
 
 You will be presented with a series of prompts.
 You can accept the default values in square brackets (`[default-option]`) by hitting the {kbd}`Enter` key, or enter your preferred values.
-For ease of documentation, we will use the default values.
+For ease of documentation, we will use the default values. 
 
 ```{tip}
 See the cookiecutter's README for how to [Use options to avoid prompts](https://github.com/plone/cookieplone/?tab=readme-ov-file#use-options-to-avoid-prompts).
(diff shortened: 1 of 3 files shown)


mkdocs.yml
stevepiercy, line 1: Please remove this file.

docs/install/create-project-cookieplone.md
```diff
@@ -189,8 +189,8 @@ before. Is it okay to delete and re-download it? [y/n] (y):
   [5/17] Author (Plone Foundation): 
   [6/17] Author E-mail (collective@plone.org): 
   [7/17] Should we use prerelease versions? (No): 
-  [8/17] Plone Version (6.1.0): 
-  [9/17] Volto Version (18.8.1): 
+  [8/17] Plone Version (6.1.0): (press enter if it is your first project)
```
stevepiercy: This is literal console output from running the command. We can't add something here that doesn't actually happen. Cookieplone would need to be updated with this altered prompt first.

Alternatively you can move instructions into the narrative documentation with the other admonitions above.

stevepiercy (changes requested):

codehashira99:
Alright sir, I will make the changes

PR #1853, opened 2025-02-12
Fix the livehtml file watcher

stevepiercy:
@sneridagh reported, and I verified during the Alpine Sprint, that when using `make livehtml`, any changes in the following symlinked paths...

```
volto
plone.api
plone.restapi
```

...would result in no changes detected, and therefore no live reload. This PR fixes that bug.

Note, this was working fine before, and even as recently as last week. There was a macOS update last week. It's unlikely to be related. Also if anyone recently set this up, it might not affect them.

----
📚 Documentation preview 📚: https://plone6--1853.org.readthedocs.build/


sneridagh (approved):
Works!

PR #1703, opened 2024-09-12
add notes to create an emergency user

1letter:
i created a subdirectory for users and groups and add a file with a short description to create an emergency user in a pip and buildout based installation of Plone.

Closes #948
Closes #1702

- https://plone6--1703.org.readthedocs.build/admin-guide/index.html
- https://plone6--1703.org.readthedocs.build/admin-guide/add-emergency-user.html

----
📚 Documentation preview 📚: https://plone6--1703.org.readthedocs.build/

PR #1854, opened 2025-02-12
Zope Manager Users

ale-rt:
This should fix some comments on the PR #1703

See https://plone6--1854.org.readthedocs.build/admin-guide/add-emergency-user.html

----
📚 Documentation preview 📚: https://plone6--1854.org.readthedocs.build/


ale-rt:
I decided to use "Zope Manager Users" because you might want to add the user even if there is no emergency.

1letter (approved):
Thanks for contribution! LGTM.

PR #1857, opened 2025-02-12
API packages in `docs/api`

sneridagh:
----
📚 Documentation preview 📚: https://plone7--1857.org.readthedocs.build/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,8 +7,9 @@
 /styles/Microsoft
 
 # symlinked from submodule
-docs/plone.api
-docs/plone.restapi
+docs/api/plone.api
+docs/api/plone.restapi
+docs/api/registry
 docs/volto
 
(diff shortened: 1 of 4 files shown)


stevepiercy:
Per our Discord call yesterday, we're going to put this PR on pause until I release Plone Sphinx Theme 2.0.0 as described in https://github.com/plone/Products.CMFPlone/issues/4097. We noticed that only the primary toctree's caption displays in the sidebar navigation, and gets clumped together.

We can wait for an open PR to get merged in PyData Sphinx Theme, https://github.com/pydata/pydata-sphinx-theme/pull/1241, or we can incorporate the PR into Plone Sphinx Theme and ignore the conflicting setting that we would not use anyway.

Alternatively we could use subfolders, but then the navigation gets unnecessarily nested.

PR #1858, opened 2025-02-12
Added windows support in install Plone with pip docs

pranshustuff:
## Description
I made this contributing as I was trying to install Plone with pip on my Windows laptop to test its feature.  These are fixes to errors I faced along the way.

The command ```venv/bin/``` is for Linux/MacOS

For Windows it is ```.\venv\Scripts```

Plus I was facing issues with pipx run cookie cutter.
If you're already in a venv, you can install cookiecutter with pip, so I added instructions for that.

## Page which was modified 
https://6.docs.plone.org/admin-guide/install-pip.html

----
📚 Documentation preview 📚: https://plone6--1858.org.readthedocs.build/admin-guide/install-pip.html

### DIFF
--- a/docs/admin-guide/install-pip.md
+++ b/docs/admin-guide/install-pip.md
@@ -31,7 +31,7 @@ For other installation options, see {ref}`get-started-install-label`.
 
 ## Prerequisites for installation
 
-```{include} /_inc_/_install-operating-system1.md
+```{include} /_inc/_install-operating-system1.md
 ```
 
 -   For Plone 6.1, Python {{SUPPORTED_PYTHON_VERSIONS_PLONE61}}
--- a/docs/admin-guide/install-pip.md
+++ b/docs/admin-guide/install-pip.md
@@ -31,8 +31,14 @@ For other installation options, see {ref}`get-started-install-label`.
 
 ## Prerequisites for installation
 
-```{include} /_inc/_install-operating-system1.md
-```
+-   An operating system that runs all the requirements mentioned. 
+Most UNIX-based operating systems are supported, including many Linux distributions, macOS, or Windows Subsystem for Linux (WSL) on Windows. 
+A UNIX-based operating system is recommended.
+
+```{important} 
+Windows alone is not recommended because Plone has not been extensively tested on it.
+If you get Plone to run on Windows alone, please feel free to document and share your process.
+``` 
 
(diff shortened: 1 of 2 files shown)


stevepiercy:
@pranshustuff did you use Windows Subsystem for Linux (WSL) on Windows or Windows alone?

I'm hesitant to recommend installation steps for Windows alone if they don't actually result in a functional Plone site, and I have no way to verify that. See the Important admonition at https://6.docs.plone.org/install/create-project-cookieplone.html#prerequisites-for-installation.

If you used WSL, then we might want to include that admonition into the pip installation docs.

pranshustuff:
I used Windows only.
I understand your concerns, since I have only created a basic site, I can't guarantee the working of more advanced features.

I can add the admonition instead. 

Although, I would like to leave in the activation of venv and installation of cookiecutter through pip directly instead of pipx as an alternative, since my issue with pipx was with something to do with file path, which I assume can happen with WSL too. I can write a WSL version for those.

docs/admin-guide/install-pip.md
```diff
@@ -55,12 +55,31 @@ Create a Python virtual environment.
...
+```shell
+source venv/bin/activate
+```
+
+For Windows:
+
+```shell
+.\venv\Scripts\activate
+```
+
```
stevepiercy: No. We do never activate virtual environments, as it causes more problems than it solves, especially for first-timers. Instead we call the binary by its explicit path, such as `venv/bin/pip install <package>`, which is much clearer. Please remove this entire section.

stevepiercy (changes requested):
I'm on the fence about adding commands for Windows for this installation method until and unless we have feedback confirming that a fully functional Plone site is running.

If we do go about it, then I would prefer to use [Sphinx Design's synchronized tabs](https://sphinx-design.readthedocs.io/en/latest/tabs.html#synchronised-tabs) to better display the options, as in this [example](https://6.docs.plone.org/contributing/documentation/setup-build.html#graphviz), with Linux as the default selected tab.

docs/admin-guide/install-pip.md
pranshustuff: I understand. I will just add the admonition and delete the activation section along with the windows only versions.

I apologise for not discussing it in the forum before opening the PR. I'll follow the process moving forward.

Thank you for your guidance!

pranshustuff:
I am closing the pull request and moving this discussion to the forums.

Thank you for your time and guidance! I learnt a lot today.

PR #1860, opened 2025-02-14
Clarify and rework the docs for adding a Zope manager user

ale-rt:
----
📚 Documentation preview 📚: https://plone6--1860.org.readthedocs.build/


acsr:
@ale-rt @stevepiercy I can confirm that these latest changes are a good improvement to layout the scope and purpose of this user and could be released in this state as a next step.

I suggest to either reopen the issue after publishing this state and add the section for using the zopeadduser command inside a Docker based deployment or crete a followup ticket.

Can you please take care of this since I have to enter another OpneSource related meeting outside of Plone during the Friday standup of the sprint, and can only jump back later next week. I already wrote my standup notes there.

PR #1864, opened 2025-02-16
Enhance Plone documentation style guide

stevepiercy:
- Clarify why we use one sentence per line
- Add links to Microsoft Style Guide for verb tense, usage, and examples.
- Add `.vscode/settings.json` reference and example file.

----
📚 Documentation preview 📚: https://plone6--1864.org.readthedocs.build/

### DIFF
--- a/docs/_static/settings.json
+++ b/docs/_static/settings.json
@@ -1,0 +1,5 @@
+{
+  "[markdown]": {
+    "editor.formatOnSave": false
+  }
+}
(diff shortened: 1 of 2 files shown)


gforcada (approved):
Although I'm a huge fan of semantic line breaks, and that you can submit suggestions for multiple lines (see [docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request)), the other changes look good 👍🏾

davisagli (approved):

stevepiercy:
In addition to the reasons against semantic line breaks already stated in this PR, they are a barrier to contributors who are not fluent in English. How would you teach it to volunteer contributors without scaring them away?

So between that, PEP8, and avoiding horizontal scrolling during diff reviews, the one sentence per line pattern is a reasonable compromise for our international audience.

Thanks for the link for the multiple line suggestion. TIL. I'll amend the docs. It would be good to point people to that. Almost all of the suggestions I see are single-line.

Issue #1645, opened 2024-03-20
Restructure Classic UI / Theming Chapter


MrTango:
### Links of affected pages in Plone Documentation, if any.

- https://6.docs.plone.org/classic-ui/theming/index.html

### Description

The sub Chapters like we have them in the trainings, is not help full here. It will lead to not knowing where to put information and having them in more than one place.

I suggest we keep it as flat as possible in this reference documentation.
For now is see chapters like this:

On index page, explain in short the 3 ways of theming.
- Change Logo & Favicon
- Custom CSS TTW 
- CSS Variables
  (Custom Properties)
- Creating Theme Packages
  'only the creating part (plonecli), not to much details, reference training for details'
  ## Custom Barceloneta Theme
  ## Custom Diazo Theme
  ## Custom Theme from scatch
- SCSS Structure
- Color modes

stevepiercy:
@MrTango does this issue supersede https://github.com/plone/documentation/issues/1286, and should the other be closed?

I'll try to get you in Discord to get more details.

kcm5474:
This would be a good idea. Also providing download links to some example basic custom theme files within the chapter would be helpful to provide reference to newcomers as to what an example of a finished theme should look like. The current Theme "downloads" on the Plone website are just images with no theme files available: https://plone.org/download/theme/ , and the themes I did find online for free were outdated or otherwise incorrect, and did not seem to properly target the elements currently in Plone 6 classic UI.

stevepiercy:
I closed #1286 as this issue supersedes it.

It doesn't make sense to move around empty and incomplete topics. Let's discuss in the @plone/classicui-team meeting this week, as this has been a long-standing issue. We need to determine the long-term goals of how to structure Theming Classic UI.

Here's the current state, and short-term recommendations.

[Theming of Classic UI](https://6.docs.plone.org/classic-ui/theming/index.html) - Currently only an index and table of contents.
-   [Classic UI theming based on Barceloneta](https://6.docs.plone.org/classic-ui/theming/barceloneta.html) - Complete
-   [Classic UI theming with Diazo](https://6.docs.plone.org/classic-ui/theming/diazo.html) - Empty. Recommend (1) update link on that page to this issue, (2) create a new issue for that page alone, and (3) update this issue's description to point to (2).
-   [Classic UI theming from scratch](https://6.docs.plone.org/classic-ui/theming/from-scratch.html) - Only an outline, incomplete. Same recommendations as Classic UI theming with Diazo.
[...]

PR #1870, opened 2025-02-20
Provide the content for the Diazo theming part

erral:
Can be seen as part of #1645

----
📚 Documentation preview 📚: https://plone6--1870.org.readthedocs.build/


docs/classic-ui/theming/diazo.md
```diff
@@ -0,0 +1,136 @@
...
+Diazo allows you to apply a theme contained in a static HTML web page to a dynamic website created using any server-side technology.
+With Diazo, you can take an HTML wireframe created by a web designer, and turn it into a theme for your favourite CMS, redesign the user interface of a legacy web application without even having access to the original source code, or build a unified user experience across multiple disparate systems, all in a matter of hours, not weeks.
+
+When using Diazo, you will work with syntax and concepts familiar from working with HTML and CSS.
+And by allowing you to seamlessly integrate XSLT into your rule files, Diazo makes common cases simple and complex requirements possible.
+
+
+## Create an add-on package
+
+To create a Diazo theme, you need to create an add-on package with [`plonecli`](https://github.com/plone/plonecli).
```
stevepiercy, line 25: What are the pre-requisites to running `plonecli`?

docs/classic-ui/theming/diazo.md
```diff
@@ -0,0 +1,136 @@
...
+        css:theme-children="#content"
+        css:content-children="article#content"
+        css:if-content="#content"/>
+```
+
+
+## How to theme using Diazo
+
+You can start with the provided {file}`rules.xml` file.
+You will need to write your own rules to bring the dynamic content from Plone into the theme.
```
stevepiercy, line 171: How will I write any rules, when there is no reference documentation linked here?

docs/classic-ui/theming/diazo.md
```diff
@@ -0,0 +1,136 @@
...
+
+Sometimes you will face difficult situations where you may find it hard to put items in the same place that Plone produces in very different places.
+
+For instance, you may need to put together the main menu, the language change, and the search box.
+Sometimes it is easier to override the corresponding template in Plone, build the new HTML structure there, and replace one thing in the {file}`rules.xml` file than trying to write complex Diazo rules or writing XSLT.
+
+The size of the {file}`rules.xml` file and the number of rules it contains can negatively impact the performance of your site.
```
stevepiercy, line 178: That won't be a problem, since there is no reference to how to write these rules. 🙃

docs/classic-ui/theming/diazo.md
```diff
@@ -0,0 +1,136 @@
...
+├── package.json
+├── README.rst
+├── rules.xml
+├── styles
+│   ├── theme.min.css
+│   └── theme.scss
+└── tinymce-templates
+    ├── bs-dark-hero.html
+    ├── bs-hero-left.html
+    └── bs-pricing.html
```
stevepiercy, line 62: It would be good to describe the purpose of each of these files here.

stevepiercy (changes requested):
I restructured the Diazo doc to present the critical path first, followed by advice. It flows a little better, but there are still some issues with it, namely a first-timer would have no clue what to do after reading this.

docs/classic-ui/theming/diazo.md
erral: The reference of writing those rules is at https://docs.diazo.org/en/latest/basic.html

erral: I have added a link to Diazo docs.

docs/classic-ui/theming/diazo.md
erral: See comment above :wink:

docs/classic-ui/theming/diazo.md
erral: I have reworded this stating that you don't need to install plonecli, and you can run it using pipx, like we run cookieplone: (pipx run plonecli)

stevepiercy: Sweet! Old way meets new!

docs/classic-ui/theming/diazo.md
```diff
@@ -0,0 +1,178 @@
...
+
+{file}`rules.xml`
+:   The file where you will write your rules to bring Plone content into the theme in the {file}`index.html` file.
+
+{file}`styles` folder
+:   Where you can save your theme's CSS files.
+
+{file}`tinymce-templates`
+:   Template files that can be loaded into the TinyMCE editor in Plone.
+    It requires additional configuration in the add-on's profile's {file}`registry.xml` file.
```
stevepiercy, line 90: I searched within the Classic UI folder for a reference to the file `registry.xml`, but came up empty. Is that the correct filename? There is documentation on TinyMCE in https://6.docs.plone.org/classic-ui/tinymce-customization.html that uses the filename `profiles/default/registry/tinymce.xml`.

Where can I get more information? We should add a `seealso` admonition to whatever that reference should be, in case someone wants to fiddle with TinyMCE. Here's my guess. Please let me know. Thank you!

````suggestion
    It requires additional configuration in the add-on's profile's registry.

    ```{seealso}
    {ref}`classic-ui-tinymce-customization-insert-preconfigured-html-blocks-label`
    ```
````

**Review by stevepiercy — CHANGES_REQUESTED** [review]
Almost there! A couple of questions regarding the `registry.xml` file for TinyMCE.

File: docs/classic-ui/theming/diazo.md (hunk 5 above)
Comment (erral) [reply docs/classic-ui/theming/diazo.md:90]: The file `registry.xml` present in `profiles/default` can be split in several files under `profiles/default/registry` with arbitrary names ending in `.xml`, so `registry.xml` and `profiles/default/registry/tinymce.xml` is valid and `profiles/default/registry/plone.xml` is valid.

The docs lack information about tinymce templates. I will add a PR with that information and then we can link there.

Comment (erral) [reply docs/classic-ui/theming/diazo.md:90]: @stevepiercy I have added the relevant information in the tinymce docs in this very PR. 

I tried to use the same example content as created by the plonecli template. 

Let's see if it is understandable :)

Comment (stevepiercy) [reply docs/classic-ui/theming/diazo.md:90]: Yup, this is great! I tidied it up a bit. However, you should take a look at a compatibility issue with Plone 6.1/TinyMCE 7 and the Template plugin in https://plone6--1870.org.readthedocs.build/classic-ui/tinymce-customization.html#insert-preconfigured-html-blocks. 😞 

I also revised the suggestion, and added your comment as a tip to the TinyMCE docs.

Comment (erral) [reply docs/classic-ui/theming/diazo.md:90]: Then we better add a warning saying that this only works for Plone 6.0 and not 6.1 right?

Comment (stevepiercy) [reply docs/classic-ui/theming/diazo.md:90]: Good question. There was a [similar issue with the accordion plugin](https://6.docs.plone.org/backend/upgrading/version-specific-migration/upgrade-to-61.html#tinymce-upgraded-in-classic-ui). It appears that there was work on the Template plugin in mockup in https://github.com/plone/mockup/pull/1384. @petschki can you advise?

I don't know if this requires an entry in the 6.1 upgrade guide.

Comment (petschki) [reply docs/classic-ui/theming/diazo.md:90]: the `template` plugin has moved to a premium plugin since tiny version `6.4` (see https://www.tiny.cloud/docs/tinymce/latest/advanced-templates/) but we've reimplemented the "community template plugin" for version 7.x in mockup. So its there and it works in Plone 6.1 too.

File: docs/classic-ui/tinymce-customization.md [hunk 6]
```diff
@@ -168,3 +167,46 @@ You can also exclude certain URLs from being sandboxed as follows.
...
+## Insert preconfigured HTML blocks
+
+You can add custom HTML blocks in TinyMCE.
+You can insert them in your content using the TinyMCE {menuselection}`Insert Template` menu option using the [TinyMCE Template plugin](https://www.tiny.cloud/docs/tinymce/6/template/).
+
+```{important}
+The Template plugin has been completely removed from the TinyMCE 7.0 release used in Plone 6.1.
+As an alternative solution, you can use the [Advanced Template Premium plugin](https://www.tiny.cloud/docs/tinymce/6/advanced-templates/). 
+```
+
```
petschki: This can be removed since we've backported the community edition of `template` plugin to TinyMCE 7

stevepiercy (approved):

docs/classic-ui/tinymce-customization.md
stevepiercy: Done in https://github.com/plone/documentation/pull/1870/commits/2d47570f0de137adfdaee38dbfed18bd47ed1888. Thanks @petschki! Will merge the chain of PRs when all green.

stevepiercy:
The failing test is a linkcheck failure that is fixed in the related PR. Merging! 🎉 Thanks @erral for all the work, and @petschki for the assist!

Issue #1882, opened 2025-03-02
Purge empty pages and todos


stevepiercy:
We've had empty pages and TODOs for over three years in Plone 6 Documentation, especially under the Backend section, but also under Deployment, Classic UI, Volto, and Contributing. The [Plone 6.1 milestone](https://github.com/plone/documentation/milestone/9) has only 17% of its issues completed _after_ the release of Plone 6.1. The lack of current and accurate documentation for Plone 6 is the single greatest barrier to entry for all audiences. This reflects extremely poorly upon Plone, where documentation is perceived as less than an afterthought.

The following list of URLs are affected.

Purge empty pages. A page is empty when either it has no content or its only content is a TODO. If there is no existing issue for the page, then create a new issue.

All other pages in the list have content with at least one TODO. For these pages, preserve the TODO. If the TODO does not contain a link to an existing issue, then create a new issue and add its URL to the TODO, except for obsolete TODOs, such as Upgrading Plone 4.x to 5.0.

Configure `conf.py` to hide the display of TODOs, changing code from the following...

https://github.com/plone/documentation/blob/b4462a84bac18a56ea8a37390cd87b027e8b85bd/docs/conf.py#L381-L385

...to the following.

```py
# -- Options for sphinx.ext.todo -----------------------

# See http://sphinx-doc.org/ext/todo.html#confval-todo_include_todos
todo_include_todos = False
todo_emit_warnings = True
todo_link_only = True

```

Ref: https://www.sphinx-doc.org/en/master/usage/extensions/todo.html#configuration

- [x] https://6.docs.plone.org/backend/annotations.html
- [ ] https://6.docs.plone.org/backend/behaviors.html#built-in-behaviors below the table. TODO: Linked issue is complete.
- [ ] https://6.docs.plone.org/backend/configuration-registry.html
- [ ] https://6.docs.plone.org/backend/control-panels.html - https://github.com/plone/documentation/pull/1909
- [ ] https://6.docs.plone.org/backend/portal-actions.html#expressions TODO
- [ ] https://6.docs.plone.org/backend/search.html TODO and empty headings
- [ ] https://6.docs.plone.org/backend/security.html
- [ ] https://6.docs.plone.org/backend/sending-email.html
- [x] https://6.docs.plone.org/backend/subscribers.html - https://github.com/plone/documentation/pull/1918
- [ ] https://6.docs.plone.org/backend/traversal-acquisition.html
[...]

PR #1885, opened 2025-03-03
Chapter ZODB

jensens:
see https://github.com/plone/documentation/issues/1882 - fill in some gaps

Other may have more ideas whats important here, but it covers the basics.

----
📚 Documentation preview 📚: https://plone6--1885.org.readthedocs.build/


jensens:
crap, I had the wrong base branch when branching locally. Will look at it later

PR #1910, opened 2025-03-14
touch `plone.api-install` to avoid running pip repeatedly

stevepiercy:
When there is a separate target to only install `plone.api`, it runs pip every time to see if it is installed, which can take a second or two. It's an annoyance that adds up.

So I modified the Makefile to fake it, by making the target `plone.api-install` create a file with the same name. Make will skip that target's commands when a file with the name of the target exists, and, of course, it's not a phony target.

On the first run, `plone.api-install` will create the file `plone.api-install` and install plone.api. On subsequent runs, since a file with the name of the target exists, the target will be skipped.

If you want to purge the installation, use `make distclean`.

Also moved `@echo`s to the last command for each target to make it easier to see its final statement and to group it together in the console.

Note: while coming up with this, I was listening to the Peter Gabriel Live album, "I Have the Touch". Subliminal messages... 🤯

----
📚 Documentation preview 📚: https://plone6--1910.org.readthedocs.build/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,5 @@ docs/plone.api
 docs/plone.restapi
 docs/volto
 
-# created on first make command
-plone.api-install
-
 # editor files
 .vscode
--- a/Makefile
+++ b/Makefile
@@ -35,22 +35,22 @@ distclean: clean ## Clean docs build directory, Python virtual environment, and 
 	rm docs/plone.api
 	rm docs/plone.restapi
 	rm docs/volto
-	@echo
 	@echo "Cleaned docs build directory, Python virtual environment, and symlinks."
+	@echo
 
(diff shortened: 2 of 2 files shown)


davisagli (changes requested):
The idea makes sense but:
1. distclean needs to delete this file
2. the file needs to be added to .gitignore
3. it could be nice to put it in `venv` instead of the top level, so it's out of the way (that would also take care of the first two bullet points since distclean already deletes venv and it is already gitignored)

stevepiercy:
> it could be nice to put it in `venv` instead of the top level, so it's out of the way (that would also take care of the first two bullet points since distclean already deletes venv and it is already gitignored)

FTW! Done in 490fd6486b647448460d40bf245efadd3a8ece54

davisagli (approved):

PR #1916, opened 2025-03-16
Remove the external link icon for internal links in submodules

stevepiercy:
Remove the external link icon from Plone Sphinx Theme for links in submodules that refer to Plone 6 Documentation via Intersphinx.

Although these links have `external` in their class, they are actually internal to Plone 6 Documentation.

For example, see the index page of Volto UI.

https://6.docs.plone.org/volto/

And see that the external link icon for "Plone content management system" is correct, but the link for "First-time contributors" is actually internal but has the external link icon.

----
📚 Documentation preview 📚: https://plone6--1916.org.readthedocs.build/volto/

### DIFF
--- a/docs/_static/documentation.css
+++ b/docs/_static/documentation.css
@@ -1,0 +1,17 @@
+/*
+  Remove the external link icon from Plone Sphinx Theme for links
+  in submodules that refer to Plone 6 Documentation via Intersphinx.
+  Although these links have `external` in their class, they are
+  actually internal to Plone 6 Documentation.
+*/
+.reference.external::after {
+  all: unset;
+}
+a:not([title="(in Plone Documentation v6)"]).reference.external::after {
+  margin-left: .3em;
+  display: inline-block;
+  content: var(--pst-icon-external-link);
+  white-space: nowrap;
+  font: var(--fa-font-solid);
+  font-size: .75em;
+}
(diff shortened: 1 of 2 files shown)


stevepiercy:
@Manas-Kenge would you please take a look? It's an interesting side effect of using Intersphinx with submodules.

ksuess:
It's more than the icon. 
Tippy does not tip.

stevepiercy:
Tippy does not work for external links or Intersphinx links.

The only fix is to stop using submodules and author documentation only in this repo, but that's a battle for another day.

Manas-Kenge:
According to [this](https://www.sphinx-doc.org/en/master/usage/quickstart.html#intersphinx), Sphinx-Tippy can be used to create tooltips for external documentation, but it requires some configuration. I guess it's a bit of a hassle for now.

stevepiercy:
Sphinx-Tippy supports only [limited external documentation](https://sphinx-tippy.readthedocs.io/en/latest/#external-apis), specifically RTD, Wikipedia, and DOI hosted docs.

stevepiercy:
Just to clarify, Sphinx-Tippy works for internal links, but not internal Intersphinx links.

Screenshot 2025-03-16 at 5 01 18 PM

I imagine it could, however, be modified to work with any arbitrary Intersphinx reference.

PR #1962, opened 2025-07-17
Fix Typos in recipes.md

krissik:
Just two small typo fixes

----
📚 Documentation preview 📚: https://plone6--1962.org.readthedocs.build/


stevepiercy (approved):
Thank you!

We should definitely run code samples before releasing them to unsuspecting readers.

PR #1981, opened 2025-10-26
Explain how to switch to add native namespace

gforcada:
To go along with the https://github.com/plone/Products.CMFPlone/issues/3928

Disclaimer: I have no idea where would be the best place to put this documentation within the structure, and probably the admonitions style is not the right one.

We can polish, but before nitpicking on this details, let's agree on the structure of the document and if it explains all the relevant steps and are clear enough.

Once that's done we can do the polish.

@mauritsvanrees @davisagli: given a random Plone contributor, do you think the steps cover all the steps, and are clear enough? I made it generic enough (with tips to skip sections) so that it can be reused for `collective` or private packages, where we can not convert all ourselves.

----
📚 Documentation preview 📚: https://plone6--1981.org.readthedocs.build/

### DIFF
--- a/docs/developer-guide/native-namespaces.md
+++ b/docs/developer-guide/native-namespaces.md
@@ -1,0 +1,201 @@
+# Native namespace
+
+This document explains the steps needed to convert a python distribution from `pkg_resources` namespace to native namespaces.
+
+## Background
+
+Python, since Python 3.3, added support for native namespaces, see [PEP 420](https://peps.python.org/pep-0420/).
+
+Plone has been using `pkg_resources`-style namespaces, but they are deprecated in `setuptools`.
+
+`setuptools` is planning to remove `pkg_resources`'s namespaces support by the end of 2025.
+
+[PLIP 3928](https://github.com/plone/Products.CMFPlone/issues/3928) tracks the changes needed for Plone to adapt to the _new_ native namespaces.
+
+## Steps
+
+To convert a given (`$package`) python distribution to use native namespaces follow these steps.
+
+### Create maintenance branch
+
+```{note}
+Only relevant for Plone core packages
+```
+
+```{tip}
+This part is only needed when the `main` or `master` branch is used on multiple versions of the Plone core development buildout.
+```
+
+Clone the repository or ensure you are at latest changes:
+
+```shell
+git clone git@github.com:plone/$package
+cd $package
+# or update
+git fetch -p
+git checkout main # or master
+git rebase
+```
+
+Find out what's the last release and create a branch for it:
+
+```shell
+# list all tags
+git for-each-ref --sort=taggerdate --format '%(tag)' refs/tags
+# get the last tag's major number
+MAJOR=`git for-each-ref --sort=taggerdate --format '%(tag)' refs/tags | tail -n1 | cut -d"." -f1`
+# create a branch for it
+git checkout -b $MAJOR.x
+# push the newly created branch
+git push
+```
+
+### Update buildout.coredev
+
+```{note}
+Only relevant for Plone core packages
+```
+
+Update `buildout.coredev`'s branch 6.1 to use the newly created branch
+
+```shell
+# go to the repository or clone it
+cd buildout.coredev
+# ensure you are at latest changes
+git fetch -p
+git checkout 6.1
+git rebase
+# update the branch being used by the python distribution
+sed -i "s/$package.git branch=master/$package.git branch=$MAJOR.x/" sources.cfg
+# add the changes, commit and push
+git add sources.cfg
+git commit -m"chore: use branch $MAJOR.x for $package"
+git push
+```
+
+Now check if you need to do the same on the 6.0 branch of `buildout.coredev`.
+
+```{tip}
+You can use this [handy table](https://jenkins.plone.org/roboto/branches) to know which branch is used of a given package on each Plone version
+```
+
+```{tip}
+To lower the amount of builds in Jenkins, either do a few at a time or add a `[ci-skip]` on the commit message
+```
+
+### Numbers before
+
+One risk of changing to the native namespaces is that some files, or tests, might be left behind.
+
+To ensure all tests and files are kept after the switch, gather the numbers before the change:
+
+Get the list of tests:
+
+```shell
+tox run -e test -- --list-tests | wc -l
+```
+
+```{note}
+Adapt to whichever way you are using to run the tests.
+The above is meant for repositories that follow `plone.meta` conventions.
+The `--list-tests` comes from `zope.testrunner`, if you are using that but not `plone.meta` you can use that as well.
+```
+
+Create a distribution to get a listing of how many files are currently packaged:
+
+```shell
+rm -rf dist/
+uvx --from build pyproject-build
+```
+
+A `dist` folder is created with two archives in it: a `.tar.gz` and a `.whl`.
+
+To get the number of files on them run the following commands:
+
+```shell
+python -c "import glob; import tarfile; print(len(tarfile.open(glob.glob('dist/*.tar.gz')[0], 'r:gz').getnames()))"
+python -c "import glob; from zipfile import ZipFile; print(len(ZipFile(glob.glob('dist/*.whl')[0]).namelist()))"
+```
+
+Keep these numbers around for later.
+
+### Build backend
+
+To ensure the package continues to build, ensure that `setuptools` is defined as its build backend.
+
+For that inspect the `pyproject.toml`, it should have these lines:
+
+```toml
+[build-system]
+requires = ["setuptools>=68.2,<80", "wheel"]
+```
+
+If they are not there, add them, commit and push the changes.
+
+### Convert to native namespace
+
+Use `plone.meta`'s `switch-to-pep420` script:
+
+```shell
+cd $package
+uvx --from plone.meta switch-to-pep420 --no-tests .
+```
+
+```{tip}
+This will also bump the version to a new major release.
+
+If the `main` or `master` branch is already an alpha version that is only used in Plone 6.2, you can specify that you don't want this version bump by adding the `--no-breaking` option.
+```
+
+### Update the test matrix
+
+```{note}
+Only relevant for Plone core packages
+```
+
+As the switch to the native namespace has to be coordinated, all python distributions need to be only for the same Plone version, in this case it was decided to do it for the Plone 6.2 version.
+
+Thus, we need to ensure that the test matrix, only tests against this Plone version.
+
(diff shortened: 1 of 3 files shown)


stevepiercy:
@gforcada this is a how-to guide for developers, so it belongs in `docs/developer-guide` with a toctree entry in `docs/developer-guide/index`. Would you please move it there?

I'll do a thorough non-technical review after the technical structure and details are reviewed and approved.

Thank you!

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+
+```shell
+# list all tags
+git for-each-ref --sort=taggerdate --format '%(tag)'
+# get the last tag's major number
+MAJOR=`git for-each-ref --sort=taggerdate --format '%(tag)' | tail -n1 | cut -d"." -f1`
+# create a branch for it
+git checkout -b $MAJOR.x
+# push the newly created branch
+git push $MAJOR.x
```
mauritsvanrees: This complains that for example 4.x is not a remote.  This seems enough:

```suggestion
git push
```

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+git fetch -p
+git checkout main # or master
+git rebase
+```
+
+Find out what's the last release and create a branch for it:
+
+```shell
+# list all tags
+git for-each-ref --sort=taggerdate --format '%(tag)'
```
mauritsvanrees: I have not used the `for-each-ref` command before, thanks for sharing.
When I try it, I get some empty lines as well, for references in `refs/heads` and `refs/remotes`.  We can restrict to only tags:

```suggestion
git for-each-ref --sort=taggerdate --format '%(tag) refs/tags
```

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+# get the last tag's major number
+MAJOR=`git for-each-ref --sort=taggerdate --format '%(tag)' | tail -n1 | cut -d"." -f1`
```
mauritsvanrees: Same here for good measure, only tags:

```suggestion
MAJOR=`git for-each-ref --sort=taggerdate --format '%(tag)' refs/tags | tail -n1 | cut -d"." -f1`
```

docs/developer-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+## Steps
+
+To convert a given (`$package`) python distribution to use native namespaces follow these steps.
+
+### Create maintenance branch
+
+!!! note
+    Only relevant for Plone core packages
+
+Clone the repository or ensure you are at latest changes:
```
mauritsvanrees: I would add a line of explanation:

```suggestion
This part is only needed when the `main` or `master` branch is used on multiple versions of the Plone core development buildout.

Clone the repository or ensure you are at latest changes:
```

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+git rebase
+# update the branch being used by the python distribution
+sed -i "s/$package.git branch=master/$package.git branch=$MAJOR.x/" sources.cfg
+# add the changes, commit and push
+git add sources.cfg
+git commit -m"chore: use branch $MAJOR.x for $package"
+git push
+```
+
+!!! note
```
mauritsvanrees:
```suggestion
Now check if you need to do the same on the 6.0 branch of `buildout.coredev`.

!!! note
```

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+    To lower the amount of builds in Jenkins, either do a few at a time or add a `[ci-skip]` on the commit message
+
+### Numbers before
+
+One risk of changing to the native namespaces is that some files, or tests, might not be left behind.
```
mauritsvanrees: I suppose the main risk of this was actually already when moving to a src-layout.

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+```shell
+uvx --from build pyproject-build
+```
+
+A `dist` folder is created with two archives in it: a `.tar.gz` and a `.whl`.
+
+To get the number of files on them run the following commands:
+
+```shell
+python -c "import tarfile; print(len(tarfile.open('dist/$package.tar.gz', 'r:gz').getnames()))"
```
mauritsvanrees:
```suggestion
python -c "import glob; import tarfile; print(len(tarfile.open(glob.glob('dist/*.tar.gz')[0], 'r:gz').getnames()))"
```

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+python -c "from zipfile import ZipFile; print(len(ZipFile('dist/$package.whl').namelist()))"
```
mauritsvanrees:
```suggestion
python -c "import glob; import zipfile; print(len(zipfile.ZipFile(glob.glob('dist/*.whl')[0]).namelist()))"
```

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+Likewise, create the distribution files and compare the numbers with the previous run:
+
+```shell
+python -c "import tarfile; print(len(tarfile.open('dist/$package.tar.gz', 'r:gz').getnames()))"
+python -c "from zipfile import ZipFile; print(len(ZipFile('dist/$package.whl').namelist()))"
+```
+
+If all numbers match, review the changes once more, push the branch with the changes for others to review it.
```
mauritsvanrees: There will be a few less files in the new distributions.

```suggestion
It is okay if the numbers are slightly lower.
For obvious reasons the old source distribution will have one or more extra `__init__.py` files.
Both old distributions are expected to have an extra `namespace_packages.txt` file and possible an `x-nspkg.pth` file.
If you lose more than a few files though, something is wrong.
If the numbers are close enough, review the changes once more, and push the branch with the changes for others to review it.
```

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+
+```toml
+[tox]
+test_matrix = {"6.2" = ["*"]}
+```
+
+And update the scaffolding files with `plone.meta`:
+
+```shell
+uvx --from plone.meta config-package .
```
mauritsvanrees: In my initial comment in https://github.com/plone/meta/pull/297 I first run config-package passing a specific branch name matching the one used by the pep-420 script, then update the test matrix, run config-package again, and finally run the pep-420 script.

You run the pep-420 script first, which I think makes sense.  But then we should tell config-package to use the current branch:


```suggestion
uvx --from plone.meta config-package --branch current .
```

docs/reference-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+
+!!! note
+    If the diff is quite big, run `config-package` before all the changes, get that on a Pull Request, approved and merged and then do a follow up Pull Request to move to native namespace.
+
+### Compare numbers
+
+Get the list of tests, like before and compare the lists to ensure that the same amount of tests are found.
+
+```shell
+tox run -e test -- --list-tests
```
mauritsvanrees: I think we can just get the number of lines:

```suggestion
tox run -e test -- --list-tests | wc -l
```

docs/developer-guide/native-namespaces.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Convert to native namespace
+
+Use `plone.meta`'s `switch-to-pep420` script:
+
+```shell
+cd $package
+uvx --from plone.meta switch-to-pep420 .
+```
+
```
mauritsvanrees:
```suggestion

Note that this will also bump the version to a new major release.
If the `main` or `master` branch is already an alpha version that is only used in Plone 6.2, you can specify that you don't want this version bump by adding the `--no-breaking` option.

```

I should probably add this logic to the `bumpversion` command from `zest.releaser`.

mauritsvanrees (commented):
Overall good, but I made some suggestions.

docs/reference-guide/native-namespaces.md
gforcada: Oh nice, I did not know that you can further filter by refs 🎉

docs/developer-guide/native-namespaces.md
gforcada: Well, I add the `Only relevant for Plone core packages, but sure, extra tips can never hurt :)

docs/reference-guide/native-namespaces.md
gforcada: oh, right, so we can drop this parts then, which makes it all much simpler and straightforward 😄

docs/reference-guide/native-namespaces.md
gforcada: Should we keep this section or remove it, as your previous suggestion was that it was not that problematic anymore, as we already moved to the `src/` layout?

🤔 as the idea is to make it relevant for collective and private repositories, we might keep it, so it also helps there, where they might not have moved to a `src` layout.

gforcada:
@mauritsvanrees thanks for all the tips and improvements! 🙇🏾 that's much better ✨ 

@stevepiercy I moved the file to the location you suggested and add a ToC entry.

Now is probably the time to improve the wording, which probably needs plenty 😅

stevepiercy:
@gforcada I requested another review from @mauritsvanrees, and I'd like @davisagli to review as well, before I make my final pass for non-technical stuff.

docs/developer-guide/native-namespaces.md
```diff
@@ -0,0 +1,201 @@
...
+### Update the test matrix
+
+```{note}
+Only relevant for Plone core packages
+```
+
+As the switch to the native namespace has to be coordinated, all python distributions need to be only for the same Plone version, in this case it was decided to do it for the Plone 6.2 version.
+
+Thus, we need to ensure that the test matrix, only tests against this Plone version.
```
mauritsvanrees: Superfluous comma.

```suggestion
Thus, we need to ensure that the test matrix only tests against this Plone version.
```

mauritsvanrees (dismissed):
LGTM.  One minor textual suggestion.

davisagli (commented):
Content looks good to me

stevepiercy (approved):
Thanks for the guide! I made a few minor enhancements for readability, grammar, and syntax. Will merge when green.

PR #2002, opened 2025-12-10
Create upgrade guide for Plone 6.2 backend.

mauritsvanrees:
For technical review, I tag @davisagli @gforcada @petschki.

I have fixed most Vale feedback, except these, which I don't know how to write better:

```
venv/bin/vale --no-wrap  docs/backend/upgrading/version-specific-migration/upgrade-to-62.md

 docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
 33:11  suggestion  'is deprecated' looks like passive voice.    Microsoft.Passive         
 33:44  suggestion  'be removed' looks like passive voice.       Microsoft.Passive         
 62:29  warning     Avoid using acronyms in a title or heading.  Microsoft.HeadingAcronyms 
```

That last one is for "Classic UI" in the last header, I think.

----
📚 Documentation preview 📚: https://plone6--2002.org.readthedocs.build/

Specifically this page: [`/backend/upgrading/version-specific-migration/upgrade-to-62`](https://plone6--2002.org.readthedocs.build/backend/upgrading/version-specific-migration/upgrade-to-62.html).

### DIFF
--- a/docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
+++ b/docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
@@ -1,0 +1,83 @@
+---
+myst:
+  html_meta:
+    "description": "How to upgrade to Plone 6.2"
+    "property=og:description": "How to upgrade to Plone 6.2"
+    "property=og:title": "How to upgrade to Plone 6.2"
+    "keywords": "Upgrade, Plone 6"
+---
+
+(backend-upgrade-plone-v62-label)=
+
+# Upgrade Plone 6.1 to 6.2
+
+Plone 6.2 has seen the following major changes.
+Some may require changes in your setup.
+
+
+## Move to native namespaces
+
+Plone 6.2 has migrated all core Python packages from `pkg_resources`-style namespaces to native namespaces.
+
+The Plone Python backend consists of over 200 packages.
+Most of these are namespace packages.
+Examples are `plone.batching` and `plone.api`, which share the `plone` namespace.
+
+There are three styles or techniques of namespaces.
+For Plone only two are important: `pkg_resources` and native (sometimes referred to as implicit).
+The third style is `pkgutil`, but Plone has never used it.
+
+Native namespaces exist since Python 3.3.
+Because Plone started in the days of Python 2, it has always used `pkg_resources`.
+`pkg_resources` is part of the `setuptools` package.
+This part is deprecated: it's scheduled to be removed around the end of 2025, in `setuptools` 81.
+This means Plone needs to move to native namespaces.
+
+In general, this move shouldn't cause problems for integrators.
+To install Plone 6.2 you can keep using the same version of `pip` (or `uv`), or `zc.buildout` as you do for Plone 6.1.
+This assumes that all packages within one namespace use the same namespace style or technique.
+
+You may encounter problems when this assumption isn't true.
+As an example, say your Plone project uses a package `collective.example`, and this still uses `pkg_resources`-style namespaces.
+Plone itself depends on a package `collective.monkeypatcher` which is now using native namespaces.
+This may cause problems at startup, especially `ModuleNotFoundError`.
+In a normal install:
+
+* `pip` works fine.
+* `zc.buildout` 4 won't work.
+* `zc.buildout` 5 works fine.
+
+If for one or both of these packages you don't use a final release from https://pypi.org, but an editable install, it may not work.
+You can solve this by installing an extra package with `pip`: `horse-with-no-namespace`.
+
+```{note}
+If you let `zc.buildout` install `horse-with-no-namespace`, it won't work.
+It really must be installed by `pip` into the same virtual environment where you have installed `zc.buildout`.
+```
+
+The recommendation is to migrate all namespace packages to native namespaces.
+This includes the `collective` namespace that a lot of Plone add-ons use.
+This {doc}`guide </developer-guide/native-namespace>` describes the needed steps.
+
+## Start separating Classic UI code from the core
+
+In Plone 6, the Python backend includes the Classic UI frontend.
+You can't get one without the other.
+If you use the Volto frontend, this means that you get lots of code that you never use.
+
+Plone 6.2 starts separating Classic UI code from the core of Plone.
+This is ongoing work, described in {term}`PLIP` [3953](https://github.com/plone/Products.CMFPlone/issues/3953).
+You won't yet get less code: this is preparation for a more complete separation in Plone 7.
+
+The focus currently is on moving page templates into the `plone.app.layout` package.
+Templates from the following packages are now in a new location:
+
+* `plone.locking`
+* `plone.protect`
+
+```{note}
+If you use the `z3c.jbot` add-on to override a template that has been moved, your override will still work.
+This is because we keep a mapping from the old to the new location.
+For example, `plone.locking` registers that `plone.locking.browser.info.pt` has a new location: `plone.app.layout.viewlets.locking.pt`.
+You should rename your override to the new location if you no longer need compatibility with Plone 6.1 or earlier.
+```
--- a/docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
+++ b/docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
@@ -24,8 +24,16 @@ Most of these are namespace packages.
 Examples are `plone.batching` and `plone.api`, which share the `plone` namespace.
 
 There are three styles or techniques of namespaces.
-For Plone only two are important: `pkg_resources` and native (sometimes referred to as implicit).
+For Plone only two are important: `pkg_resources` and native.
 The third style is `pkgutil`, but Plone has never used it.
+
+Native namespaces are also referred to as implicit namespaces.
+The two terms mean the same.
+
+```{seealso}
+- [Python Packaging Guide on native namespaces](https://packaging.python.org/en/latest/guides/.packaging-namespace-packages/#native-namespace-packages)
+- [PEP 420 - Implicit Namespace Packages](https://peps.python.org/pep-0420/)
+```
 
 Native namespaces exist since Python 3.3.
 Because Plone started in the days of Python 2, it has always used `pkg_resources`.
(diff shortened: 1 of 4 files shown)


stevepiercy:
You can ignore Vale warnings and suggestions, but errors should get a look. As part of my routine maintenance, I try to address the errors and reduce them. There's a long way to go yet, so the only thing I really care about for now is to not introduce new errors. Often it turns out to be a candidate for a new term in the glossary, and sometimes it's just some bad MyST markup.

I'll do a thorough editorial review in the next day or so, after the smart people do a technical review. Thanks for the PR!

gforcada (dismissed):
Nice write up!

petschki (dismissed):
thanks @mauritsvanrees ... currently there is nothing more to say for the template movements. Progress is tracked here https://github.com/plone/Products.CMFPlone/issues/3953

styles/config/vocabularies/Plone/accept.txt
```diff
@@ -28,6 +28,8 @@ jQuery
 libxslt
 middleware
 Mockup
+namespace
+namespaces
```
stevepiercy: Vale supports regular expressions.

```suggestion
namespaces?
```

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
```diff
@@ -0,0 +1,83 @@
...
+
+Plone 6.2 has migrated all core Python packages from `pkg_resources`-style namespaces to native namespaces.
+
+The Plone Python backend consists of over 200 packages.
+Most of these are namespace packages.
+Examples are `plone.batching` and `plone.api`, which share the `plone` namespace.
+
+There are three styles or techniques of namespaces.
+For Plone only two are important: `pkg_resources` and native (sometimes referred to as implicit).
+The third style is `pkgutil`, but Plone has never used it.
```
stevepiercy, line 28: We should include references to the following terms.

- [Native namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages)
- [PEP 420 – Implicit Namespace Packages](https://peps.python.org/pep-0420/)

If I read all that correctly, a native namespace is a subcategory of implicit namespaces. Is that right?

If so, then I'd make the following suggestion.

````suggestion
There are three styles or techniques of namespaces.
For Plone, only two are important: `pkg_resources` and native, the latter of which is a subcategory of implicit namespace packages where you omit the `__init.py__` file from the namespace package directory.
The third style is `pkgutil`, but Plone has never used it.

```{seealso}
- [Native namespace packages](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages)
- [PEP 420 – Implicit Namespace Packages](https://peps.python.org/pep-0420/)
```
````

File: docs/backend/upgrading/version-specific-migration/upgrade-to-62.md [hunk 3]
```diff
@@ -0,0 +1,83 @@
...
+
+Native namespaces exist since Python 3.3.
+Because Plone started in the days of Python 2, it has always used `pkg_resources`.
+`pkg_resources` is part of the `setuptools` package.
+This part is deprecated: it's scheduled to be removed around the end of 2025, in `setuptools` 81.
```
stevepiercy:
```suggestion
This part is deprecated.
It's scheduled to be removed around the end of 2025, in `setuptools` 81.
```

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
```diff
@@ -0,0 +1,83 @@
...
+Plone itself depends on a package `collective.monkeypatcher` which is now using native namespaces.
+This may cause problems at startup, especially `ModuleNotFoundError`.
+In a normal install:
+
+* `pip` works fine.
+* `zc.buildout` 4 won't work.
+* `zc.buildout` 5 works fine.
+
+If for one or both of these packages you don't use a final release from https://pypi.org, but an editable install, it may not work.
+You can solve this by installing an extra package with `pip`: `horse-with-no-namespace`.
```
stevepiercy:
```suggestion
You can solve this by installing an extra package with `pip`, `horse-with-no-namespace`.
```

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Start separating Classic UI code from the core
+
+In Plone 6, the Python backend includes the Classic UI frontend.
+You can't get one without the other.
+If you use the Volto frontend, this means that you get lots of code that you never use.
+
+Plone 6.2 starts separating Classic UI code from the core of Plone.
+This is ongoing work, described in {term}`PLIP` [3953](https://github.com/plone/Products.CMFPlone/issues/3953).
+You won't yet get less code: this is preparation for a more complete separation in Plone 7.
```
stevepiercy:
```suggestion
You won't yet get less code yet.
This PLIP is preparation for a more complete separation in Plone 7.
```

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
```diff
@@ -0,0 +1,83 @@
...
+The focus currently is on moving page templates into the `plone.app.layout` package.
+Templates from the following packages are now in a new location:
+
+* `plone.locking`
+* `plone.protect`
+
+```{note}
+If you use the `z3c.jbot` add-on to override a template that has been moved, your override will still work.
+This is because we keep a mapping from the old to the new location.
+For example, `plone.locking` registers that `plone.locking.browser.info.pt` has a new location: `plone.app.layout.viewlets.locking.pt`.
```
stevepiercy:
```suggestion
For example, `plone.locking` registers that `plone.locking.browser.info.pt` has a new location, `plone.app.layout.viewlets.locking.pt`.
```

stevepiercy (changes requested):
Minor tweaks, and a question and suggestion for reference to Python docs. I'm not sure if I got it right. Thanks for your work on this. It's really good!

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
stevepiercy: @mauritsvanrees I committed all but this one suggestion, as I'd like your thoughts on it.

davisagli: This suggestion seems helpful to me.

davisagli (commented):
Aside from the pending suggestion, this looks good to me.

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
mauritsvanrees: Adding the links is good.

But I don't see where you got the impression that native namespace are a subset of implicit namespaces.  As far as I know, there is no difference, they are two terms for the exact same thing.

Your [first link](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages) says: "Python 3.3 added implicit namespace packages from [PEP 420](https://peps.python.org/pep-0420/). All that is required to create a native namespace package is that you just omit `__init__.py` from the namespace package directory."  The text points to your second link, the official PEP, and the term "native" is not even mentioned there.

I asked Duck.AI, but that answer was just clearly wrong, restricting native to namespaces that exist within one package and still have an `__init__.py`, though empty.

@davisagli Do you know a difference between the two?

mauritsvanrees:
I have rebased on 6.0 and force pushed, because the merge and revert of the multilingual PR which added the same file, gave a conflict.

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
stevepiercy: > Your [first link](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages) says: "Python 3.3 added implicit namespace packages from [PEP 420](https://peps.python.org/pep-0420/). All that is required to create a native namespace package is that you just omit `__init__.py` from the namespace package directory." The text points to your second link, the official PEP, and the term "native" is not even mentioned there.

Yes, that's exactly why I was uncertain. I can ask in Python's doc Discord channel to see what they say, too.

petschki:
I've created a TinyMCE license key config documentation here https://github.com/plone/documentation/pull/2012 

we should mention here, that TinyMCE got updated to v8 and that you're able to provide commercial tinymce license keys now.

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
davisagli: I misread Steve's suggestion. Native namespace packages are not a subcategory of implicit namespaces. They are 2 synonyms for the same kind (which is defined in PEP 420). They are implicit because you don't have to explicitly declare them in `__init__.py`, and they are native because they are built in to Python without installing an extra package like setuptools.

mauritsvanrees: I have rephrased it in commit 0a1ee963b147be7eb4180c4bbf914da0a04b7e24.

stevepiercy: I also got confirmation from the Python Docs Discord server members that implicit and native are synonymous, and I'll submit a PR there to replace "native" with "implicit". No one seems to know for sure where "native" came from, other than it's a "batteries included" thing. "implicit" is preferred and the only term used in PEP 420.

stevepiercy:
> I've created a TinyMCE license key config documentation here #2012
> 
> we should mention here, that TinyMCE got updated to v8 and that you're able to provide commercial tinymce license keys now.

@petschki would you please amend your PR to do just that? It'd be better to keep the docs for a given feature together in the same PR.

stevepiercy (approved):

docs/backend/upgrading/version-specific-migration/upgrade-to-62.md
mauritsvanrees: In Plone, Zope, and Buildout, as far as I am aware we have always called them "native" the past years, in basically all PRs, issues, and changelog entries.  I have hardly ever seen the term "implicit", or used it myself.  So in the Plone docs I would prefer to keep using the term "native".

stevepiercy: Yeah, I see that Plone is a rebel, which is why I wasn't going to demand a change in our docs. 😉

Python docs, however, ought not to throw out terms without at least a glossary entry. They should know better.

PR #2005, opened 2025-12-14
Align wheel version with dist for 6.1.3

stevepiercy:
From https://dist.plone.org/release/6.1.3/constraints.txt

----
📚 Documentation preview 📚: https://plone6--2005.org.readthedocs.build/

### DIFF
--- a/requirements-initial.txt
+++ b/requirements-initial.txt
@@ -2,4 +2,4 @@
 # From https://dist.plone.org/release/6.1.3/constraints.txt
 pip==25.2
 setuptools==80.9.0
-wheel==0.46.1
+wheel==0.45.1


davisagli (commented):
@stevepiercy I'd suggest replacing everything in this file with `-r https://dist.plone.org/release/6.1.3/requirements.txt` to include the requirements defined there.

stevepiercy:
@davisagli wouldn't that include everything to set up Plone, when that might not be what is needed to build only documentation? I know that we already need to build plone.api and plone.restapi, so maybe the difference, if any, is negligible. I don't know how to audit such a thing.

Also in those two packages, we have different constraints.

```
# plone.api chains
# submodules/plone.api/requirements.txt
-c constraints.txt
# submodules/plone.api/constraints.txt
-c constraints_plone62.txt
# submodules/plone.api/constraints_plone62.txt
-c https://dist.plone.org/release/6.2-latest/constraints.txt

# submodules/plone.api/requirements-docs.txt
-c constraints_plone62.txt  # See previous file



# plone.restapi chains
# submodules/plone.restapi/requirements.txt
-c constraints_plone62.txt
# submodules/plone.restapi/constraints_plone62.txt doesn't exist!

# submodules/plone.restapi/requirements-docs.txt
# no constraints, but I went down a rabbit hole due to its dependency on setuptools for `pkg_resources`.
```

For the last one, see https://github.com/plone/plone.restapi/issues/1967.

I'm not sure how to proceed.

davisagli:
@stevepiercy The purpose of requirements-initial.txt in this repository is just to install the basic install tools (pip & co) with pinned versions, not everything for building documentation. The purpose of requirements.txt from dist.plone.org is similar (it's NOT all of Plone) but it also includes buildout. So I don't think it would hurt to use the requirements.txt from dist.plone.org, but it's ok not to if you want to avoid installing buildout.

> Also in those two packages, we have different constraints.

That's to be expected, since the default branches of those packages are now for Plone 6.2. You can decide whether to update the documentation repo to Plone 6.2, or use those packages from the recently created branches that are for Plone 6.1. (And in the short term, as long as nothing's breaking, it's probably fine as is. There haven't been significant changes in these packages for Plone 6.2 aside for the namespace package thing.)

```
# plone.restapi chains
# submodules/plone.restapi/requirements.txt
-c constraints_plone62.txt
# submodules/plone.restapi/constraints_plone62.txt doesn't exist!
```

That's not what I see here: https://github.com/plone/plone.restapi/blob/main/requirements.txt

stevepiercy:
OK, I think I'll leave the docs requirements-initial.txt as is.

> ```
> # plone.restapi chains
> # submodules/plone.restapi/requirements.txt
> -c constraints_plone62.txt
> # submodules/plone.restapi/constraints_plone62.txt doesn't exist!
> ```
> 
> That's not what I see here: https://github.com/plone/plone.restapi/blob/main/requirements.txt

That file exists, but it's supposed to find this file, which is missing!

https://github.com/plone/plone.restapi/blob/main/constraints_plone62.txt

davisagli:
@stevepiercy You're seeing something different than I am. https://github.com/plone/plone.restapi/blob/main/requirements.txt contains:
```
-c constraints.txt
plone.restapi[test]
```
with no reference to constraints_plone62.txt

stevepiercy:
After double-checking, it all aligns. I must have done a bad copy-paste. Sorry for the noise, and thanks for double-checking.

PR #2030, opened 2026-01-09
Update 6.2 release requirements for Volto and Classic UI

stevepiercy:
See https://github.com/plone/Products.CMFPlone/issues/4251

----
📚 Documentation preview 📚: https://plone6--2030.org.readthedocs.build/


petschki (approved):

mauritsvanrees (commented):
I don't know this part.

stevepiercy:
@mauritsvanrees @petschki I found it when I searched for "version" in the documentation repo. This is one of those details that should be explicitly included in every release, including minor and patch versions. I want to expand the details of the release checklist for documentation, so we don't forget. "If it ain't documented, it's broken."

Of course, it's important that the Classic UI Team to update this page whenever they update its dependencies, too. We all need to do better. I'll do my best to cover our collective butts, but it'd be even better to keep things current for the general public.

davisagli (approved):

PR #2057, opened 2026-02-13
Replace Plone 5 links with Plone 6 links in views chapter

MrTango:
Updated links to Plone 6 documentation and replaced direct URLs with internal documentation references.

----
📚 Documentation preview 📚: https://plone6--2057.org.readthedocs.build/

### DIFF
--- a/docs/classic-ui/views.md
+++ b/docs/classic-ui/views.md
@@ -715,17 +715,17 @@ However, in most cases:
 Replace links to Plone 5.2 docs with links to Plone 6 docs.
 Specifically:
 
--   [TAL page template](https://5.docs.plone.org/adapt-and-extend/theming/templates_css/template_basics)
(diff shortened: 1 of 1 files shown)


petschki (dismissed):

docs/classic-ui/views.md
```diff
@@ -715,17 +715,17 @@ However, in most cases:
 Replace links to Plone 5.2 docs with links to Plone 6 docs.
```
davisagli: @MrTango Shouldn't this todo be removed as part of this?

MrTango: the template link, yes, good catch.
The interfaces reference could stay as is too, there is a refernece to backend/interfaces which we don't have yet, but it's not rendered, so fine with me. We can live with the link to github.

davisagli (approved):

PR #2071, opened 2026-03-25
Prepare for Plone 6.2 release

stevepiercy:
DO NOT MERGE until Plone 6.2 final is released.

@mauritsvanrees here ya go!

Merge https://github.com/plone/volto/pull/8051 before this one, so that I can update its submodule.

----
📚 Documentation preview 📚: https://plone6--2071.org.readthedocs.build/

### DIFF
--- a/docs/_inc/_install-python-plone62.md
+++ b/docs/_inc/_install-python-plone62.md
@@ -1,5 +1,5 @@
 Installing Python is beyond the scope of this documentation.
-However, it is recommended to use a Python version manager, {term}`pyenv`, that allows you to install multiple versions of Python on your development environment without destroying your system's Python.
+However, it is recommended to use a Python version manager, such as {term}`pyenv`, that allows you to install multiple versions of Python on your development environment without destroying your system's Python.
 
 ```{warning}
 Do not create or activate a Python virtual environment at this time.
--- a/docs/_inc/_install-python-plone62.md
+++ b/docs/_inc/_install-python-plone62.md
@@ -6,4 +6,4 @@ Do not create or activate a Python virtual environment at this time.
 The instructions below will create one.
 ```
 
-Plone 6.1 requires Python version {{SUPPORTED_PYTHON_VERSIONS_PLONE61}}.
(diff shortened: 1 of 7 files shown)


mauritsvanrees (dismissed):
LGTM, thanks.

I think the changes you made to `docs/backend/upgrading/version-specific-migration/upgrade-to-62.md` and `docs/conf.py` can already be merged.

I guess theoretically we should wait with those until the next release is out, whether that is beta 1 or RC 1, because the current alpha 1 does not officially support Python 3.14.  But I don't think we need to do that.  And I suppose we don't want to litter this with notes like "added in beta 1" which no one will care about once the final release is there.  Then again, we could clean that up after a brief check after any release.

Does it make sense to add a note at the top of the 6.2 upgrade doc saying "This release is not final yet, some of the notes in here may be for a version that is soon to be released." ?

I may be overthinking this. :-)

stevepiercy:
I'd wait for a final release to avoid the need to explain details and adding confusion. Bleeding edge developers already know what's going on, but not the general public.

docs/_inc/_install-python-plone62.md
```diff
@@ -0,0 +1,9 @@
+Installing Python is beyond the scope of this documentation.
+However, it is recommended to use a Python version manager, such as {term}`pyenv`, that allows you to install multiple versions of Python on your development environment without destroying your system's Python.
```
davisagli: I suggest we mention uv as an option here too:

```suggestion
However, it is recommended to use a Python version manager, such as {term}`uv` or {term}`pyenv`, that allows you to install multiple versions of Python on your development environment without destroying your system's Python.
```

stevepiercy: Yeah, we might as well sneak that one in there, too. Thank you!

mauritsvanrees:
I would still like to merge the `docs/backend/upgrading/version-specific-migration/upgrade-to-62.md` into the 6.0 branch already.  Or I guess after the RC 2 that I plan for this week.  Then the upgrade notes are accurate for the latest release.

stevepiercy:
I messed up by committing a submodule see #2080 with the same changes in a single commit without the submodule.

PR #2072, opened 2026-03-25
Update docs to cover uv managed installations

### LINKED ISSUE #354: Update docs and track advanced feature we want to bring into a uv managed backend
This is a follow up from PR #350  that was merged, but we need to update the docs for this and keep track of some advanced features that would be great to have documented and/or added to the cookieplone templates. 

We would need to do some research a few of those further, but ideas and options are already in the (closed) PR discussion. 

* Override use a different version of a released package than the KGS managed by  uv in the pyproject.toml I don't fully understand yet why this should be done with an 'extra' list in pyproject.toml and changes in repoplone. yet. Once the project template is generated you update /backend/pyproject.toml is my  impression. 

* Then we want to support source checkouts: this works by using  uv.tool.sources in the pyproject.toml, but the checkout itself to the filesystem is not done by uv (it is done by mxdev).  AFAICS what @erral  suggested in the PR is to lext mxdev do this part stll. But @ericof mentioned repoplone could take care of this as well.  How much I have used and like the mxdev features, keeping only this part in mxdev for doing checkouts and it can work, is a guarantee to confuse people using the generated project scaffold.  

@davisagli  posted a list of where mxdev is mentioned. on the PR  Some questions:

* Can we 'migrate' the features in short enough time that we can update the documentation in one go?  Or will this take to olong and should we do the update in two phases?  How long do we give ourselves for that (timeboxing a bit).   

*  @stevepiercy Should we copy the resulting  "pages to update with what"  into one or two plone/documentation issues? 

@ericof  Created a separate documentation issue on https://github.com/plone/cookieplone/issues/132, but this is about the improved cookieplone 2.0 functionality AFAICS and about the template tooling, not wisdom/FAQ's about using the tools in the generated scaffold to develop/run/release your project. 

Meta "how to make future efforts  easier: we have been discussing this before, with  updates to the frontend/backend tooling in a generated plone project scaffold, spreading the basics and advanced features over 5-10 pages in the documentation is not much 'fun' to track  and update. 

We already did a great job to let trainings refer to docs.plone.org,  but we should check those as well. But also inside docs.plone.org, the fewer pages we have on the details and we refer to , the easier future updates will be. 

But just as there is 'philosophy' about cookieplone, there is also extra tips/tricks/lore around the generated cookieplone scaffold and how to use it. 

This is an explorartory and 'what should we do issu"  so we are aware and can update each other of the scope and attempt for a logical 'order'. What am I missing, or what are better alternatives?

# To do

The following docs should be reviewed and updated:
- [x] https://6.docs.plone.org/overview/index.html
- [x] https://6.docs.plone.org/conceptual-guides/package-management.html
- [x] https://6.docs.plone.org/admin-guide/override-core.html
- [x] https://6.docs.plone.org/conceptual-guides/make-backend-build.html
- [x] https://6.docs.plone.org/admin-guide/add-ons.html

davisagli:
Fixes https://github.com/plone/cookieplone-templates/issues/354

Summary:
- Update conceptual guide on packaging to cover more of the tools used in the community
- Update "Override core Plone packages" how-to to include uv-managed installations, and organize into tabs based on the different package managers.
- Update "Install Plone add-ons" how-to to include uv-managed installations, and organize into tabs based on the different package managers.

----
📚 Documentation preview 📚: https://plone6--2072.org.readthedocs.build/

### DIFF
--- a/docs/admin-guide/add-ons.md
+++ b/docs/admin-guide/add-ons.md
@@ -23,12 +23,50 @@ See {doc}`/volto/development/add-ons/index`.
 For instructions to install an add-on that is released on {term}`PyPI`, select the tab below according to your Python package manager.
 
 ```{tip}
-Select the tab for uv if you have a project that was created using Cookieplone.
+Select the tab for uv if you have a project that was created using Cookieplone, and you have `managed = true` set in the `[tool.uv]` section of the file {file}`backend/pyproject.toml`.
+Select the tab for pip if you have a project that was created using Cookieplone that does not have this setting.
 ```
 
 `````{tab-set}
 
 ````{tab-item} uv
+
+Add the name of your add-on in the file {file}`backend/pyproject.toml` in the section `dependencies`.
+This example adds [`collective.easyform`](https://pypi.org/project/collective.easyform/).
+
+```{code-block} toml
+:emphasize-lines: 6
+dependencies = [
+    "Products.CMFPlone==6.1.1",
+    "plone.api",
+    "plone.classicui",
+    "plone.app.caching",
+    "collective.easyform==4.4.0",
+]
+```
+
+```{tip}
+Including the add-on version, or "pinning a version", ensures that it won't unintentionally get upgraded in the future.
+```
+
+Also add the add-on to `zcml_package_includes` in the file {file}`backend/instance.yaml` to make sure its configuration will be loaded.
+
+```yaml
+default_context:
+    zcml_package_includes: project_title, collective.easyform
+```
+
+Stop the backend with {kbd}`ctrl-c`.
+
+To actually download and install the new add-on, run the following command.
+
+```shell
+make backend-build
+```
+
+````
+
+````{tab-item} pip
 
 Add the name of your add-on in the file {file}`backend/pyproject.toml` in the section `dependencies`.
 This example adds [`collective.easyform`](https://pypi.org/project/collective.easyform/).
@@ -126,7 +164,8 @@ An unreleased add-on can be installed from a source control system such as GitHu
 For instructions to install an add-on that is released on {term}`PyPI`, select the tab below according to your Python package manager.
 
 ```{tip}
-Select the tab for uv if you have a project that was created using Cookieplone.
+Select the tab for uv if you have a project that was created using Cookieplone, and you have `managed = true` set in the `[tool.uv]` section of the file {file}`backend/pyproject.toml`.
+Select the tab for pip if you have a project that was created using Cookieplone that does not have this setting.
 ```
 
 `````{tab-set}
@@ -160,6 +199,58 @@ default_context:
 ```
 
 Stop the backend with {kbd}`ctrl-c`.
+````
+
+````{tab-item} pip
+
+Add the name of your add-on in the file {file}`backend/pyproject.toml` in the section `dependencies`.
+This example adds [`collective.easyform`](https://pypi.org/project/collective.easyform/).
+
+```{code-block} toml
+:emphasize-lines: 6
+dependencies = [
+    "Products.CMFPlone==6.1.1",
+    "plone.api",
+    "plone.classicui",
+    "plone.app.caching",
+    "collective.easyform",
+]
+```
+
+```{note}
+When installing an add-on from source, it's best not to pin a version.
+This way you always get the version that's currently available in the source control system.
+```
+
+Next add the add-on to `zcml_package_includes` in the file {file}`backend/instance.yaml` so that its configuration will load.
+
+```yaml
+default_context:
+    zcml_package_includes: project_title, collective.easyform
+```
+
+Finally, add the package's source to the file {file}`mx.ini`.
+
+```cfg
+[collective.easyform]
+url=git@github.com:collective/collective.easyform.git
+branch=dev-branch-name
+extras=test
+```
+
+```{seealso}
+The {file}`mx.ini` file configures a tool called {term}`mxdev`.
+For an explanation of why Plone uses `mxdev`, see {ref}`manage-packages-mxdev-label`.
+```
+
+Stop the backend with {kbd}`ctrl-c`.
+
+To actually download and install the new add-on, run the following command.
+
+```shell
+make backend-build
+```
+
 ````
 
 ````{tab-item} Buildout
--- a/docs/admin-guide/add-ons.md
+++ b/docs/admin-guide/add-ons.md
@@ -20,7 +20,13 @@ See {doc}`/volto/development/add-ons/index`.
 
 ## Install an add-on from PyPI
 
-For instructions to install an add-on that is released on {term}`PyPI`, select the tab below according to your Python package manager.
+This section describes how to install an add-on that is released on {term}`PyPI`
+
+
+### Configure add-on installation
+
+First, configure your project according to the instructions in the tabbed interface below.
+Select the tab according to your Python package manager.
 
 ```{tip}
 Select the tab for uv if you have a project that was created using Cookieplone, and you have `managed = true` set in the `[tool.uv]` section of the file {file}`backend/pyproject.toml`.
--- a/docs/admin-guide/add-ons.md
+++ b/docs/admin-guide/add-ons.md
@@ -18,13 +18,17 @@ The Volto frontend has its own system of add-ons using Node.js packages.
 See {doc}`/volto/development/add-ons/index`.
 ```
 
+## Install an add-on from PyPI
 
-## Cookieplone
+For instructions to install an add-on that is released on {term}`PyPI`, select the tab below according to your Python package manager.
 
-Use the following instructions if you installed Plone with Cookieplone.
+```{tip}
+Select the tab for uv if you have a project that was created using Cookieplone.
+```
 
+`````{tab-set}
 
-### Install an add-on
+````{tab-item} uv
 
 Add the name of your add-on in the file {file}`backend/pyproject.toml` in the section `dependencies`.
 This example adds [`collective.easyform`](https://pypi.org/project/collective.easyform/).
@@ -59,83 +63,9 @@ To actually download and install the new add-on, run the following command.
 make backend-build
 ```
 
-Now restart the backend.
+````
 
-```{seealso}
-{doc}`run-plone`
-```
-
-In your web browser, and assuming you are currently logged in as an administrator, visit the URL http://localhost:8080/Plone/prefs_install_products_form.
-
-Then click the {guilabel}`Install` button next to your add-on to complete installation of the add-on.
-
-Some add-ons have configuration options.
-To configure such add-ons, return to the {guilabel}`Site Setup` control panel.
-At the bottom of the page, you should see the heading {guilabel}`Add-on Configuration`, and a control panel to configure the add-on that you just installed.
-
-
-### Install an add-on from source
-
-An add-on can be installed from a source control system such as GitHub.
-
-Add a line with the name of your add-on in the file {file}`backend/requirements.txt`.
-This example uses [`collective.easyform`](https://pypi.org/project/collective.easyform/).
-
-```
-collective.easyform
-```
-
-```{note}
-When installing an add-on from source, it's best not to pin a version.
-This way you always get the version that's currently available in the source control system.
-```
-
-Next add the add-on to `zcml_package_includes` in the file {file}`backend/instance.yaml` so that its configuration will load.
-
-```yaml
-default_context:
-    zcml_package_includes: project_title, collective.easyform
-```
-
-Finally, add the package's source to the file {file}`mx.ini`.
-
-```cfg
-[collective.easyform]
-url=git@github.com:collective/collective.easyform.git
-branch=dev-branch-name
-extras=test
-```
-
-```{seealso}
-The {file}`mx.ini` file configures a tool called {term}`mxdev`.
-See the [documentation of `mxdev` in its README.md](https://github.com/mxstack/mxdev/blob/main/README.md) for complete information.
-```
-
-Stop the backend with {kbd}`ctrl-c`.
-
-To actually download and install the new add-on, run the following command.
-
-```shell
-make backend-build
-```
-
-Now restart the backend.
-
-```{seealso}
-{doc}`run-plone`
-```
-
-In your web browser, and assuming you are currently logged in as an administrator, visit the URL http://localhost:8080/Plone/prefs_install_products_form.
-An upgrade step might need to be performed in the Plone control panel.
-Follow the upgrade information, if present.
-Else click the {guilabel}`Install` button to complete installation of the add-on.
-
-
-## Buildout
-
-Use the following instructions if you installed Plone with Buildout.
-
-### Install an add-on
+````{tab-item} Buildout
 
 Update the file {file}`buildout.cfg`.
 This example uses [`collective.easyform`](https://pypi.org/project/collective.easyform/).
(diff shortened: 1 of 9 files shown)


docs/conceptual-guides/package-management.md
```diff
@@ -24,23 +24,35 @@ Python itself has a complex and convoluted history with package management, as [
...
-plone.api>=2.1.0
+```shell
+bin/pip install -c https://dist.plone.org/release/6.1-latest/constraints.txt Plone
 ```
 
-Unfortunately `pip` does not allow overriding constraints this way.
-{term}`mxdev` solves this issue.
+In the Plone community, constraints are sometimes called "version pins."
+
+As a best practice, pip should always be used inside a specific {term}`virtualenv` to keep the packages separate from other applications.
```
stevepiercy:
```suggestion
As a best practice, pip should always be used inside a specific Python {term}`virtual environment` to keep the packages separate from other applications.
```

docs/glossary.md
```diff
@@ -964,4 +964,7 @@ Plate
     [Plate](https://platejs.org/) is a {term}`Slate`-based editor, introduced in Seven.
     Plate has a large community and provides a rich set of plugins to customize the editor experience.
     Key features include the single page editor, as well as AI integration.
+
+virtualenv
+    A virtualenv is an isolated Python environment with its own set of installed packages.
```
stevepiercy: virtualenv is the name of the Python tool to create Python virtual environments. I added a few more terms to future use.

I want to make sure they render correctly in the [Glossary](https://plone6--2072.org.readthedocs.build/glossary.html#term-virtualenv), too, as I'm doing this all through GitHub without building.

```suggestion
virtual environment
    A virtual environment is an isolated Python environment with its own set of installed packages.
    A virtual environment is created on top of an existing Python installation, known as the virtual environment's "base" Python, and by default is isolated from the packages in the base environment, so that only those explicitly installed in the virtual environment are available.
    
    ```seealso
    -   {term}`virtualenv`
    -   {term}`venv`
    ```

virtualenv
    [`virtualenv`](https://virtualenv.pypa.io/en/latest/) is a tool to create isolated Python environments.
    Since Python 3.3, a subset of it has been integrated into the standard library under the {term}`venv` module.

venv
    The {doc}`venv` module in the Python standard library supports creating lightweight {term}`virtual environment`s, each with their own independent set of Python packages installed in their {doc}`site` directories.
```

docs/admin-guide/override-core.md
```diff
@@ -14,26 +14,26 @@ myst:
...
+Choose a tab depending on your Python package manager.
 
-Use the following instructions if you installed Plone with Cookieplone.
+`````{tab-set}
 
+````{tab-item} uv
 
-### Override a core Plone package
+```{tip}
+Use the following instructions if you have a project that was created using Cookieplone after March 2026.
```
stevepiercy: If I create a project today, that's during March, and I'd have no instructions to follow.

An explicit date is preferred. Better yet, let's refer to Cookieplone v1.0.0 instead of a date. In which case, this would be a `versionchanged` admonition instead of `tip`.

If necessary, we can also use nested tabs for pre-1.0.0 and 1.0.0+.

stevepiercy: Actually, now that I see it applies for both uv and pip, let's move this admonition before the tabbed UI so that readers know which is relevant to their situation before they click a tab.

docs/admin-guide/override-core.md
```diff
@@ -50,19 +50,26 @@ Now restart the backend.
...
 
+````
 
-### Install a core Plone package from source
+````{tab-item} pip
 
-You can also use `mxdev` to install core Plone packages from a source control system such as GitHub.
+```{tip}
+Use the following instructions if you have a project that was created using Cookieplone before March 2026.
+```
```
stevepiercy: See above comment.

docs/admin-guide/override-core.md
```diff
@@ -50,19 +50,26 @@ Now restart the backend.
...
-extras = test
+```
+[settings]
+version-overrides =
+    plone.api==2.0.0a3
+```
+
+```{seealso}
+The {file}`mx.ini` file configures a tool called {term}`mxdev`.
+For an explanation of why Plone uses `mxdev`, see {ref}`manage-packages-mxdev-label`.
```
stevepiercy:
```suggestion
For an explanation of why Plone uses `mxdev` for projects using pip as the Python package manager, see {ref}`manage-packages-mxdev-label`.
```

docs/admin-guide/override-core.md
```diff
@@ -79,12 +86,13 @@ Now restart the backend.
...
+````
 
-## Buildout
+````{tab-item} Buildout
 
+```{tip}
 Use the following instructions if you installed Plone with Buildout.
-
-### Override a core Plone package
+```
```
stevepiercy: This is redundant to the instruction "Choose a tab depending on your Python package manager."

```suggestion
```

docs/admin-guide/override-core.md
```diff
@@ -14,26 +14,26 @@ myst:
 Plone includes a few hundred Python packages.
 Sometimes you will need to override one or more package versions to fix a bug.
 
+## Override the version of a core Plone package
 
-## Cookieplone
+Choose a tab depending on your Python package manager.
```
stevepiercy:
```suggestion
For instructions of how to override a core Plone package, select the tab below according to your Python package manager.
```

docs/admin-guide/override-core.md
```diff
@@ -124,11 +132,92 @@ Then restart your instance.
...
+
+This example uses `plone.restapi`.
+
+Clone the repository into a local directory.
+
+```shell
+git clone git@github.com:plone/plone.restapi.git
+```
+
+Add the local directory to your `uv` project as an editable package.
```
stevepiercy: The name is just plain old "uv", and not an inline literal. If it's a command, then `uv` is appropriate.

```suggestion
Add the local directory to your uv project as an editable package.
```

docs/admin-guide/override-core.md
```diff
@@ -124,11 +132,92 @@ Then restart your instance.
...
+````
+
+`````
 
-### Install a core Plone package from source
+## Install a core Plone package from source
 
 A core Plone package can be installed from a source control system such as GitHub.
 
+Choose a tab depending on your Python package manager.
```
stevepiercy: From here onward, I'd like to make sure that the instructions for selecting a given tab are consistent with my suggestions above.

docs/conceptual-guides/package-management.md
```diff
@@ -56,9 +68,31 @@ You or your development tools, such as GNU Make, must perform that step.
 {doc}`/admin-guide/add-ons`
 ```
 
+### uv
+
+More recently, {term}`uv` has become popular as a way to install Python packages.
+This package manager is popular for its speed, its ability to manage the installation of Python itself, and its ability to consistently reproduce installed packages using a {file}`uv.lock` file.
+
+When a project is fully managed using `uv`, it is configured in `pyproject.toml` and the packages are installed using `uv sync`.
```
stevepiercy:
```suggestion
When a project is fully managed using uv, it is configured in `pyproject.toml` and the packages are installed using `uv sync`.
```

docs/conceptual-guides/package-management.md
```diff
@@ -56,9 +68,31 @@ You or your development tools, such as GNU Make, must perform that step.
...
+
+`uv` also has a backwards-compatible mode which works more like `pip`, and installs packages into a virtualenv using `uv pip install`.
```
stevepiercy:
```suggestion
uv also has a backwards-compatible mode which works more like pip, and installs packages into a virtual environment via the command `uv pip install`.
```

docs/conceptual-guides/package-management.md
```diff
@@ -24,23 +24,35 @@ Python itself has a complex and convoluted history with package management, as [
...
 
 
-(manage-backend-python-packages-label)=
-
 ## Manage backend Python packages
 
-If you want to check out a Plone core package for development, or want to override the constraints of Plone, normally you would define constraints with a file {file}`constraints.txt` to tell `pip` to install a different version of a Plone package.
+### pip
+
+By convention in the Python community, {term}`pip` is commonly used to install Python packages. It is one supported way to install the Plone backend.
```
stevepiercy:
```suggestion
By convention in the Python community, {term}`pip` is commonly used to install Python packages.
It is one supported way to install the Plone backend.
```

docs/conceptual-guides/package-management.md
```diff
@@ -24,23 +24,35 @@ Python itself has a complex and convoluted history with package management, as [
...
 
-### `mxdev` to the rescue!
+### mxdev
+
+During development, it is sometimes necessary to override the Plone version constraints.
+This makes it possible to:
+- install a newer version of a core Plone package that was released with a bugfix
+- install an unreleased core Plone package from a source control system
+
+Unfortunately `pip` does not allow overriding constraints this way.
```
stevepiercy:
```suggestion
Unfortunately pip does not allow overriding constraints this way.
```

docs/conceptual-guides/package-management.md
```diff
@@ -56,9 +68,31 @@ You or your development tools, such as GNU Make, must perform that step.
...
+### uv
+
+More recently, {term}`uv` has become popular as a way to install Python packages.
+This package manager is popular for its speed, its ability to manage the installation of Python itself, and its ability to consistently reproduce installed packages using a {file}`uv.lock` file.
+
+When a project is fully managed using `uv`, it is configured in `pyproject.toml` and the packages are installed using `uv sync`.
+
+`uv` also has a backwards-compatible mode which works more like `pip`, and installs packages into a virtualenv using `uv pip install`.
+
+If you create a Plone project using Cookieplone, it creates a backend managed by `uv` (starting in March 2026).
```
stevepiercy: Also align with Cookieplone version per above comment and `versionchanged` admonition.

```suggestion
If you create a Plone project using Cookieplone, it creates a backend managed by uv.
```

docs/conceptual-guides/package-management.md
```diff
@@ -56,9 +68,31 @@ You or your development tools, such as GNU Make, must perform that step.
...
+
+### buildout
+
+{term}`Buildout` is a tool for installing Python packages that has been used in the Plone community since about 2007, and is still preferred by some members of the community.
+
+It not only installs packages, but can set up other things using an extensible system of "recipes."
+
+Buildout does not install Python packages into a virtualenv.
```
stevepiercy:
```suggestion
Buildout does not install Python packages into a virtual environment.
```

docs/conceptual-guides/package-management.md
```diff
@@ -56,9 +68,31 @@ You or your development tools, such as GNU Make, must perform that step.
...
 ## Manage frontend Node.js packages
 
-```{todo}
-Why do we use pnpm?
-```
+### pnpm
+
+Plone uses {term}`pnpm` to install Node.js packages.
+
+Compared to the standard {term}`NPM`, it has features that help with developing multiple Node.js packages in the same workspace.
```
stevepiercy: Brand name is lowercase.

```suggestion
Compared to the standard {term}`npm`, it has features that help with developing multiple Node.js packages in the same workspace.
```

stevepiercy (changes requested):
First pass on draft. I think we'll need another pass or two. Thanks for your work on this!

docs/admin-guide/override-core.md
davisagli: We can't use the Cookieplone version. The change happened in cookieplone-templates, not Cookieplone. Cookieplone clones the main branch of cookieplone-templates (unless it's told to use a specific tag by setting the COOKIEPLONE_REPOSITORY_TAG environment variable, or some other set of templates). There's actually no release of Cookieplone 2.x yet, but the updated templates are already used.

That's why I went with something based on a date. I can update it to mention a specific day. And I like the idea to move the admonition above the tabs to help the reader decide. Will update.

docs/admin-guide/override-core.md
davisagli: I was thinking it would provide confirmation to the user that they're on the right tab, but I trust your judgment here.

docs/conceptual-guides/package-management.md
davisagli: You don't think the clarification about when that became effective is helpful?

docs/conceptual-guides/package-management.md
davisagli: I was following the capitalization I saw in the glossary.

docs/glossary.md
davisagli: Good point. Thanks.

davisagli (commented):
@stevepiercy Thanks for the quick first-pass review. I applied most of the suggestions and will follow up later on the others.

docs/admin-guide/override-core.md
davisagli: Updated in https://github.com/plone/documentation/pull/2072/commits/cd84028519885bb0ee59dc3a8f0a66a2c7d37948

docs/conceptual-guides/package-management.md
davisagli: I fixed it in the glossary in https://github.com/plone/documentation/pull/2072/commits/0f5e395a5ec5252743d1ff4dd19e43974f0e14b6

docs/glossary.md
davisagli: I fixed a few things in https://github.com/plone/documentation/pull/2072/commits/0f5e395a5ec5252743d1ff4dd19e43974f0e14b6

docs/admin-guide/override-core.md
davisagli: Actually I realized using a date doesn't work well either, because the classic_project template hasn't been updated to a uv-managed approach. I updated it to refer to the managed setting in pyproject.toml as a way to tell which way to use.

davisagli:
I've now updated a couple more chapters. I think this is ready for review.

stevepiercy:
@davisaglis how about on every commit to `main` in `cookieplone-templates`, run CI to automatically tag it with a timestamp, then use that tag when running Cookieplone?

I just now looked, and it appears that @ericof had a similar thought, based on https://github.com/plone/cookieplone-templates/releases/tag/20260320.1. Should I create a new issue in `cookieplone-templates` to have this discussion?

docs/admin-guide/add-ons.md
```diff
@@ -18,13 +18,18 @@ The Volto frontend has its own system of add-ons using Node.js packages.
...
 
+## Install an add-on from PyPI
 
-## Cookieplone
+For instructions to install an add-on that is released on {term}`PyPI`, select the tab below according to your Python package manager.
 
-Use the following instructions if you installed Plone with Cookieplone.
+```{tip}
+Select the tab for uv if you have a project that was created using Cookieplone, and you have `managed = true` set in the `[tool.uv]` section of the file {file}`backend/pyproject.toml`.
+Select the tab for pip if you have a project that was created using Cookieplone that does not have this setting.
```
stevepiercy: Also maybe use the tag of https://github.com/plone/cookieplone-templates/releases/tag/20260320.1 with a `versionadded`, assuming that's true.

docs/admin-guide/add-ons.md
```diff
@@ -59,58 +64,35 @@ To actually download and install the new add-on, run the following command.
...
+    "plone.app.caching",
+    "collective.easyform==4.4.0",
+]
 ```
 
-```{note}
-When installing an add-on from source, it's best not to pin a version.
-This way you always get the version that's currently available in the source control system.
+```{tip}
+Including the add-on version, or "pinning a version", ensures that it won't unintentionally get upgraded in the future.
```
stevepiercy: Let's combine both the original and new options. Its target audience is first-timers.

```suggestion
You can control which version of an add-on to install through "version pinning."
Specify the add-on version to avoid its unintentional upgrade.
Leave it off to always install the latest version.
```

docs/admin-guide/add-ons.md
stevepiercy:
```suggestion
Also add the add-on in the file {file}`backend/instance.yaml` under the key `default_context` as the value to the key `zcml_package_includes` to load its configuration.
```

docs/admin-guide/override-core.md
stevepiercy: See https://github.com/plone/documentation/pull/2072#issuecomment-4132204551.

docs/conceptual-guides/package-management.md
stevepiercy: I do, but see https://github.com/plone/documentation/pull/2072#issuecomment-4132204551.

stevepiercy (commented):
I created a separate PR for `docs/admin-guide/add-ons.md` only. I haven't reviewed the other files closely yet. I think that a lot of the work in #2073 could be reused here.

davisagli:
> @davisaglis how about on every commit to main in cookieplone-templates, run CI to automatically tag it with a timestamp, then use that tag when running Cookieplone? I just now looked, and it appears that @ericof had a similar thought, based on https://github.com/plone/cookieplone-templates/releases/tag/20260320.1. Should I create a new issue in cookieplone-templates to have this discussion?

@stevepiercy I think it's worth discussing if we can improve our release process for cookieplone-templates. But I want to keep that out of scope for this PR, where my goal is to document the status quo better. So yes, a new issue there sounds good.

(Also it doesn't really help here, because the distinction that matters is whether or not the project is configured as uv-managed, and as I realized yesterday, that varies by other factors than just the cookieplone-templates version.)

docs/admin-guide/override-core.md
davisagli: I don't see how your suggestion there helps, since we have to explain how to check if the project is uv-managed regardless of the cookieplone-templates version.

I did have an idea for a simpler way to check this, which I'll explain in a new comment on the current version of the file.

docs/admin-guide/add-ons.md
davisagli: Like I said in other comments, there are other factors than just the version.

I realized there's an easier way to check. If backend/mx.ini exists, use the pip tab. Otherwise, use the uv tab. I'll make this update.

davisagli: Done in https://github.com/plone/documentation/pull/2072/commits/4dd55dd564d31e33dcc0768b3a57eb37536ced9b

davisagli:
@stevepiercy I tried to update override-core.md to follow your improvements from add-ons.md. I also moved the repeated instructions for restarting the instance into a snippet in the _inc folder.

erral:
I have created a PR for mxdev adding support for the `uv.tool.sources` option, hopefully we can use that option too in case someone sticks to mxdev for handling checkouts https://github.com/mxstack/mxdev/pull/81

stevepiercy:
@erral would you like to create either a suggestion or a separate PR against this branch to document the option, or should we wait until your PR in mxdev gets merged and create a new issue as a placeholder to update these docs?

I'll do a final review of this PR tonight. I think it's good, but I ought to check against the recent merge of the other PR.

stevepiercy (approved):
I think this is ready to merge. @davisagli if you agree, would you please do the honors?

stevepiercy:
@erral I created a new issue as a placeholder for whenever your PR gets merged.

https://github.com/plone/documentation/issues/2076

PR #2073, opened 2026-03-26
Build upon #2702

stevepiercy:
See commit history for explanation of my process.

----
📚 Documentation preview 📚:

https://plone6--2073.org.readthedocs.build/admin-guide/add-ons.html


davisagli (approved):
@stevepiercy Thanks, these all look like good improvements. I'll merge, then try to update override-core.md to match.

I had been thinking it would be nice if there was a way to sync the tabs, but didn't think to check if the option already existed!

PR #2074, opened 2026-03-27
Another revision of uv-managed branch

stevepiercy:
Revises #2072 

----
📚 Documentation preview 📚: https://plone6--2074.org.readthedocs.build/

### DIFF
--- a/docs/admin-guide/add-ons.md
+++ b/docs/admin-guide/add-ons.md
@@ -210,7 +210,7 @@ extras=test
 
 ```{seealso}
 The {file}`mx.ini` file configures a tool called {term}`mxdev`.
-For an explanation of why Plone uses `mxdev`, see {ref}`manage-packages-mxdev-label`.
+For an explanation of why Plone uses mxdev, see {ref}`manage-packages-mxdev-label`.
 ```
 ````
 
--- a/docs/admin-guide/add-ons.md
+++ b/docs/admin-guide/add-ons.md
@@ -126,13 +126,8 @@ Next, restart the backend.
 {doc}`run-plone`
 ```
 
-In your web browser, and assuming you are currently logged in as an administrator, visit the URL http://localhost:8080/Plone/prefs_install_products_form.
-
-Then click the {guilabel}`Install` button next to your add-on to complete installation of the add-on.
-
-Some add-ons have configuration options.
-To configure such add-ons, return to the {guilabel}`Site Setup` control panel.
-At the bottom of the page, you should see the heading {guilabel}`Add-on Configuration`, and a control panel to configure the add-on that you just installed.
+```{include} /_inc/_add-on-complete-install.md
+```
 
 
 ## Install an add-on from source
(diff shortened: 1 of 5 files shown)


davisagli:
@stevepiercy Since this has the 6.0 branch as a target instead of the uv-managed branch, it is hard to see what you changed. I will change the target to uv-managed so I can review.

docs/admin-guide/add-ons.md
```diff
@@ -98,17 +98,39 @@ You can control which version of an add-on to install through "version pinning."
 
 ### Install the add-on
 
-```{include} /_inc/_build-and-restart.md
+If the backend is running, stop it with {kbd}`ctrl-c`.
```
davisagli, line 101: @stevepiercy I was curious why you stopped using the _build-and-restart include here. Ah, it's because we have different tabs for this one.

That's why I originally repeated the uv and pip tabs even though the instructions happen to be the same for this task, to keep them consistent with the other tasks.

But I defer to your preference here.

docs/admin-guide/override-core.md
```diff
@@ -16,11 +16,11 @@ Sometimes you will need to override one or more package versions to fix a bug.
 
 ## Override the version of a core Plone package
 
-The Python packages which are dependencies of Plone are pinned to specific versions at the time a Plone release is created.
+The Python packages on which Plone depends are pinned to specific versions at the time a Plone release is created.
```
davisagli: I intentionally phrased it to use the word "dependencies" which is more likely to be recognized in this context. But I don't feel strongly.

Maybe could be simpler:
```suggestion
Plone's Python package dependencies are pinned to specific versions at the time a Plone release is created.
```

docs/conceptual-guides/package-management.md
```diff
@@ -26,6 +26,21 @@ Python itself has a complex and convoluted history with package management, as [
 
 ## Manage backend Python packages
 
+uv, pip with mxdev, buildout and are supported tools to manage the Python packages in the Plone backend.
```
davisagli:
```suggestion
uv, pip with mxdev, and buildout are supported tools to manage the Python packages in the Plone backend.
```

davisagli (approved):
A couple suggestions, but I'm happy to merge this as is. I'll merge the fix for the obvious typo

stevepiercy:
> @stevepiercy Since this has the 6.0 branch as a target instead of the uv-managed branch, it is hard to see what you changed. I will change the target to uv-managed so I can review.

Thanks for the assist. It was past my bedtime. Checking suggestions now.

docs/admin-guide/add-ons.md
stevepiercy: There are a total of four discrete sections with similar but not exact content, two on each page. My reasoning is that a reader wants only one of the four sections at a time, so make it as simple as possible for them.

docs/admin-guide/override-core.md
stevepiercy: I like. Yeah, the original was an awkward mouthful.

stevepiercy:
I'll merge when green. Thank you, @davisagli!

PR #2088, opened 2026-05-28
Update installation docs with latest cookieplone

pbauer:
Adapt to the updated installation process with cookieplone 2.0.0 with https://github.com/plone/cookieplone-templates/pull/400 (release pending):

* There is now only one project that asks for the frontend you want to use
* Addons can be checked out using mx.ini
* Fix docs for overriding core packages (needs to be `override-dependencies` instead of `constraint-dependencies`)


stevepiercy (dismissed):
Thanks @pbauer!

Editorial content and HTML rendering look good:

- https://plone6--2088.org.readthedocs.build/admin-guide/add-ons.html
- https://plone6--2088.org.readthedocs.build/admin-guide/override-core.html

Does anything else need revision, such as?

- https://plone6--2088.org.readthedocs.build/admin-guide/override-core.html#install-a-core-plone-package-from-source

Let's get a technical review from at least one other person before merging. @davisagli might have a full plate, so who else at the BSSP can review? Thank you!

docs/install/create-project-cookieplone.md
```diff
@@ -119,22 +120,20 @@ Node.js is required only for Volto projects and not for Classic UI projects.
...
 uvx cookieplone project
 ```
 
 You will be presented with a series of prompts.
 You can accept the default values in square brackets (`[default-option]`) by hitting the {kbd}`Enter` key, or enter your preferred values.
-For ease of documentation, we will use the default values.
+
+For ease of documentation, we will use the default values and thus create a project using the Volto frontend. 
+
+To create a project using the ClassicUI frontend instead answer `No` when asked `Use Volto as frontend (Yes/no)`.
```
stevepiercy: Please revert all these changes. It was intentionally written this way to distinguish the two UIs and their requirements.

docs/install/create-project-cookieplone.md
```diff
@@ -294,7 +294,7 @@ To install both the Plone backend and Volto as frontend, use the following comma
 make install
 ```
 
-This will take a few minutes.
+This will take a few minutes if you run it for the first time.
```
stevepiercy, line 286: Revert. The context is always a Volto project installed the first time.

docs/install/create-project-cookieplone.md
```diff
@@ -304,7 +304,7 @@ When the process completes successfully, it will exit with no message.
 ```
 
 
-### Start Plone
+### Start Plone backend and Volto frontend
```
stevepiercy: Revert. See subsections.

docs/install/create-project-cookieplone.md
```diff
@@ -321,30 +321,29 @@ make backend-start
...
 Change your current working directory to {file}`project-title`.
-Start the Plone frontend with the following command.
+Start the Volto frontend with the following command.
 
 ```shell
 make frontend-start
 ```
 
-The Plone frontend server starts up and emits messages to the console, and should end with output similar to the following.
+The Volto frontend server starts up and emits messages to the console, and should end with output similar to the following.
```
stevepiercy: Revert.

stevepiercy (dismissed):
I'm sorry, but these changes totally shred carefully thought out wording that's been user-tested over the last 3 years. The only changes that should be preserved are updated Cookieplone output.

erral:
We checked it during the BSSP sprint and in the technical side it is tested and OK.

erral (dismissed):

docs/install/create-project-cookieplone.md
pbauer: I know, because I wrote it. We'll discuss if we want to duplicate the cookieplone-output. The big change is that now the projects are basically the same, just once with and once without Volto. I tried to to simplify the docs by removing the noisy cookieplone-output. Once cookieplone 2.0.0 is released the current documentation is no longer accurate.

docs/install/create-project-cookieplone.md
pbauer: This is not true, people may have installed Plone before or will do again at a later time and (like me) check the docs before. Thanks to uv- und pnpm-caches subsequent installations will be a lot faster.

pbauer:
@stevepiercy This needs to be merged as soon as possible since cookieplone 2.0.0 is released.

PR #2094, opened 2026-07-15
Add a Testing chapter for backend packages

jensens:
## Why

Plone 6 has no backend testing documentation. The Developer guide has no Testing section (only Volto testing, under Contributing), and the only rendered HTML for `plone.testing` and `plone.app.testing` lives on the legacy `5.docs.plone.org` domain.

@davisagli sketched a full outline for such a chapter [in a comment on plone/pytest-plone#16](https://github.com/plone/pytest-plone/issues/16) about two years ago, covering Volto/Jest/Cypress, unittest/`zope.testrunner`, pytest, Robot, and testbrowser. This PR picks up the two pieces of that outline that cover **backend Python testing** and deliberately leaves the rest to build on top. Please treat the structure as a starting point — I expect there are opinions on placement, and I am happy to move things.

## What

Five pages, Diataxis-classified:

**Explanation** — `conceptual-guides/testing.md`
: The kinds of tests (unit / integration / functional / acceptance), why Plone has testing layers at all, and how to choose between the two runners.

**How-to** — `developer-guide/testing/`
: - `write-a-testing-layer.md` — the keystone, placed first: writing the `testing.py` layer both runners share. Shows the declarative `PloneWithPackageLayer` form for simple add-ons and the `PloneSandboxLayer` subclass form for complex ones.
: - `zope-testrunner.md` — write and run tests with `unittest` and `zope.testrunner`.
: - `pytest.md` — write and run tests with pytest, linking to the published [pytest-plone documentation](https://plone.github.io/pytest-plone/).

## Scope

- **Backend Python only.** Volto testing lives in the Volto repository and is cross-referenced, not duplicated.
- The deep API reference for the layer machinery stays in `plone.app.testing` / `plone.testing`; the how-to links there rather than reproducing it.
- A follow-up could modernize and host those two package docs, then link the hosted versions from here.

## Verification

- Every `zope-testrunner` command shown was executed against a real package. The two layer forms are taken from real coredev packages (`plone.app.dexterity`, `plone.volto`) and the `plone.app.testing` README, with every import checked against the installed packages.
- Vale reports **0 errors, 0 warnings, 0 suggestions** on the new pages; the vocabulary additions it needed are included.
- `make html` produces no new warnings, and `linkcheck` finds no broken links on the new pages.

## Related

Companion to the newly published [pytest-plone documentation](https://plone.github.io/pytest-plone/) (plone/pytest-plone#56), which this chapter's pytest guide links to.


styles/config/vocabularies/Plone/accept.txt
```diff
@@ -74,3 +74,21 @@ webpack
...
+plone.app.testing
+plone.testing
+zope.testrunner
+PloneSandboxLayer
+PloneWithPackageLayer
+IntegrationTesting
+FunctionalTesting
+setUpZope
+setUpPloneSite
+applyProfile
```
gforcada: these should be sorted like the rest of the file

docs/conceptual-guides/testing.md
```diff
@@ -0,0 +1,130 @@
...
+    Most tests you write for an add-on are integration tests.
+
+Functional test
+:   Runs against a real Plone site that commits real transactions, so a separate process, such as a browser or an HTTP client, can see the result.
+    Slower than an integration test, because the isolation is more expensive.
+    Use it when a request has to arrive over the network.
+
+Acceptance test
+:   Drives the whole system as a user would, through a browser.
+    In Plone this means Robot Framework or a Zope testbrowser for Classic UI, and Cypress for Volto.
```
gforcada: Should we rather use `Blicca` already? and `Aurora`? 🤔 

I remember reading that volto/aurora at some point they want to switch (or already did) to playwright and not use cypress anymore?

docs/conceptual-guides/testing.md
```diff
@@ -0,0 +1,130 @@
...
+The [pytest-plone](https://plone.github.io/pytest-plone/) plugin bridges the gap: it takes the layers you already declared in `testing.py` and turns them into pytest fixtures.
+
+### Choosing
+
+Neither runner is deprecated, and neither is going away.
+
+Choose `zope.testrunner` if you contribute to Plone core, or maintain a package whose tests already use it.
+There is no reward for converting a working suite.
+
+Choose pytest for new work, unless you have a reason not to.
```
gforcada:
```suggestion
Choose `pytest` for new work, unless you have a reason not to.
```

docs/developer-guide/testing/pytest.md
```diff
@@ -0,0 +1,135 @@
...
+
+```shell
+pytest --durations=0
+```
+
+## Measure coverage
+
+```shell
+pytest --cov=my.addon --cov-report term-missing
+```
```
gforcada, line 133: This example will not work if you don't do a `pip install pytest-cov`, I guess it is worth mentioning.

docs/developer-guide/testing/zope-testrunner.md
```diff
@@ -0,0 +1,186 @@
...
+Only the layer changes.
+
+## Run the tests
+
+```shell
+zope-testrunner --auto-path -s my.addon
+```
+
+`-s` takes the dotted name of the package to search.
+`--auto-path` adds that package's directory to the test search path, which is what you want for a package installed in development mode.
```
gforcada: I personally never had to use it. Not in `zc.buildout` contexts, nor in `zc.buildout`-free contexts (i.e. `plone.meta`'s `tox`-based environments), there we use `--test-path`

gforcada (commented):
Nice job! 🌟

jensens:
Thanks for the review, @gforcada! All five addressed in the last commit:

- **`--test-path` vs `--auto-path`** — you're right, and thanks for catching it. My `--auto-path` lead came from `buildout.coredev`'s `run-tests.sh`, not the common case. The primary command is now `zope-testrunner --test-path src -s my.addon` (the form the `plone.meta` `tox` environments generate), with `--auto-path` kept as the pip-installed-in-dev-mode alternative. Verified both run.
- **Cypress for Volto** — agreed the frontend tooling is in flux (Cypress → Playwright, Aurora, …). I dropped the specific tool and defer to the Volto documentation instead, so this backend page does not go stale on a frontend decision. Happy to add a concrete link if there's a canonical Volto testing page to point at.
- **`pytest --cov`** — added a `pip install pytest-cov` note.
- **Vocabulary sorting** — sorted the additions into place, and trimmed them: the class names only ever appear in code spans, which Vale does not check, so they were never needed. Down to the terms that actually appear in prose.
- **Backtick `pytest`** — applied your suggestion.

Vale is back to 0/0/0 on the five pages.

jensens:
#2097 stack on top of this

gforcada (approved):

PR #2097, opened 2026-07-24
Add a Testing API reference (plone.app.testing + plone.testing)

jensens:
Builds on #2094 (the backend testing chapter). **Targets `backend-testing-docs`; merge #2094 first.** GitHub will retarget this to `6.0` once #2094 lands.

## What

Adds an autodoc-generated **API reference** and a deeper **layer-model explanation** to the `developer-guide/testing` chapter, plus two how-tos. It covers **both** testing packages as one toolkit, because a test author does not care whether a symbol lives in `plone.app.testing` or `plone.testing`.

| Page | |
|---|---|
| `testing-api-reference.md` | Organized by task: layer model, pre-built layers & fixtures, browser & HTTP, helpers, sandboxing primitives, constants. Symbols documented at their canonical location (e.g. `pushGlobalRegistry` under `plone.testing.zca`, with a re-export note), from both packages. |
| `how-testing-layers-work.md` | Deepened with the fixture/lifecycle split. |
| `install-add-ons-in-tests.md`, `drive-the-test-browser.md` | Two how-tos absorbing the test-specific patterns from `plone.app.testing`'s README. |

## How the reference is built

`sphinx-autodoc2`, which analyses the sources **statically**. So:

- **No Plone is installed in the docs build.** `sphinx.ext.autodoc` would need to import `plone.app.testing`, which pulls in ~226 packages (all of Plone). autodoc2 needs neither. Both packages are added as **source-only submodules**.
- It coexists with the existing `sphinx.ext.autodoc` (used by `plone.api`).
- The `plone.testing` submodule pins master; the `Layer` class docstring there is currently a one-liner. plone/plone.testing#134 enriches it, after which a submodule bump renders it in full. The reference is complete either way; that entry just gets richer.

## Notes

- `Browser` is a re-export from a private module (`plone.testing._z2_testbrowser`), which autodoc2 cannot resolve, so it is documented as a short prose entry.
- Legacy ZServer (`PLONE_ZSERVER`, `PLONE_FTP_SERVER`) is noted, not documented — Plone 6 runs on WSGI.
- Vale is clean on the new pages; the vocabulary additions it needed are included and sorted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,7 +45,7 @@ templates_path = ["_templates"]
 extensions = [
     "myst_parser",
     "notfound.extension",
-    "autodoc2",  # developer-guide/testing plone.app.testing reference (static, no import)
+    "autodoc2",  # static API docs from source, no package import needed
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",  # plone.api
     "sphinx.ext.doctest",  # plone.api
@@ -302,10 +302,10 @@ html_static_path = [
 # Don't show class signature with the class' name.
 autodoc_class_signature = "separated"
 
-# -- Options for autodoc2 (plone.app.testing reference) -----------------------
+# -- Options for autodoc2 -----------------------------------------------------
 # autodoc2 analyses the source statically, so no package import and no Plone
-# installation are needed. It reads the plone.app.testing submodule directly.
-# The ``module`` key gives the namespace package its full dotted name.
+# installation are needed. It reads the package submodules directly.
+# The ``module`` key gives each namespace package its full dotted name.
 autodoc2_packages = [
     {
         "path": "../submodules/plone.app.testing/src/plone/app/testing",
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,6 +45,7 @@ templates_path = ["_templates"]
 extensions = [
     "myst_parser",
     "notfound.extension",
+    "autodoc2",  # developer-guide/testing plone.app.testing reference (static, no import)
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",  # plone.api
     "sphinx.ext.doctest",  # plone.api
@@ -301,6 +302,26 @@ html_static_path = [
 # Don't show class signature with the class' name.
 autodoc_class_signature = "separated"
 
+# -- Options for autodoc2 (plone.app.testing reference) -----------------------
+# autodoc2 analyses the source statically, so no package import and no Plone
+# installation are needed. It reads the plone.app.testing submodule directly.
+# The ``module`` key gives the namespace package its full dotted name.
+autodoc2_packages = [
+    {
+        "path": "../submodules/plone.app.testing/src/plone/app/testing",
+        "module": "plone.app.testing",
+        "auto_mode": False,
+    },
+    {
+        "path": "../submodules/plone.testing/src/plone/testing",
+        "module": "plone.testing",
+        "auto_mode": False,
+    },
+]
+autodoc2_render_plugin = "myst"
+# The plone.app.testing docstrings are reStructuredText, not MyST.
+autodoc2_docstring_parser_regexes = [(r".*", "rst")]
+
 # -- Options for MyST markdown conversion to HTML -----------------------------
 
 # For more information see:
@@ -310,6 +331,7 @@ myst_enable_extensions = [
     "attrs_inline",  # Support parsing of inline attributes.
     "colon_fence",  # You can also use ::: delimiters to denote code fences, instead of ```.
     "deflist",  # Support definition lists. https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#definition-lists
+    "fieldlist",  # Render reST field lists (:param:) from autodoc2 docstrings.
     "html_image",  # For inline images. See https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#html-images
     "linkify",  # Identify "bare" web URLs and add hyperlinks.
     "strikethrough",  # See https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#syntax-strikethrough
--- a/docs/developer-guide/testing/drive-the-test-browser.md
+++ b/docs/developer-guide/testing/drive-the-test-browser.md
@@ -1,0 +1,92 @@
+---
+myst:
+  html_meta:
+    "description": "How to drive Plone with zope.testbrowser in a functional test: open URLs, follow links, submit forms."
+    "property=og:description": "How to drive Plone with zope.testbrowser in a functional test: open URLs, follow links, submit forms."
+    "property=og:title": "Drive the test browser"
+    "keywords": "Plone, testing, zope.testbrowser, functional test, browser"
+---
+
+(drive-the-test-browser)=
+
+# Drive the test browser
+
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -17,3 +17,4 @@ sphinxcontrib-youtube
 sphinxext-opengraph
 sphinxcontrib.mermaid
 vale
+sphinx-autodoc2  # developer-guide/testing plone.app.testing reference
```
gforcada: keep this sorted please :)

docs/conf.py
```diff
@@ -301,6 +302,26 @@
 # Don't show class signature with the class' name.
 autodoc_class_signature = "separated"
 
+# -- Options for autodoc2 (plone.app.testing reference) -----------------------
```
gforcada: While nice for the PR itself, it might outdate very quickly this constant reference that `autodoc2` is added due to `plone.app.testing`. As soon, say, `plone.api` is used as well, we will not go around and update all the references.

`git blame` is much better at that for this purpose IMHO

docs/developer-guide/testing/index.md
```diff
@@ -49,10 +49,28 @@ Choose pytest for new work.
...
+
+- {doc}`write-a-testing-layer`: the `testing.py` layer both runners share.
+- {doc}`zope-testrunner`: write and run tests with `unittest` and `zope.testrunner`.
+- {doc}`pytest`: write and run tests with pytest.
+- {doc}`install-add-ons-in-tests`: apply profiles and install add-ons in a test.
+- {doc}`drive-the-test-browser`: end-to-end tests with `zope.testbrowser`.
+
+## Reference and background
+
+- {doc}`testing-api-reference`: the layers, fixtures, helpers, and sandboxing, from both packages.
```
gforcada: which `both packages`? 🤔

docs/developer-guide/testing/how-testing-layers-work.md
```diff
@@ -0,0 +1,106 @@
...
+    app = self.layer["app"]
+```
+
+Resources are **stacked**.
+When a layer sets a key that one of its bases already set, the layer's value shadows the base's for the duration of that layer, and the base's value reappears when the layer tears down.
+This is how a functional layer can, for example, replace the database with a sandboxed copy without disturbing the layer it builds on.
+
+## Layers compose through bases
+
+A layer declares its bases—the layers it builds on—through `defaultBases` or the `bases` argument.
```
gforcada: If both (`defaultBases` and `bases`) serve the same purpose (which I have no idea right now) should we only mention one? Otherwise readers might just wonder exactly that: _which one to use_ 🤷🏾 ?

docs/developer-guide/testing/install-add-ons-in-tests.md
```diff
@@ -0,0 +1,90 @@
...
+(install-add-ons-in-tests)=
+
+# Install add-ons and profiles in a test
+
+This guide shows you how to install a GenericSetup profile or an add-on inside a test, and how to check that the installation did what you expect.
+
+It uses the helpers from {doc}`testing-api-reference`.
+The examples assume a layer whose fixture already loaded your add-on's ZCML—see {doc}`write-a-testing-layer`.
+
+If you use pytest, {doc}`pytest-plone </developer-guide/testing/pytest>` also offers an `installer` fixture and an `@pytest.mark.portal(profiles=[...])` marker that do the same thing with less boilerplate.
```
gforcada, line 20: that could be an admonition? to better highlight it?

docs/developer-guide/testing/testing-api-reference.md
```diff
@@ -0,0 +1,219 @@
...
+
+# Testing API reference
+
+Technical reference for Plone's testing API, drawn from two packages:
+
+- [plone.app.testing](https://github.com/plone/plone.app.testing): the Plone-specific layers and helpers.
+- [plone.testing](https://github.com/plone/plone.testing): the underlying layer model and the Zope-level tools.
+
+You import from whichever package a symbol lives in, but from a test author's point of view they are one toolkit, so this page is organized by task rather than by package.
+Each symbol is documented at its canonical location; re-exports are noted.
```
gforcada: that should be an admonition IMHO

docs/developer-guide/testing/drive-the-test-browser.md
```diff
@@ -0,0 +1,92 @@
...
+
+(drive-the-test-browser)=
+
+# Drive the test browser
+
+This guide shows you how to write an end-to-end functional test that drives Plone through `zope.testbrowser`, which acts as a web browser connected to Zope in-process.
+
+```{important}
+`zope.testbrowser` runs entirely in Python and does **not** run JavaScript.
+Use it for server-rendered pages (Classic UI). To test a Volto frontend, use the Volto documentation's end-to-end testing tools instead.
```
gforcada: Should we already name Blicca here? otherwise we keep the old name still around...

As for the Volto docs, do we have a link within the docs we can point to? 🤔

gforcada (commented):
Very nice! 🌟 I have only a few suggestions for improvements rather than blockers 👍🏾

docs/developer-guide/testing/drive-the-test-browser.md
jensens: we should at least name both Blicca (formelry ClassicUI) or such.

jensens:
Thanks @gforcada! Addressed in the last commit:

- **requirements.txt** — sorted `sphinx-autodoc2` into place.
- **conf.py comment** — good point about it outdating. Made the comments generic (*"static API docs from source, no package import needed"*) instead of naming a specific consumer, so they stay true no matter which packages use autodoc2.
- **"which both packages?"** — now names them: *"across `plone.app.testing` and `plone.testing`"*.
- **pytest note** — made it a `{tip}`.
- **Blicca** — done, and thanks for the nudge: *"Blicca, the frontend formerly called Classic UI"*, with a link to the Volto testing docs for the other side. (The rest of the docs still say "Classic UI", so this page is slightly ahead — happy to align once the wider rename lands.)

Two where I did something a bit different, curious what you think:

- **`defaultBases` vs `bases`** — they aren't quite the same purpose, so rather than drop one I clarified *when* each applies: you set `defaultBases` on a reusable layer class, and pass `bases` only when instantiating a layer directly to combine existing ones (the exception, not the rule). That should answer the "which one" question without hiding the override.
- **The intro "documented at its canonical location" note** — I kept it as prose rather than a third admonition. The page already has two (the legacy-ZServer warning and the re-export note), and the Plone docs style guide asks for admonitions sparingly, so I didn't want to spend a third on orientation text. Easy to switch if you feel strongly.

docs/developer-guide/testing/index.md
gforcada: @jensens you mentioned that it was `p.a.testing` and `p.testing` and that it was updated, but on the diff I still see the same, maybe the changes were not committed? 🤔

jensens: ops, wait

jensens: Actually fixed now in 5c152983: it reads *"across `plone.app.testing` and `plone.testing`"*. Thanks for catching it.

gforcada (approved):

PR #2101, opened 2026-08-21
Custom exporter logic

gforcada:
## Description

This is part of https://github.com/plone/plone.exportimport/issues/86

Where an easy way to customize the default content exporter is added.

### DIFF
--- a/docs/admin-guide/export-import.md
+++ b/docs/admin-guide/export-import.md
@@ -112,6 +112,66 @@ Consider a File content item with UID `3e0dd7c4b2714eafa1d6fc6a1493f953` and a P
 | `content/3e0dd7c4b2714eafa1d6fc6a1493f953/data.json` | JSON File with serialized representation of a content item |
 | `content/3e0dd7c4b2714eafa1d6fc6a1493f953/file/plone.pdf` | Blob file stored in the `file` field in the content item |
 
+
+## Custom export
+
+```{versionadded} Plone 6.3
+The `IObjectsExporter` adapter is new in Plone 6.3.
+Prior to it, although possible, it was harder to customize the content exporter.
+```
+
+By default, all content from an existing Plone site is exported.
+While that's great for migrations, it is not practical/feasible for larger sites.
+
+There are other scenarios where a custom data export makes sense:
+
+- sensitive content should not be exported
+- only a specific part of the site is relevant
+- ...
+
+For that, you can _override_ the `plone.exportimport.interfaces.IObjectsExporter` adapter.
+
+In {file}`overrides.zcml` add:
+
+```XML
+<adapter
+    factory="my.addon.adapters.ObjectsExporter"
+    provides="plone.exportimport.interfaces.IObjectsExporter"
+    for="plone.base.interfaces.siteroot.IPloneSiteRoot"
+    />
+```
+
+In {file}`my.package.adapters.py` add:
+
+```python
+from collections.abc import Generator
+
+
+class ObjectsExporter:
+
+    def __init__(self, obj):
+        self.obj = obj
+        self.errors = None
+
+    def get_objects(self, query, errors) -> Generator:
+        """Return all objects to be serialized"""
+        self.errors = errors
+
+        yield from self.gather_objects()
+
+    def gather_objects(self):
+        # custom logic to select which specific content gets exported
+```
+
+With this, the default `plone-exporter` will no longer export **all content**.
+
+```{warning}
+As soon as you override the export be aware that other parts of the export might not work.
+
+Carefully check that your custom objects exporter works as expected.
+```
+
+
 ## Related content
 
 -   {doc}`/admin-guide/backup-restore-plone-buildout`


stevepiercy:
@gforcada would you please request a review from whoever can provide a technical review? I've already requested two folks, based on the linked PR and the related training author.

gforcada:
@stevepiercy thanks for doing that, I will try to remember next time. I will add another one, just to have higher chances of one actually reviewing it

stevepiercy:
I'll do an editorial review, after the technical parts are verified. Thanks for your collaboration.

docs/admin-guide/export-import.md
```diff
@@ -112,6 +112,58 @@ Consider a File content item with UID `3e0dd7c4b2714eafa1d6fc6a1493f953` and a P
...
 
+
+## Custom export
+
+By default, all content from an existing Plone site is exported.
+While that's great for migrations, it is not practical/feasible for larger sites.
+
+There are other scenarios where a custom data export makes sense:
+
+- sensible content should not be exported
```
davisagli:
```suggestion
- sensitive content should not be exported
```

"sensible" means "it makes sense"
"sensitive" means "delicate" or "secret"

docs/admin-guide/export-import.md
```diff
@@ -112,6 +112,58 @@ Consider a File content item with UID `3e0dd7c4b2714eafa1d6fc6a1493f953` and a P
...
+On `my.package.adapters.py` add:
+
+```python
+class ObjectsExporter:
+
+    def __init__(self, obj):
+        self.obj = obj
+        self.errors = None
+
+    def get_objects(self, query, errors) -> Generator:
```
davisagli, line 156: Show the import of Generator

docs/admin-guide/export-import.md
```diff
@@ -112,6 +112,58 @@ Consider a File content item with UID `3e0dd7c4b2714eafa1d6fc6a1493f953` and a P
...
+
+There are other scenarios where a custom data export makes sense:
+
+- sensible content should not be exported
+- only a specific part of the site is relevant
+- ...
+
+For that, you can _override_ the `plone.exportimport.interfaces.IObjectsExporter` adapter.
+
+On `overrides.zcml` add:
```
davisagli:
```suggestion
In {file}`overrides.zcml` add:
```

docs/admin-guide/export-import.md
```diff
@@ -112,6 +112,58 @@ Consider a File content item with UID `3e0dd7c4b2714eafa1d6fc6a1493f953` and a P
...
+
+```XML
+<adapter
+    factory="my.addon.adapters.ObjectsExporter"
+    provides="plone.exportimport.interfaces.IObjectsExporter"
+    for="plone.base.interfaces.siteroot.IPloneSiteRoot"
+    />
+```
+
+On `my.package.adapters.py` add:
```
davisagli:
```suggestion
In {file}`my.package.adapters.py` add:
```

docs/admin-guide/export-import.md
```diff
@@ -112,6 +112,58 @@ Consider a File content item with UID `3e0dd7c4b2714eafa1d6fc6a1493f953` and a P
...
+        yield from self.gather_objects()
+
+    def gather_objects(self):
+        # custom logic to select which specific content gets exported
+```
+
+With this, the default `plone-exporter` will no longer export **all content**.
+
+```{warning}
+As soon as you override the export be aware that other parts of the export might not work.
```
davisagli, line 169: I was more worried about whether the import works. For example if 10% of content is exported by the content step but 100% of relations are exported by the relations step, then the import might fail when there is a relation that refers to one of the content items that was not included. Can you at least do a sanity check that this works in your case?

docs/admin-guide/export-import.md
```diff
@@ -112,6 +112,58 @@ Consider a File content item with UID `3e0dd7c4b2714eafa1d6fc6a1493f953` and a P
...
+        """Return all objects to be serialized"""
+        self.errors = errors
+
+        yield from self.gather_objects()
+
+    def gather_objects(self):
+        # custom logic to select which specific content gets exported
+```
+
+With this, the default `plone-exporter` will no longer export **all content**.
```
davisagli, line 166: That's a bit of a limitation. Maybe you want all of the content sometimes, and different subsets at other times. This makes me wonder whether we should have named IObjectsExporter adapters, and make it possible to specify which one to use in the CLI

davisagli (commented):
This looks accurate for the proposed implementation in https://github.com/plone/plone.exportimport/issues/86, but I also asked a question there to confirm whether @gforcada has confirmed this does what he needs.

docs/admin-guide/export-import.md
gforcada: Yes, this is working for us, there are quite a lot of messages mentioning that, but either `plone.exportimport` or the underlying relation machinery handles it gracefully 🎉

docs/admin-guide/export-import.md
gforcada: I had quite a few ideas to improve what gets exported:

- add a control panel that lists what gets exported (a fancy `print` of what `IObjectsExporter` returns)
- add dexterity behavior that allows to mark objects to be always/never exported
  - thinking about privacy or mandatory objects
- on that control panel add a checkbox to use either the full or subset export

I'm not sure if there are enough users for such use cases, that's why we started with first allowing to customize the logic, and then we can expand further if there is a need for it

gforcada:
@davisagli thanks for the review, all points are either addressed or commented 😄

davisagli (commented):
Ok from me from a technical perspective.

It should have a `{versionadded}` notice since it wasn't available in existing releases.

ericof (changes requested):
I did bump into this requirement few days ago, and I  realized there is a small fix that could make importing and exporting part of a site quite easy.

Right now we support passing options to the Importer / Exporter but we do not pass these options down to the right parameter of a Content exporter.

